PR #26, opened 2023-11-10
feature: initial implementation of Files API

Maxim-Gadalov:


Maxim-Gadalov:
/deploy-review
>Deployment status: success

Maxim-Gadalov:
/deploy-review
>Deployment status: success

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,192 @@
...
+
+@Slf4j
+public class BlobStorage {
+
+    private static final String FILE_NAME_METADATA_KEY = "file_name";
+    private static final String CONTENT_TYPE_METADATA_KEY = "content_type";
+
+    private final BlobStoreContext storeContext;
+    private final BlobStore blobStore;
+    private final StorageCache cache;
```
astsiapanay: Please add a description the cache storage is used  temporary and will be replaced in the future
and mention about limitations like

- Doesn't support multiple core instances
- The state is lost after the instance restart

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Override
+    public synchronized boolean writeQueueFull() {
+        return isBufferFull;
+    }
+
+    @Override
+    public WriteStream<Buffer> drainHandler(Handler<Void> handler) {
+        vertx.executeBlocking((promise) -> {
+            synchronized (this) {
```
astsiapanay: I believe you want to synchronize instance of BlobWriterStream. Right?
If so please change `synchronized(this)` to `synchronized(BlobWriteStream.this)`

src/main/java/com/epam/aidial/core/storage/InputStreamReader.java
```diff
@@ -0,0 +1,150 @@
...
+        return this;
+    }
+
+    @Override
+    public synchronized ReadStream<Buffer> fetch(long amount) {
+        queue.fetch(amount);
+        return this;
+    }
+
+    private synchronized void readDataFromStream() {
```
astsiapanay: Vertx instance is already synchronized. I think we can remove the keyword `synchronized` safely.

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
```diff
@@ -0,0 +1,184 @@
...
+import io.vertx.core.streams.WriteStream;
+import lombok.extern.slf4j.Slf4j;
+import org.jclouds.blobstore.domain.MultipartPart;
+import org.jclouds.blobstore.domain.MultipartUpload;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Slf4j
+public class BlobWriteStream implements WriteStream<Buffer> {
```
astsiapanay, line 25: It's worth adding a class description. The class handles the core logic for download file and deserves some explanation.

mikitabut:
/deploy-review
>Deployment status: success

mikitabut:
/deploy-review
>Deployment status: success

Maxim-Gadalov:
/deploy-review
>Deployment status: success

Maxim-Gadalov:
/deploy-review
>Deployment status: success

Maxim-Gadalov:
/deploy-review
>Deployment status: success

src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -56,7 +61,9 @@ private void start() throws Exception {
             UpstreamBalancer upstreamBalancer = new UpstreamBalancer();
 
             IdentityProvider identityProvider = new IdentityProvider(settings("identityProvider"), vertx);
-            Proxy proxy = new Proxy(client, configStore, logStore, rateLimiter, upstreamBalancer, identityProvider);
+            Storage storageConfig = Json.decodeValue(settings("storage").toBuffer(), Storage.class);
+            storage = new BlobStorage(storageConfig);
```
astsiapanay: you can pass `settings("storage")` to the constructor of BlobStorage

src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -75,6 +82,9 @@ private void stop() {
             close(server, HttpServer::close);
             close(client, HttpClient::close);
             close(vertx, Vertx::close);
+            if (storage != null) {
```
astsiapanay: It had better create a utility method `close(Closable resource)` and implement the interface `Closable` on the class `BlobStorage`

src/main/java/com/epam/aidial/core/Proxy.java
```diff
@@ -42,14 +44,16 @@ public class Proxy implements Handler<HttpServerRequest> {
     public static final String HEADER_UPSTREAM_ATTEMPTS = "X-UPSTREAM-ATTEMPTS";
     public static final String HEADER_CONTENT_TYPE_APPLICATION_JSON = "application/json";
 
-    public static final int REQUEST_BODY_MAX_SIZE = 16 * 1024 * 1024;
+    public static final int REQUEST_BODY_MAX_SIZE = 512 * 1024 * 1024;
```
astsiapanay: Please add a suffix to the variable name `_BYTES` or add a comment

src/main/java/com/epam/aidial/core/controller/DownloadFileController.java
```diff
@@ -0,0 +1,85 @@
...
+                stream.pipeTo(response, result);
+                result.future().onFailure(error -> {
+                    stream.close();
+                    context.getResponse().reset();
+                });
+            } catch (IOException e) {
+                result.fail(e);
+            }
+        }).onFailure(error -> context.respond(HttpStatus.INTERNAL_SERVER_ERROR,
+                "Failed to fetch file with ID " + path));
```
astsiapanay: Replace with `Failed to fetch file by path `

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public Future<?> list(String path) {
+        BlobStorage storage = proxy.getStorage();
+        return proxy.getVertx().executeBlocking(() -> {
+            try {
+                String absolutePath = BlobStorageUtil.buildAbsoluteFilePath(context, path);
+                List<FileMetadataBase> metadata = storage.listMetadata(absolutePath);
+                context.respond(HttpStatus.OK, metadata);
+            } catch (Exception ex) {
+                log.error("Failed to list files", ex);
```
astsiapanay, line 35: Change error message to `Failed to list files by path %s, path`

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
```diff
@@ -0,0 +1,42 @@
...
+                context.respond(HttpStatus.INTERNAL_SERVER_ERROR, "Failed to list files metadata");
+            }
+
+            return null;
```
astsiapanay, line 39: Shouldn't we return future of the file metadata? Or please explain why we should return null. Please mark the method nullable in this case

src/main/java/com/epam/aidial/core/controller/UploadFileController.java
```diff
@@ -0,0 +1,56 @@
...
+                            filename,
+                            absoluteFilePath);
+                    pipe.to(writeStream, result);
+
+                    result.future()
+                            .onSuccess(success -> context.respond(HttpStatus.OK, writeStream.getMetadata()))
+                            .onFailure(error -> {
+                                writeStream.abortUpload(error);
+                                context.respond(HttpStatus.INTERNAL_SERVER_ERROR,
+                                        "Failed to upload file: " + error.getMessage());
```
astsiapanay: Change to `Failed to upload file by path %s, path`

src/main/java/com/epam/aidial/core/data/FileMetadata.java
```diff
@@ -0,0 +1,15 @@
+package com.epam.aidial.core.data;
+
+import lombok.Getter;
+
+@Getter
+public class FileMetadata extends FileMetadataBase {
```
astsiapanay, line 10: That's the matter of design. Would it be easier just to keep a single class `FileMetadata` instead of 3?
The idea is to converge file and folder classes into one and add a field is `boolean dir`.

src/main/java/com/epam/aidial/core/security/IdentityProvider.java
```diff
@@ -147,6 +147,10 @@ private DecodedJWT verifyJwt(String encodedToken, JwkResult jwkResult) {
         }
     }
 
+    private String extractUserSub(DecodedJWT decodedJwt) {
```
astsiapanay: `static`?

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private static ContentMetadata buildContentMetadata(String contentType) {
+        ContentMetadata contentMetadata = ContentMetadataBuilder.create()
+                .contentType(contentType)
+                .build();
+        return BaseMutableContentMetadata.fromContentMetadata(contentMetadata);
+    }
+
+    private void createBucketIfNeeded(Storage config) {
+        if (config.getProvider().equals("google-cloud-storage")) {
```
astsiapanay: It's ok to fail the operation on createBucket if user explicitly set the flag `needToCreateBucket` to `true` and the account doesn't have permissions. That's an issue with permissions on provider account level.
I think we should add a config property `needToCreateBucket` as a boolean value.

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+        builder.credentials(config.getIdentity(), config.getCredential());
+        this.storeContext = builder.buildView(BlobStoreContext.class);
+        this.blobStore = storeContext.getBlobStore();
+        this.bucketName = config.getBucket();
+        createBucketIfNeeded(config);
+    }
+
+    /**
+     * Initialize s3 multipart upload
```
astsiapanay: Please remove `s3`

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+     *
+     * @param fileName    name of the file, for example: data.csv
+     * @param path        absolute path according to the bucket, for example: Users/user1/files/input data
+     * @param contentType MIME type of the content, for example: text/csv
+     */
+    public MultipartUpload initMultipartUpload(String fileName, String path, String contentType) {
+        BlobMetadata metadata = buildBlobMetadata(fileName, path, contentType, bucketName);
+        return blobStore.initiateMultipartUpload(bucketName, metadata, PutOptions.NONE);
```
astsiapanay, line 69: Is there any way to use stable API? Otherwise please add `@SuppressWarnings("UnstableApiUsage")` and add a comment why we should use the unstable API

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+        return blobStore.uploadMultipartPart(multipart, part, new BufferPayload(buffer));
+    }
+
+    /**
+     * Commit multipart upload.
+     * This method must be called after all parts/chunks uploaded
+     */
+    public void completeMultipartUpload(MultipartUpload multipart, List<MultipartPart> parts) {
+        String etag = blobStore.completeMultipartUpload(multipart, parts);
+        log.info("Stored etag: " + etag);
```
astsiapanay: I guess the log message has a little sense. Please remove the line or return it in the method if we support concurrent updates.

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+    public List<FileMetadataBase> listMetadata(String path) {
+        List<FileMetadataBase> metadata = new ArrayList<>();
+        ListContainerOptions options = buildListContainerOptions(BlobStorageUtil.normalizePathForQuery(path));
+        PageSet<? extends StorageMetadata> list = blobStore.list(bucketName, options);
+        list.forEach(meta -> {
+            StorageType objectType = meta.getType();
+            switch (objectType) {
+                case BLOB -> metadata.add(buildFileMetadata(meta));
+                case FOLDER, RELATIVE_PATH -> metadata.add(buildFolderMetadata(meta));
+                default -> throw new IllegalArgumentException("Can't list container");
```
astsiapanay: The default case is a container itself. Why shouldn't we support that?

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -0,0 +1,199 @@
...
+        storeContext.close();
+    }
+
+    private static ListContainerOptions buildListContainerOptions(String prefix) {
+        return new ListContainerOptions()
+                .prefix(prefix)
+                .delimiter(BlobStorageUtil.PATH_SEPARATOR);
+    }
+
+    private static FileMetadata buildFileMetadata(StorageMetadata metadata) {
```
astsiapanay: These two methods do the same work. Wouldn't it be possible to merge them into one?

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
```diff
@@ -0,0 +1,77 @@
+package com.epam.aidial.core.storage;
+
+import com.epam.aidial.core.ProxyContext;
+import io.vertx.core.http.impl.MimeMapping;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class BlobStorageUtil {
```
astsiapanay, line 8: Please cover the methods by Unit tests

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final String USER_ROOT_DIR_PATTERN = "Users/%s/files/%s";
+
+    private static final String API_KEY_ROOT_DIR_PATTERN = "Keys/%s/files/%s";
+
+    public static final String PATH_SEPARATOR = "/";
+    private static final char DELIMITER = PATH_SEPARATOR.charAt(0);
+
+
+    public String normalizePathForQuery(String path) {
```
astsiapanay, line 23: Please add javadoc

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
```diff
@@ -0,0 +1,77 @@
+package com.epam.aidial.core.storage;
+
+import com.epam.aidial.core.ProxyContext;
+import io.vertx.core.http.impl.MimeMapping;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class BlobStorageUtil {
+
+    private static final String USER_ROOT_DIR_PATTERN = "Users/%s/files/%s";
```
astsiapanay, line 10: Usually any path starts with `/` and may be ended with `/`. Why should we remove leading and trailing path separators every time?

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Implementation of vertx {@link io.vertx.core.streams.WriteStream} that handles data chunks (from {@link io.vertx.core.streams.ReadStream}) and writes them to the blob storage.
+ * If file content is bigger than 5MB - multipart upload will be used.
+ * Chunk size can be configured via {@link #setWriteQueueMaxSize(int)} method, but should be no less than 5 MB according to the s3 specification.
+ * If any exception is caught in between - multipart upload will be aborted.
+ */
+@Slf4j
+public class BlobWriteStream implements WriteStream<Buffer> {
+
+    private static final int MIN_PART_SIZE = 5 * 1024 * 1024;
```
astsiapanay: Please a suffix to the var name `BYTES` or add a javadoc

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
```diff
@@ -0,0 +1,187 @@
...
+    public synchronized void abortUpload(Throwable ex) {
+        if (mpu != null) {
+            storage.abortMultipartUpload(mpu);
+        }
+
+        if (errorHandler != null) {
+            errorHandler.handle(ex);
+        }
+
+        log.info("Multipart upload aborted");
```
astsiapanay: the log level should be at least warning

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.List;
+
+/**
+ * Implementation of vertx {@link io.vertx.core.streams.WriteStream} that handles data chunks (from {@link io.vertx.core.streams.ReadStream}) and writes them to the blob storage.
+ * If file content is bigger than 5MB - multipart upload will be used.
+ * Chunk size can be configured via {@link #setWriteQueueMaxSize(int)} method, but should be no less than 5 MB according to the s3 specification.
+ * If any exception is caught in between - multipart upload will be aborted.
+ */
+@Slf4j
+public class BlobWriteStream implements WriteStream<Buffer> {
```
astsiapanay, line 25: We should try to write Unit test simulating slow upload to blob storage. The blob storage could be mocked.

mikitabut:
/deploy-review
>Deployment status: success

src/main/java/com/epam/aidial/core/storage/InputStreamReader.java
```diff
@@ -0,0 +1,149 @@
...
+        Future<Buffer> fetchResult = vertx.executeBlocking(() -> {
+            try {
+                byte[] data = in.readNBytes(bufferSize);
+                return BufferImpl.buffer(Unpooled.wrappedBuffer(data));
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        });
+        fetchResult.onSuccess(buf -> {
+            if (queue.write(buf) && buf.length() > 0) {
```
astsiapanay, line 108: @Question
Is the queue is full then we don't read any more data from the input stream?

astsiapanay (commented):
We need integration tests to be implemented to cover file API.

src/main/java/com/epam/aidial/core/AiDial.java
Maxim-Gadalov: I don't think it's needed, I don't want to maintain a Storage config schema directly in constructor

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
Maxim-Gadalov: In this case I merged storage and response handling, vertx `executeBlocking` expects return something, like Future.succeed() in a deprecated API

src/main/java/com/epam/aidial/core/data/FileMetadata.java
Maxim-Gadalov: I'm not a big fan of flags in class logic, we may revisit it later to use a single pattern across the project

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: why?

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: no other way to do multipart upload, added suppress

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: Users are not allowed to list containers and work with them, if somebody configured a bucket - it's an error

src/main/java/com/epam/aidial/core/storage/InputStreamReader.java
Maxim-Gadalov: if queue is full we need to stop reading data chunks, once queue will be ready - drainHandler will continue to fetch data

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
Maxim-Gadalov: some requests provide a filenames like: /file.txt (download)
some provides a folders like: /tmp or /folder1/ (upload)
some may provide a null/empty path

S3 require some patterns to download/list files, for example file path should not start with `/` and query files should end with `/`

This is a common method to "normalize" path and construct correct one
Another place where we need it - metadata (output formatting)

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
astsiapanay: I believe we support GCP and Azure. There is no reason to state explicitly s3 in this case.

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
astsiapanay: I didn't hear about a such restriction on AWS S3 side. So I believe you.

src/main/java/com/epam/aidial/core/AiDial.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/Proxy.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/controller/DownloadFileController.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/controller/UploadFileController.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/security/IdentityProvider.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
Maxim-Gadalov: added

src/main/java/com/epam/aidial/core/storage/BlobStorageUtil.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/storage/BlobWriteStream.java
Maxim-Gadalov: let's add in a separate PR

src/test/java/com/epam/aidial/core/TestFiles.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.commons.io.file.PathUtils;
+
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Objects;
+
+@UtilityClass
+public class TestFiles {
```
astsiapanay: Let's rename to "FileUtility" or "FileHelper"

src/test/java/com/epam/aidial/core/FileApiTest.java
```diff
@@ -0,0 +1,269 @@
...
+    private static int serverPort;
+    private static Path testDir;
+
+    @BeforeAll
+    public static void init() throws Exception {
+        // initialize server
+        dial = new AiDial();
+        testDir = TestFiles.baseTestPath(FileApiTest.class);
+        BlobStorage storage = buildFsBlobStorage(testDir);
+        dial.start(storage);
```
astsiapanay: I would pass a path to dial config as cmd arg to DIAL app instead of exposing one more method.
See com.epam.aidial.core.AiDial#settings()

mikitabut:
/deploy-review
>Deployment status: success

src/main/java/com/epam/aidial/core/Proxy.java
```diff
@@ -74,15 +79,24 @@ private void handleRequest(HttpServerRequest request) throws Exception {
...
-            return;
+        String contentType = request.getHeader(HttpHeaders.CONTENT_TYPE);
+        if (contentType != null && contentType.startsWith("multipart/form-data")) {
+            if (ProxyUtil.contentLength(request, 1024) > FILES_REQUEST_BODY_MAX_SIZE_BYTES) {
+                respond(request, HttpStatus.REQUEST_ENTITY_TOO_LARGE, "Request body is too large");
+                return;
+            }
+        } else {
+            // not only the case, Content-Length can be missing when Transfer-Encoding: chunked
+            if (ProxyUtil.contentLength(request, 1024) > REQUEST_BODY_MAX_SIZE_BYTES) {
```
artsiomkorzun: extract calling ProxyUtil.contentLength to line 89 and use in both branches

src/main/java/com/epam/aidial/core/config/Storage.java
```diff
@@ -0,0 +1,35 @@
+package com.epam.aidial.core.config;
+
+import lombok.Data;
+
+import javax.annotation.Nullable;
+
+@Data
+public class Storage {
```
artsiomkorzun, line 8: Add some description in README.

astsiapanay (dismissed):

artsiomkorzun (dismissed):

astsiapanay (approved):

artsiomkorzun (approved):

Issue #32, opened 2023-11-15
Pathtrough Upvote/Downvote to Application/Adapters


avavilau:
* User's upvote / downvote actions need to be delivered all the way to custom applications/assistants
* Message need to be identified by response ID, no history will be provided
* The feature is supposed to be optional: switchable via core config. We need avoid issues if destination does not support upvote/downvote
* SDK and samples (echo) need to be modified to support the feature
* Corresponding changes need to be done in APIs and front-end

astsiapanay:
depends on https://github.com/epam/ai-dial-sdk/issues/39

PR #38, opened 2023-11-17
feat: #32 Pathtrough Upvote/Downvote to Application/Adapters

astsiapanay:


src/main/java/com/epam/aidial/core/controller/RateResponseController.java
```diff
@@ -0,0 +1,178 @@
...
+
+        if (deployment == null || !DeploymentController.hasAccessByUserRoles(context, deployment)) {
+            context.respond(HttpStatus.FORBIDDEN, "Forbidden deployment");
+            return;
+        }
+
+        if (deployment.getRateEndpoint() == null) {
+            context.getResponse().setStatusCode(200);
+            context.getResponse().send();
+            proxy.getLogStore().save(context);
```
artsiomkorzun, line 43: Why don't you use shortcut context.respond(HttpStatus.OK)?

src/main/java/com/epam/aidial/core/controller/RateResponseController.java
```diff
@@ -0,0 +1,178 @@
...
+        log.warn("Can't connect to origin: {}", error.getMessage());
+        sendRequest(); // try next
+    }
+
+    /**
+     * Called when proxy failed to send request to the origin.
+     */
+    private void handleProxyRequestError(Throwable error) {
+        log.warn("Can't send request to origin: {}", error.getMessage());
+        sendRequest(); // try next
```
artsiomkorzun: Do not retry. Otherwise you get into an infinite loop of getting this error and retrying again and again.

artsiomkorzun (changes requested):
Please address comments

src/main/java/com/epam/aidial/core/controller/RateResponseController.java
astsiapanay: missed the method. fixed.

src/main/java/com/epam/aidial/core/controller/RateResponseController.java
astsiapanay: fixed.

src/main/java/com/epam/aidial/core/controller/RateResponseController.java
```diff
@@ -164,7 +163,6 @@ private void handleProxyConnectionError(Throwable error) {
      */
     private void handleProxyRequestError(Throwable error) {
```
artsiomkorzun, line 164: You need to respond to the client (server response) with BAD GATEWAY.

artsiomkorzun (approved):

PR #49, opened 2023-11-22
fix: file API tests

Maxim-Gadalov:


src/test/java/com/epam/aidial/core/FileApiTest.java
```diff
@@ -104,50 +106,64 @@ public void testFileUpload(Vertx vertx, VertxTestContext context) {
...
+                                    Assertions.assertEquals(expectedFileMetadata, response.body());
+                                    checkpoint.flag();
+                                    promise.complete();
+                                });
+                            })
+                    );
+
+            return promise.future();
+        }).compose((mapper) -> {
+            Promise<Void> promise = Promise.promise();
```
astsiapanay: the last promise is not required

astsiapanay:
Please add a reference to the issue you worked on in the commit message.

src/main/java/com/epam/aidial/core/storage/InputStreamReader.java
```diff
@@ -50,7 +50,6 @@ public InputStreamReader(Vertx vertx, InputStream in, int bufferSize) {
             }
         });
         queue.drainHandler(v -> readDataFromStream());
-        readDataFromStream();
```
artsiomkorzun, line 53: Just call queue.pause and that is it.

artsiomkorzun (approved):

src/test/java/com/epam/aidial/core/FileApiTest.java
Maxim-Gadalov: fixed

PR #66, opened 2023-11-28
feat: exposed rate feature in the listings

### LINKED ISSUE #65: Expose rate endpoint in the core config and the listing
1. Move `rateEndpoint` to `features.rateEndpoint` in order to aggregate the similar extra endpoints (`tokenize` and `truncate_prompt`) together under the `features` field.
2. Expose `feature.rate: boolean` in the listing

adubovik:
Fixed #65


src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
```diff
@@ -158,38 +158,46 @@ private static Controller selectPost(Proxy proxy, ProxyContext context, String p
...
-            return () -> controller.handle(deploymentId);
+
+            Function<Deployment, String> getter = (model) -> {
+                return Optional.ofNullable(model)
+                    .map(d -> d.getFeatures())
+                    .map(t -> t.getRateEndpoint())
+                    .orElse(null);
+            };
+
+            DeploymentFeatureController controller = new DeploymentFeatureController(proxy, context);
```
astsiapanay, line 169: The controller behaves differently than the original one:

- Don't log request body in case if there is no rate endpoint  provided
- Return 403 if rate endpoint is not provided in the config but should return 200. 403 should be returned if deployment is not found or not eligible.

src/main/java/com/epam/aidial/core/controller/DeploymentFeatureController.java
```diff
@@ -22,21 +22,31 @@
...
+        if (deployment == null || !DeploymentController.hasAccessByUserRoles(context, deployment)) {
             context.respond(HttpStatus.FORBIDDEN, "Forbidden deployment");
             return;
         }
 
+        String endpoint = endpointGetter.apply(deployment);
+        if (endpoint == null) {
+            if (requireEndpoint) {
+                context.respond(HttpStatus.FORBIDDEN, "Forbidden deployment");
+                return;
```
astsiapanay: It's redundant

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -66,10 +64,14 @@ public Future<?> handle(String deploymentId, String deploymentApi) {
             return context.respond(HttpStatus.UNSUPPORTED_MEDIA_TYPE, "Only application/json is supported");
         }
 
-        Deployment deployment = select(deploymentId, deploymentApi);
+        Deployment deployment = context.getConfig().selectDeployment(deploymentId);
+        if (!isValidDeploymentApi(deployment, deploymentApi)) {
+            deployment = null;
```
astsiapanay, line 69: Line 72. The deployment should be set to the context after the last check line 74

src/main/java/com/epam/aidial/core/controller/DeploymentFeatureController.java
adubovik: Fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
adubovik: Fixed

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
adubovik: This difference is controlled by the `requireEndpoint` flag. It's set to `false` for the rate endpoint.

It didn't log the request body indeed. It was fixed.

astsiapanay (approved):

Issue #60, opened 2023-11-27
Change authentication workflow


avavilau:
- Core must only support authorization via key or token, not both
- Application must receive tokens with restricted to its client_id

astsiapanay:
Some clarifications about the task:

[4:09 PM] Aliaksandr Stsiapanay
Questions
1. 
Supposing just token comes in the request
We always validate user token in this case anyway. (verify signature)
As a result the field Key.userAuth is no longer needed.
2. API key
User roles are empty in this case.
3.  User token
We just validate user roles but we don't validate limits on tokens per model because API key is missed.
Supposing user have unlimited access.
 
Is that correct?
[4:14 PM] Aliaksei Vavilau
Correct

PR #72, opened 2023-12-01
feat: Change authentication workflow #60

astsiapanay:
Core must only support authorization via key or token, not both


artsiomkorzun (dismissed):

src/test/java/com/epam/aidial/core/FileApiTest.java
```diff
@@ -353,7 +339,6 @@ public void testFileDelete(Vertx vertx, VertxTestContext context) {
             // try to download deleted file
             client.get(serverPort, "localhost", "/v1/files/test_file.txt")
                     .putHeader("Api-key", "proxyKey2")
-                    .bearerTokenAuthentication(generateJwtToken("User1"))
```
Maxim-Gadalov, line 356: remove method generateJwtToken if no longer needed

Maxim-Gadalov (dismissed):

artsiomkorzun (approved):

Maxim-Gadalov (approved):

Issue #77, opened 2023-12-12
Design association key with application


astsiapanay:
- Every application has assigned API key
- Core must include API key to request on every call to application.
- Core must authorize request sent by application using API key permissions in the config

PR #85, opened 2023-12-13
feat: Design association key with application #77

astsiapanay:


src/main/java/com/epam/aidial/core/config/Deployment.java
```diff
@@ -26,4 +26,5 @@ public abstract class Deployment {
     private Features features;
     private List<String> inputAttachmentTypes;
     private Integer maxInputAttachments;
+    private String apiKey;
```
avavilau, line 29: Will it be included in listing response?

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
```diff
@@ -49,19 +51,25 @@ private void load(boolean fail) {
...
                 model.setName(name);
+                model.setApiKey(generateKey());
+                config.getDeploymentApiKeys().add(model.getApiKey());
             }
 
             for (Map.Entry<String, Addon> entry : config.getAddons().entrySet()) {
                 String name = entry.getKey();
                 Addon addon = entry.getValue();
                 addon.setName(name);
+                addon.setApiKey(generateKey());
```
avavilau: I don't think it's necessary for addons

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -163,12 +163,14 @@ private void handleProxyRequest(HttpClientRequest proxyRequest) {
         if (!deployment.isForwardApiKey()) {
             excludeHeaders.add(Proxy.HEADER_API_KEY);
         }
-        if (!deployment.isForwardAuthToken()) {
-            excludeHeaders.add(HttpHeaders.AUTHORIZATION);
-        }
+        request.headers().remove(Proxy.HEADER_API_KEY);
```
avavilau: Another thing that worries me, we have "project" stored everywhere in logs based on key used. See: GfLogStore for instance as well as controllers code. What are we going to store for requests that uses auto generated keys? 

Ideally it would be nice to store model/app/assistant name.

src/main/java/com/epam/aidial/core/Proxy.java
```diff
@@ -119,7 +119,7 @@ private void handleRequest(HttpServerRequest request) {
         } else if (apiKey != null) {
             key = config.getKeys().get(apiKey);
 
-            if (key == null) {
+            if (key == null && !config.existDeploymentApiKey(apiKey)) {
```
avavilau: We are changing essential authorization part. Would be nice to see at least some unit tests.

avavilau (dismissed):
Essential parts are missing

src/main/java/com/epam/aidial/core/config/Deployment.java
astsiapanay: no, it's not by security reasons.

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/Proxy.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/ProxyContext.java
```diff
@@ -81,6 +84,12 @@ public Future<Void> respond(HttpStatus status, String body) {
     }
 
     public String getProject() {
-        return key == null ? null : key.getProject();
+        if (key != null) {
+            return key.getProject();
+        }
+        if (originator != null) {
+            return originator.getName();
+        }
```
avavilau: Do  we really need both? What if we have both, why do we prefer originator? Having data object which can represent impossible state is a clear antipattern.

astsiapanay: There is just one choice: either project key or deployment originator. However it's technically possible due to key collision.

astsiapanay: fixed

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
```diff
@@ -98,6 +103,16 @@ private void load(boolean fail) {
...
     }
 
+    private void associateDeploymentWithApiKey(Config config, Deployment deployment) {
+        if (deployment.getApiKey() == null) {
+            deployment.setApiKey(generateKey());
+        }
+        Key key = new Key();
+        key.setKey(deployment.getApiKey());
+        key.setProject(deployment.getName());
+        config.getKeys().put(deployment.getApiKey(), key);
```
artsiomkorzun: Please assert that keys are never overwritten.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -163,12 +163,14 @@ private void handleProxyRequest(HttpClientRequest proxyRequest) {
         if (!deployment.isForwardApiKey()) {
```
artsiomkorzun: Why do you change the original request?  Is it better to always add api-key header to excluded headers to not copy it and do not modify the original request.

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -20,7 +20,7 @@ public class RateLimiter {
 
     public void increase(ProxyContext context) {
         Key key = context.getKey();
-        if (key == null) {
+        if (key == null || key.getRole() == null) {
```
artsiomkorzun, line 23: Optionally: you can still track the usage for the future config update like:
1. Key X does not have a role.
2. Config change.
3. Key X now has a role.

src/test/java/com/epam/aidial/core/ProxyTest.java
```diff
@@ -184,6 +191,37 @@ public void testHandle_UnknownApiKey() {
...
+        LinkedHashMap<String, Route> routes = new LinkedHashMap<>();
+        routes.put("route", route);
+        config.setRoutes(routes);
+        when(configStore.load()).thenReturn(config);
+
+        when(accessTokenValidator.extractClaims(any())).thenReturn(Future.succeededFuture(IdentityProvider.CLAIMS_WITH_EMPTY_ROLES));
+
+        proxy.handle(request);
+
+        verify(response).setStatusCode(OK.getCode());
```
avavilau, line 222: So the only things we are testing is that we are getting OK? 

Is there any testing for the main functionality of this PR? That app/adapter actually get generated keys which are different to the key request was originally authorized with? And that they actually can use these generated keys?

astsiapanay: there is already a such test com.epam.aidial.core.ProxyTest#testHandle_UnknownApiKey

astsiapanay: I just add a new test with the success path.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
astsiapanay: fixed

astsiapanay: It had better generate a new one.

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
astsiapanay: I guess you want to keep tracking usage of tokens. I think the role is almost never changed.
The apps, models and assistants are not tracked and won't be in the future.
Let's keep as is then.

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
astsiapanay: I guess you want to keep tracking usage of tokens. I think the role is almost never changed.
The apps, models and assistants are not tracked and won't be in the future.
Let's keep as is then.

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
astsiapanay: fixed

src/test/java/com/epam/aidial/core/ProxyTest.java
astsiapanay: fixed

artsiomkorzun (approved):

Issue #86, opened 2023-12-14
Handle OpenAI request with API and Auth headers


astsiapanay:
OpenAI client injects auth header with the same value as API key even a caller doesn't set Auth header explicitly.
If both API key and Auth header has the same value treat API key has a higher priority and ignore auth header

PR #87, opened 2023-12-15
fix: Handle OpenAI request with API and Auth headers #86

astsiapanay:


src/main/java/com/epam/aidial/core/Proxy.java
```diff
@@ -114,8 +116,18 @@ private void handleRequest(HttpServerRequest request) {
             respond(request, HttpStatus.UNAUTHORIZED, "At least API-KEY or Authorization header must be provided");
             return;
         } else if (apiKey != null && authorization != null) {
-            respond(request, HttpStatus.BAD_REQUEST, "Either API-KEY or Authorization header must be provided but not both");
-            return;
+            // Special case handling. OpenAI client sends both API key and Auth headers even if a caller sets just API Key only
```
Maxim-Gadalov: I would merge inner if statement as following
```java
// if both provided and it's not a openAI case
} else if (apiKey != null && authorization != null && !apiKey.equals(extractTokenFromHeader(authorization))) {
            respond(request, HttpStatus.BAD_REQUEST, "Either API-KEY or Authorization header must be provided but not both");
            return;
 } else if (apiKey != null) {
            key = config.getKeys().get(apiKey);

            if (key == null) {
                respond(request, HttpStatus.UNAUTHORIZED, "Unknown api key");
                return;
            }
        } else {
            key = null;
        }
```

astsiapanay: fixed

Maxim-Gadalov (approved):

PR #89, opened 2023-12-15
fix: API key is not sent to Base Assistant #88

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -168,6 +172,10 @@ private void handleProxyRequest(HttpClientRequest proxyRequest) {
         ProxyUtil.copyHeaders(request.headers(), proxyRequest.headers(), excludeHeaders);
 
         if (deployment.isForwardApiKey()) {
+            if (isBaseAssistant(deployment)) {
+                // base assistant is created in runtime and doesn't have assigned api key
+                associateDeploymentWithApiKey(context.getConfig(), deployment);
```
artsiomkorzun: You concurrently update a not-concurrent map. You had better to create it when the app is initialized and use later. You can put it into map each time when config is reloaded.

artsiomkorzun (dismissed):

artsiomkorzun (approved):

PR #96, opened 2023-12-18
fix: route Files API by encoded path, url parameter in metadata response is URL encoded

Maxim-Gadalov:
1. Core should urldecode path segments, not the full path (as https://github.com/epam/ai-dial-core/blob/development/src/main/java/com/epam/aidial/core/controller/ControllerSelector.java#L47C1-L47C1)
2. url in the metadata should be correct url. No extra encoding or decoding is required to access it.
3. name and parentPath in metadata a NOT parts on the url. They should be encoded when building url (as per URL RFC).
4. Symbol / is reserved as a path segments separator and not allowed in file names and directory names. Requests with / in a file name or directory name will be 400 Bad request. Following requests will end up with 400:
PUT v1/files/a/b/c%2Fd
GET v1/files/a/b/c%2Fd
GET v1/files/metadata/a/b/c%2Fd

### DIFF
--- a/src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
+++ b/src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
@@ -1,13 +1,11 @@
 package com.epam.aidial.core.storage;
 
+import com.epam.aidial.core.util.UrlUtil;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import org.apache.commons.lang3.StringUtils;
 
-import java.net.URLDecoder;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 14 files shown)


src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -60,31 +66,56 @@ public String getParentPath() {
...
         String name = "/";
-        if (elements.length > 0) {
-            name = elements[elements.length - 1];
+        if (!elements.isEmpty()) {
+            name = elements.get(elements.size() - 1);
         }
-        if (elements.length > 1) {
-            String parentPath = relativeFilePath.substring(0, relativeFilePath.length() - name.length() - 1);
+        if (elements.size() > 1) {
+            String parentPath = urlEncodedRelativePath.substring(0, urlEncodedRelativePath.length() - name.length() - 1);
```
artsiomkorzun: Name is decoded segment which can have different size. Add an unit test for parsing. For example, /x/y/%1Fz.txt

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -60,31 +66,56 @@ public String getParentPath() {
...
             if (!parentPath.isEmpty() && !parentPath.equals(BlobStorageUtil.PATH_SEPARATOR)) {
                 parentFolders = List.of(parentPath.split(BlobStorageUtil.PATH_SEPARATOR));
```
artsiomkorzun: use encodedElements to take parent path

src/main/java/com/epam/aidial/core/util/UrlUtil.java
```diff
@@ -0,0 +1,28 @@
...
+            return uri.toASCIIString();
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public String decodePath(String path) {
+        try {
+            URI uri = new URI(path);
+            return uri.getPath();
```
artsiomkorzun, line 26: Check that rawFragment and rawQuery is null.

src/test/java/com/epam/aidial/core/util/UrlUtilTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    public void testPathEncoding() {
+        assertEquals("folder%201", UrlUtil.encodePath("folder 1"));
+        assertEquals("folder%20%23", UrlUtil.encodePath("folder #"));
+        assertEquals("fo$l+=d,e%23r%201", UrlUtil.encodePath("fo$l+=d,e#r 1"));
+    }
+
+    @Test
+    public void testPathDecoding() {
+        assertEquals("folder 1", UrlUtil.decodePath("folder%201"));
```
artsiomkorzun, line 20: Add tests for /folder1/file#5.txt
Add tests for /folder1/q?file=5.txt

artsiomkorzun (approved):

Issue #110, opened 2024-01-05
Core rejects requests if trace id is not found


astsiapanay:
The team agreed the Core to accept requests with trace ID even if the trace is not found in the Core.

The Core registers trace if id is not found and proceeds with the existing logic.
The Core should remove trace once the request is completed and the trace is not found before.

PR #111, opened 2024-01-05
fix: Core rejects requests if trace id is not found #110

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -57,6 +57,7 @@ public class DeploymentPostController {
 
     private final Proxy proxy;
     private final ProxyContext context;
+    private boolean unregisterTrace;
```
artsiomkorzun, line 60: Please save trace id instead of flag and use it in unregister.

astsiapanay: fixed

artsiomkorzun (approved):

PR #112, opened 2024-01-05
feat: support resources

artsiomkorzun:


src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -45,23 +52,27 @@
 import java.util.concurrent.TimeUnit;
 
 @Slf4j
+@Setter
+@Getter
 public class AiDial {
 
     private JsonObject settings;
+    private JsonObject extraSettings = new JsonObject();
```
Maxim-Gadalov, line 60: Why do you need it? you can put test configs directly to the file

src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -81,8 +92,17 @@ void start() throws Exception {
             }
             EncryptionService encryptionService = new EncryptionService(Json.decodeValue(settings("encryption").toBuffer(), Encryption.class));
             TokenStatsTracker tokenStatsTracker = new TokenStatsTracker();
-            proxy = new Proxy(vertx, client, configStore, logStore, rateLimiter, upstreamBalancer, accessTokenValidator,
-                    storage, encryptionService, apiKeyStore, tokenStatsTracker);
+
+            redis = openRedis();
```
Maxim-Gadalov, line 96: resource service - mandatory core component, we should fail if some configs are missing

src/main/java/com/epam/aidial/core/controller/AccessControlBaseController.java
```diff
@@ -13,22 +13,21 @@
...
     final boolean checkFullAccess;
 
-
     /**
      * @param bucket url encoded bucket name
-     * @param filePath url encoded file path
+     * @param path url encoded resource path
      */
-    public Future<?> handle(String bucket, String filePath) {
+    public Future<?> handle(String folder, String bucket, String path) {
```
Maxim-Gadalov: ResourceType instead of folder?

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
```diff
@@ -39,6 +40,9 @@ public class ControllerSelector {
 
     private static final Pattern PATTERN_FILES_METADATA = Pattern.compile("/v1/metadata/files/([a-zA-Z0-9]+)/(.*)");
 
+    private static final Pattern PATTERN_RESOURCE = Pattern.compile("/v1/(conversations|prompts)/([a-zA-Z0-9]+)/(.*)");
```
Maxim-Gadalov, line 43: why not merge with files pattern? conversations, prompts, files - all resources

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
```diff
@@ -135,15 +139,33 @@ private static Controller selectGet(Proxy proxy, ProxyContext context, String pa
...
         match = match(PATTERN_FILES, path);
         if (match != null) {
             String bucket = match.group(1);
             String filePath = match.group(2);
             DownloadFileController controller = new DownloadFileController(proxy, context);
-            return () -> controller.handle(bucket, filePath);
+            return () -> controller.handle("files", bucket, filePath);
+        }
+
+        match = match(PATTERN_RESOURCE, path);
```
Maxim-Gadalov, line 153: route is not consistent.
selectGet
selectPost
selectPut
selectDelete

in each select we pick a controller based on pattern
For single responsibility each controller related to a single action (Get element, or collection)

You implemented a Controller that handles all HTTP verbs internally and for my opinion it's harder to read.

We should either implement is the same way like we did for Files or refactor Files API to be consistent with Resources

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -0,0 +1,147 @@
...
+        this.metadata = metadata;
+    }
+
+    @Override
+    protected Future<?> handle(ResourceDescription descriptor) {
+        if (context.getRequest().method() == HttpMethod.GET) {
+            return metadata ? getMetadata(descriptor) : getResource(descriptor);
+        }
+
+        if (context.getRequest().method() == HttpMethod.PUT) {
```
Maxim-Gadalov, line 38: `method = context.getRequest().method();`

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -0,0 +1,147 @@
...
+            return putResource(descriptor);
+        }
+
+        if (context.getRequest().method() == HttpMethod.DELETE) {
+            return deleteResource(descriptor);
+        }
+
+        return context.respond(HttpStatus.BAD_GATEWAY);
```
Maxim-Gadalov: context.respond(HttpStatus.BAD_GATEWAY, "No route");

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private Future<?> putResource(ResourceDescription descriptor) {
+        if (descriptor.isFolder()) {
+            return context.respond(HttpStatus.BAD_REQUEST, "Folder not allowed: " + descriptor.getUrl());
+        }
+
+        int contentLength = ProxyUtil.contentLength(context.getRequest(), 0);
+        int contentLimit = service.getMaxSize();
+
+        if (contentLength > contentLimit) {
```
Maxim-Gadalov, line 104: Should we just bound `maxSize` to the `maxRequestSize`? see `Proxy.handleRequest`

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -0,0 +1,147 @@
...
+        return context.getRequest().body().compose(bytes -> {
+                    if (bytes.length() > contentLimit) {
+                        String message = "Resource size: %s exceeds max limit: %s".formatted(bytes.length(), contentLimit);
+                        throw new HttpException(HttpStatus.REQUEST_ENTITY_TOO_LARGE, message);
+                    }
+
+                    String body = bytes.toString(StandardCharsets.UTF_8);
+                    return vertx.executeBlocking(() -> service.putResource(descriptor, body));
+                })
+                .onSuccess((ok) -> context.respond(HttpStatus.OK))
```
Maxim-Gadalov, line 118: we should return a metadata of saved resource

src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -0,0 +1,401 @@
...
+    private ResourceFolderMetadata getFolderMetadata(ResourceDescription descriptor, String token, int limit) {
+        String blobKey = blobKey(descriptor);
+        PageSet<? extends StorageMetadata> set = blobStore.list(blobKey, token, limit);
+
+        if (set.isEmpty()) {
+            return null;
+        }
+
+        List<MetadataBase> resources = set.stream().map(meta -> {
+            Map<String, String> metadata = meta.getUserMetadata();
```
Maxim-Gadalov, line 113: Some blob storage provides do not provide user-metadata without additional API call, see `ListOptions.withDetails()`

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
```diff
@@ -161,6 +186,19 @@ public MetadataBase listMetadata(ResourceDescription resource) {
         }
     }
 
+    public PageSet<? extends StorageMetadata> list(String absoluteFilePath, String afterMarker, int maxResults) {
```
Maxim-Gadalov, line 189: let's consolidate with `listMetadata`

src/main/java/com/epam/aidial/core/AiDial.java
artsiomkorzun: This way most of the tests will fail. Let's postpone enforcing always redis available.

src/main/java/com/epam/aidial/core/controller/AccessControlBaseController.java
artsiomkorzun: Renamed.

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
artsiomkorzun: It is badly mergeable. Redis is not used with files.

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
artsiomkorzun: Resource controller is responsible for handling whole API related to resources. In the same way as DeploymentController handles listing and getting a specific deployment.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: added the body

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: Not really, we want to limit conversation/prompt size. The request body size depends on specific API.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: Debatable.

src/main/java/com/epam/aidial/core/storage/BlobStorage.java
artsiomkorzun: Let's discuss.

src/main/java/com/epam/aidial/core/AiDial.java
artsiomkorzun: Reworked

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: Added.

Issue #117, opened 2024-01-08
Calculate token usage statistics using call stack


astsiapanay:
Calculate token usage statistics for applications/assistants using call stack.

PR #118, opened 2024-01-08
feat: Calculate token usage statistics using call stack #117

astsiapanay:


src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -129,11 +147,72 @@ public String toString() {
...
+        public synchronized void link(String parentSpanId, String childSpanId) {
+            CallInfo callInfo = spanIdToCallInfo.get(parentSpanId);
+            if (callInfo == null) {
+                log.warn("Parent span is not found by id: {}", parentSpanId);
+                return;
+            }
+            callInfo.childSpanIds.add(childSpanId);
+        }
+
+        public synchronized void setTokeUsage(String spanId, TokenUsage tokenUsage) {
```
artsiomkorzun, line 179: Please correct type: setToke**n**Usage

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -81,22 +85,36 @@ public RateLimitResult limit(ProxyContext context) {
...
         }
-        return entity != null;
+        entity.addCall(context.getCurrentSpanId());
+        return result;
+    }
+
+    public void calculateTokenUsage(ProxyContext context) {
+        Entity entity = getEntityFromTracingContext(context);
+        if (entity == null) {
+            log.warn("Entity is not found by traceId={}", context.getTraceId());
```
artsiomkorzun, line 114: Please add prefix to locate the problem more easily: "Failed to calculate token usage"

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -129,11 +147,72 @@ public String toString() {
...
+        }
+
+        public synchronized void addCall(String spanId) {
+            spanIdToCallInfo.put(spanId, new CallInfo());
+        }
+
+        public synchronized void link(String parentSpanId, String childSpanId) {
+            CallInfo callInfo = spanIdToCallInfo.get(parentSpanId);
+            if (callInfo == null) {
+                log.warn("Parent span is not found by id: {}", parentSpanId);
```
artsiomkorzun, line 173: Please improve logging in this class.

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -81,22 +85,36 @@ public RateLimitResult limit(ProxyContext context) {
...
+        return result;
+    }
+
+    public void calculateTokenUsage(ProxyContext context) {
+        Entity entity = getEntityFromTracingContext(context);
+        if (entity == null) {
+            log.warn("Entity is not found by traceId={}", context.getTraceId());
+            return;
+        }
+        context.setTokenUsage(entity.calculate(context.getCurrentSpanId()));
```
artsiomkorzun, line 117: Is it better to do it only if it is missing? Some applications/models can compute it more precisely.

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -129,11 +147,72 @@ public String toString() {
...
 
-    private record Entity(String id, List<String> roles, String name, boolean user) {
+    @Data
+    private static class Entity {
+
+        private final String id;
+        private final List<String> roles;
+        private final String name;
+        private final boolean user;
+        private final Map<String, CallInfo> spanIdToCallInfo = new HashMap<>();
```
artsiomkorzun, line 157: Please move the logic related to token tracking to a separate class. e.g. TokenTracker

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
astsiapanay: We can rely on applications/assistants any longer to compute statistics

astsiapanay:
no longer valid since the core is moved to per request key authorization

PR #145, opened 2024-01-16
feat: Calculate token usage statistics using call stack #117

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -288,38 +298,51 @@ private void handleResponse() {
...
-
-            if (tokenUsage == null) {
-                log.warn("Can't find token usage. Trace: {}. Span: {}. Key: {}. Deployment: {}. Endpoint: {}. Upstream: {}. Status: {}. Length: {}",
-                        context.getTraceId(), context.getSpanId(),
-                        context.getProject(), context.getDeployment().getName(),
-                        context.getDeployment().getEndpoint(),
-                        context.getUpstreamRoute().get().getEndpoint(),
-                        context.getResponse().getStatusCode(),
-                        context.getResponseBody().length());
+        Future<TokenUsage> tokenUsageFuture;
```
artsiomkorzun: Initialize with Future.succeededFuture() to save on code lines.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -288,38 +298,51 @@ private void handleResponse() {
...
+                            context.getResponseBody().length());
+                } else {
+                    Model model = (Model) context.getDeployment();
+                    tokenUsage.calculateCost(model.getPricing());
+                }
+            } else {
+                tokenUsageFuture = Future.succeededFuture();
             }
+        } else {
+            tokenUsageFuture = proxy.getTokenStatsTracker().getTokenStats(context).andThen(result -> context.setTokenUsage(result.result()));
```
artsiomkorzun, line 318: Should you check for status to be 200/OK?

src/main/java/com/epam/aidial/core/log/GfLogStore.java
```diff
@@ -95,7 +96,27 @@ private void append(ProxyContext context, LogEntry entry) {
...
+            append(entry, Long.toString(tokenUsage.getCompletionTokens()), true);
+            append(entry, ",\"prompt_tokens\":", false);
+            append(entry, Long.toString(tokenUsage.getPromptTokens()), true);
+            append(entry, ",\"total_tokens\":", false);
+            append(entry, Long.toString(tokenUsage.getTotalTokens()), true);
+            if (tokenUsage.getCost() != null) {
+                append(entry, ",\"cost\":", false);
+                append(entry, tokenUsage.getCost().toString(), true);
+            }
+            if (tokenUsage.getAggCost() != null) {
```
artsiomkorzun: maybe totalCost is better?

src/main/java/com/epam/aidial/core/token/TokenUsage.java
```diff
@@ -1,12 +1,49 @@
...
+        if (pricing == null) {
+            return;
+        }
+        String unit = pricing.getUnit();
+        if (!"token".equals(unit)) {
+            return;
+        }
+        double cost = 0.0;
+        if (pricing.getPrompt() != null) {
+            cost += promptTokens * Double.parseDouble(pricing.getPrompt());
```
artsiomkorzun, line 24: Can throw an exception, so the calling code should take care of this :(

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/log/GfLogStore.java
astsiapanay: We agreed with @avavilau to name this kind of costs as aggregated. If you have objections feel free to discuss that with the team.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: If application returns http status code different than 200 we shouldn't ignore costs associated with the app anyway.

Supposing app A requests model A. Model A spent some tokens and returned response to the App. The app A failed request in the middle for some reason.

At this point the app A has already spent some tokens and we should report them even if the response is not 200.

Agree?

src/main/java/com/epam/aidial/core/token/TokenUsage.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -238,6 +238,8 @@ void handleProxyRequest(HttpClientRequest proxyRequest) {
         proxyRequest.putHeader(HttpHeaders.CONTENT_LENGTH, Integer.toString(requestBody.length()));
         context.getRequestHeaders().forEach(proxyRequest::putHeader);
 
+        proxy.getTokenStatsTracker().startSpan(context);
```
artsiomkorzun: Start a span when a request is received. This method is reentrant.

artsiomkorzun (dismissed):

artsiomkorzun (approved):

PR #147, opened 2024-01-16
fix: refactor Files metadata to generic resource metadata

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/data/ResourceMetadataCollection.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+@NoArgsConstructor
+public class ResourceMetadataCollection extends MetadataBase {
```
artsiomkorzun: I would use it be *Folder to match with NodeType enum.

src/main/java/com/epam/aidial/core/data/ResourceType.java
```diff
@@ -0,0 +1,15 @@
+package com.epam.aidial.core.data;
+
+import lombok.Getter;
+
+@Getter
+public enum ResourceType {
+    FILE("files");
+
+    private final String resourceGroup;
```
artsiomkorzun: nit: folder? or just group.

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
```diff
@@ -21,7 +21,7 @@ protected Future<?> handle(ResourceDescription resource) {
         BlobStorage storage = proxy.getStorage();
```
artsiomkorzun, line 21: Are you going to change metadata API in this pr?

Maxim-Gadalov: I don't have other resources to deal with right now. Let's implement is later

artsiomkorzun: The frontend will be able to migrate on these change as soon as possible.

Maxim-Gadalov: I think I misunderstood your point. We should fix metadata routing in this PR

artsiomkorzun (approved):

PR #148, opened 2024-01-16
fix: retry when client fails to send response header

artsiomkorzun:


src/main/java/com/epam/aidial/core/upstream/UpstreamBalancer.java
```diff
@@ -9,18 +9,20 @@
 
 public class UpstreamBalancer {
 
+    private static final int CONNECTION_ERROR_EXTRA_ATTEMPTS = 2;
```
astsiapanay: Is this connection error extra attempts per upstream or in total?

src/main/java/com/epam/aidial/core/upstream/UpstreamBalancer.java
```diff
@@ -9,18 +9,20 @@
...
-            return new UpstreamRoute(upstreams, null, 0);
+        if (upstreams.size() > 1) {
+            AtomicLong counter = counters.computeIfAbsent(name, (key) -> new AtomicLong());
+            offset = (int) (counter.getAndIncrement() % upstreams.size());
         }
 
-        AtomicLong counter = counters.computeIfAbsent(name, (key) -> new AtomicLong());
-        int offset = (int) (counter.getAndIncrement() % upstreams.size());
-        return new UpstreamRoute(upstreams, counter, offset);
+        return new UpstreamRoute(upstreams, offset, upstreams.size() + CONNECTION_ERROR_EXTRA_ATTEMPTS);
```
astsiapanay: If the upstream size is 3 you will make 5 attempts instead of 2 on connection reset error.
May be put just a single value but not a sum?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -6,25 +6,26 @@
...
     private final List<Upstream> upstreams;
-    private final AtomicLong counter;
     private final int offset;
+    private final int maxAttempts;
 
-    private Upstream current;
-    private int count;
+    private Upstream upstream;
+    private int attempts;
+    private int next;
```
astsiapanay, line 22: We can use a single field `count` at it was before.
```
int index = (offset + count++) % upstreams.size();
current = upstreams.get(index);
```
on retry
```
count--; // we want to keep the previous upstream
```

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -6,25 +6,26 @@
...
+    private int prev;
 
     public int attempts() {
-        return count;
+        return attempts;
```
astsiapanay: the field accumulates all attempts made on all upstreams.
I think we should track attempts per upstream like:
```
  return attempts[(count + offset) % upstreams.size()]
```

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -6,25 +6,26 @@
...
     }
 
     @Override
     public boolean hasNext() {
-        return count < upstreams.size();
+        return next < upstreams.size() && attempts < maxAttempts;
```
astsiapanay:
```
return count < upstreams.size() && attempts[(count + offset) % upstreams.size()] < maxAttempts
```

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -6,25 +6,26 @@
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicLong;
 
 @RequiredArgsConstructor
 public class UpstreamRoute implements Iterator<Upstream> {
```
astsiapanay, line 11: Please add javadoc for the fields. They require some explanation.

astsiapanay (changes requested):
Please add Unit tests to cover your changes.

src/main/java/com/epam/aidial/core/upstream/UpstreamBalancer.java
artsiomkorzun: In total

src/main/java/com/epam/aidial/core/upstream/UpstreamBalancer.java
artsiomkorzun: I will make 5 instead of 3. How does it help?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: I thought of this. Count-- is not safe against api misuse. Plus you still need to track the total number of attempts. Because count is really index in this case.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: We can track, but why?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: What you are trying to solve?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: Added.

src/main/java/com/epam/aidial/core/upstream/UpstreamBalancer.java
astsiapanay: I thought `CONNECTION_ERROR_EXTRA_ATTEMPTS` is per upstream. You want to calculate a total allowed attempts. That's correct.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
astsiapanay: I was thinking about the concept we want to track the number of retries per upstream but not in total.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
astsiapanay: You still use `prev` and `next` and `prev` == `next` - 1. You track the total number of attempts with `attempts`.
Having two fields seems to be redundant.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
astsiapanay: checking the number of retries per upstream. But since we want to do that I discard the comment.

astsiapanay (dismissed):

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: That is to prevent misusage when someone calls this method twice.

astsiapanay (approved):

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
astsiapanay: ok that makes sense.

Issue #154, opened 2024-01-18
Calculate cost for models with "char without space"


astsiapanay:
See reference at 
https://github.com/epam/ai-dial-analytics-realtime/blob/release-0.6/aidial_analytics_realtime/rates.py#L32

PR #155, opened 2024-01-18
feat: Calculate cost for models with "char without space" #154

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -297,6 +298,11 @@ void handleResponse() {
                 context.setTokenUsage(tokenUsage);
                 proxy.getRateLimiter().increase(context);
                 tokenUsageFuture = Future.succeededFuture(tokenUsage);
+                try {
+                    tokenUsage.setCost(ModelCostCalculator.calculate(context));
+                } catch (Throwable e) {
+                    log.warn("Failed to calculate cost for model={}", context.getDeployment().getName());
```
artsiomkorzun: Add trace and span.

src/main/java/com/epam/aidial/core/util/ModelCostCalculator.java
```diff
@@ -0,0 +1,133 @@
...
+            return cost;
+        }
+        return null;
+    }
+
+    private static int getResponseContentLength(ModelType modelType, Buffer responseBody) {
+        if (modelType == ModelType.EMBEDDING) {
+            return 0;
+        }
+        try (InputStream stream = new ByteBufInputStream(responseBody.getByteBuf())) {
```
artsiomkorzun: How does it work with streaming?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/util/ModelCostCalculator.java
astsiapanay: `DeploymentPostController` sets responseBody to a buffer when the stream is completed.
see com.epam.aidial.core.controller.DeploymentPostController#handleProxyResponse

astsiapanay: fixed

artsiomkorzun (approved):

PR #158, opened 2024-01-19
fix: Model cost calculation fails if token usage stats is not found in the response #157

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -292,9 +293,22 @@ void handleResponse() {
         context.setResponseBody(responseBody);
         context.setResponseBodyTimestamp(System.currentTimeMillis());
         Future<TokenUsage> tokenUsageFuture = Future.succeededFuture();
-        if (context.getDeployment() instanceof Model) {
+        if (context.getDeployment() instanceof Model model) {
             if (context.getResponse().getStatusCode() == HttpStatus.OK.getCode()) {
                 TokenUsage tokenUsage = TokenUsageParser.parse(responseBody);
+                if (tokenUsage == null) {
+                    Pricing pricing = model.getPricing();
```
artsiomkorzun, line 300: Why not always warn about missing token usage?

astsiapanay: there might be a model where cost is not calculated by token, e.g. by content lengths of response and requests. So the token usage is always null and we shouldn't log warning message in this case.

artsiomkorzun (approved):

Issue #168, opened 2024-01-25
Print execution path as JSON array to prompt logs


astsiapanay:
1. I suggest using "/" instead of "->", because it is more standard path separator.
 * It is a single character
 * It does not add a new rules on what characters can or cannot be used in deployment names, because "/" is already used in urls paths.
2. I suggest using string in influx tag (because it can only be string), but using an array of strings in json log from Core. I.e. do not glue it into string before it is necessary to keep it more machine-parsable.

PR #169, opened 2024-01-25
feat: Print execution path as JSON array to prompt logs #168

astsiapanay:


src/main/java/com/epam/aidial/core/log/GfLogStore.java
```diff
@@ -96,11 +99,10 @@ private void append(ProxyContext context, LogEntry entry) {
...
         }
 
-        String executionPath = context.getExecutionPath();
+        List<String> executionPath = context.getExecutionPath();
         if (executionPath != null) {
-            append(entry, ",\"execution_path\":\"", false);
-            append(entry, executionPath, true);
-            append(entry, "\"", false);
+            append(entry, ",\"execution_path\":", false);
+            append(entry, ProxyUtil.MAPPER.writeValueAsString(executionPath), true);
```
artsiomkorzun: do you want to see a json array in the log? append works only for a string

artsiomkorzun: ["x", "y"] -> willl be encoded as [""x"", ""y""]

astsiapanay: fixed

artsiomkorzun: ["x"y"] -> ["x"y"] - incorrect
["x"y"] -> ["x\"y"] - correct

You need to escape each element in a list.

artsiomkorzun (dismissed):

artsiomkorzun (approved):

PR #186, opened 2024-01-31
chore: clean up settings/config files

### LINKED ISSUE #84: Cleanup configuration file
Core [configuration file](https://github.com/epam/ai-dial-core/blob/development/config/aidial.config.json) is filled with some test data and containerized alongside the application.

The current file content neither represents full configuration nor declares out-of-the-box defaults.\
Please consider one of the following:
- review & keep the file with empty values so it can be referenced as a configuration scheme
- exclude it from the container

artsiomkorzun:
Fixes https://github.com/epam/ai-dial-core/issues/84


sample/aidial.settings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "config": {
```
astsiapanay, line 2: I would add a full set of settings including IdentityProvider, encryption, blob storage.
So the user should be aware of what settings are available.

artsiomkorzun: Please collaborate on IdentiyProvider.

astsiapanay (approved):

PR #204, opened 2024-02-12
feat: added seed, tools, system_prompt features and display_version flag to the listing

### LINKED ISSUE #163: Extend deployment listing with new fields (Feature flags: system_prompt, tools, seed) and DisplayVersion

adubovik:


src/main/java/com/epam/aidial/core/config/FileConfigStore.java
```diff
@@ -127,7 +127,7 @@ private static InputStream openStream(String path) {
         }
     }
 
-    private static void setMissingFeatures(Deployment model, Features features) {
+    private static void inheritMissingFeatures(Deployment model, Features features) {
```
astsiapanay: The name of the first argument confuses a bit: there is a base class but the name is model. Moreover it's always Assistant.

src/main/java/com/epam/aidial/core/controller/DeploymentController.java
```diff
@@ -100,13 +100,24 @@ private static DeploymentData createDeployment(Model model) {
...
         data.setTruncatePrompt(features.getTruncatePromptEndpoint() != null);
+
+        data.setSystemPrompt(booleanOrDefault(features.getSystemPromptSupported(), true));
+        data.setTools(booleanOrDefault(features.getToolsSupported(), false));
+        data.setSeed(booleanOrDefault(features.getSeedSupported(), false));
+
         return data;
     }
+
+    public static boolean booleanOrDefault(Boolean obj, boolean def) {
```
astsiapanay: Nit. Instead of introducing a new method we can update those fields conditionally.

adubovik: Fixed

src/main/java/com/epam/aidial/core/config/FileConfigStore.java
adubovik: Renamed back to `setMissingFeatures` to avoid confusion

astsiapanay (approved):

PR #209, opened 2024-02-13
fix: clean up shared access on resource deletion

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/controller/DeleteFileController.java
```diff
@@ -27,6 +36,13 @@ protected Future<?> handle(ResourceDescription resource) {
         Future<Void> result = proxy.getVertx().executeBlocking(() -> {
             try {
                 storage.delete(absoluteFilePath);
+                // clean shared access
+                // TODO remove check when redis become mandatory
+                if (shareService != null) {
```
artsiomkorzun, line 40: move it before deleting a file

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -148,7 +155,21 @@ private Future<?> deleteResource(ResourceDescription descriptor) {
...
         }
 
-        return vertx.executeBlocking(() -> service.deleteResource(descriptor))
+        return vertx.executeBlocking(() -> {
+                    boolean isDeleted = service.deleteResource(descriptor);
+                    if (isDeleted) {
+                        // clean shared access
+                        // TODO remove check when redis become mandatory
+                        if (shareService != null) {
+                            Set<ResourceLink> resourceLinks = new HashSet<>();
```
artsiomkorzun: move it up and does not check for share service to be null

artsiomkorzun (approved):

PR #212, opened 2024-02-14
fix: NPE when using revoke/discard on non-shared resource

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -229,7 +229,9 @@ public void computeResource(ResourceDescription descriptor, Function<String, Str
         try (var ignore = lockService.lock(redisKey)) {
             String body = getResource(descriptor, false);
             String updatedBody = fn.apply(body);
-            putResource(descriptor, updatedBody, true, false);
+            if (updatedBody != null) {
```
astsiapanay: we do we not update a resource with nullable body?

src/main/java/com/epam/aidial/core/service/ShareService.java
```diff
@@ -268,7 +273,12 @@ public void discardSharedAccess(String bucket, String location, ResourceLinkColl
         for (ResourceLink link : resourceLinks) {
             ResourceDescription resource = getResourceFromLink(link.url());
             ResourceType resourceType = resource.getType();
-            removeSharedResource(bucket, location, link, resourceType);
+            try {
+                removeSharedResource(bucket, location, link, resourceType);
+            } catch (Throwable e) {
+                e.printStackTrace();
```
astsiapanay: Is this done intentionally?

Maxim-Gadalov: no, forgot to remove, thanks

src/main/java/com/epam/aidial/core/service/ResourceService.java
Maxim-Gadalov: putResource expect non-null result

src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -229,7 +229,9 @@ public void computeResource(ResourceDescription descriptor, Function<String, Str
...
+                putResource(descriptor, updatedBody, true, false);
```
artsiomkorzun: If it is null, you had better to delete it. Similar to Map.compute contract.

astsiapanay (dismissed):

src/main/java/com/epam/aidial/core/service/ResourceService.java
astsiapanay: yes, makes sense.

artsiomkorzun: Otherwise it is expected to fail

Maxim-Gadalov: in most current cases null means - resource absent, we don't need to delete it. Let's not rush in this direction right away.
Added resource body check to avoid redundant updates

artsiomkorzun (approved):

PR #230, opened 2024-02-20
fix: normalize URL for comparison

Maxim-Gadalov:
#224

### DIFF
--- a/src/main/java/com/epam/aidial/core/util/ProxyUtil.java
+++ b/src/main/java/com/epam/aidial/core/util/ProxyUtil.java
@@ -2,6 +2,8 @@ package com.epam.aidial.core.util;
 
 import com.epam.aidial.core.Proxy;
 import com.epam.aidial.core.config.ApiKeyData;
+import com.epam.aidial.core.security.EncryptionService;
+import com.epam.aidial.core.storage.ResourceDescription;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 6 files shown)


src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -102,7 +104,14 @@ public static void collectAttachedFiles(ObjectNode tree, ApiKeyData apiKeyData)
...
                 JsonNode url = attachment.get("url");
                 if (url != null) {
-                    apiKeyData.getAttachedFiles().add(url.textValue());
+                    try {
+                        String urlValue = url.textValue();
+                        // build
+                        ResourceDescription resource = ResourceDescription.fromLink(urlValue, encryptionService);
+                        apiKeyData.getAttachedFiles().add(resource.getUrl());
+                    } catch (Exception e) {
+                        // ignore, probably a public link
```
astsiapanay, line 115: We should check if url is public then skip it. OR at least log a warning message for the further analysis.

astsiapanay: How would we process links like that?

- `http://google.com/revenew-2023.txt`
- `/<bucket>/fies/file1.txt`
- `http://dial.epam.com/v1/files/<bucket>/file1.txt`

Is that possible?

Maxim-Gadalov: added log message

astsiapanay (approved):

PR #231, opened 2024-02-20
fix: verify caller has access to attached files

astsiapanay:


src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -102,7 +106,11 @@ public static void collectAttachedFiles(ObjectNode tree, ApiKeyData apiKeyData)
                 JsonNode attachment = attachments.get(j);
                 JsonNode url = attachment.get("url");
                 if (url != null) {
-                    apiKeyData.getAttachedFiles().add(url.textValue());
+                    if (checkAccessFn.apply(url.textValue())) {
+                        apiKeyData.getAttachedFiles().add(url.textValue());
+                    } else {
+                        throw new HttpException(FORBIDDEN, "Access denied to the file %s");
```
Maxim-Gadalov: missing argument for formatting (url)

artsiomkorzun: add file url for %s template

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -83,7 +87,7 @@ private static int contentLength(MultiMap header, int defaultValue) {
         return defaultValue;
     }
 
-    public static void collectAttachedFiles(ObjectNode tree, ApiKeyData apiKeyData) {
+    public static void collectAttachedFiles(ObjectNode tree, ApiKeyData apiKeyData, Function<String, Boolean> checkAccessFn) throws Exception {
```
artsiomkorzun, line 90: `Consumer<String>` accessChecker? Do you really need to return boolean?

src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -0,0 +1,39 @@
...
+        String expectedAppDataBucket = BlobStorageUtil.buildAppDataBucket(context);
+        if (expectedAppDataBucket != null && expectedAppDataBucket.equals(decryptedBucket)) {
+            return filePath.startsWith(BlobStorageUtil.APPDATA_PATTERN.formatted(UrlUtil.encodePath(context.getSourceDeployment())));
+        }
+        return false;
+    }
+
+    public boolean hasWriteAccess(String url, ProxyContext context) {
+        if (url == null) {
+            return false;
```
artsiomkorzun, line 31: Is it not possible to use a third-party resource? Since we do not check the prefix schema://host:port

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
astsiapanay: Do you want to throw an exception in the consumer auction directly?

artsiomkorzun: Yes

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -102,7 +106,11 @@ public static void collectAttachedFiles(ObjectNode tree, ApiKeyData apiKeyData)
                 JsonNode attachment = attachments.get(j);
                 JsonNode url = attachment.get("url");
                 if (url != null) {
-                    apiKeyData.getAttachedFiles().add(url.textValue());
+                    if (checkAccessFn.apply(url.textValue())) {
```
Maxim-Gadalov: this would not work, try to insert a valid public link and see how it's work

http: considered as wrong bucket and lead to access - false, which leads to 403 - but should be 200

src/test/java/com/epam/aidial/core/util/ProxyUtilTest.java
```diff
@@ -109,8 +109,93 @@ public void testCollectAttachedFiles_EmbeddingRequest() throws IOException {
...
+                        ]
+                      }
+                    }
+                  ],
+                  "id": "id"
+                }
+                """;
+        ObjectNode tree = (ObjectNode) ProxyUtil.MAPPER.readTree(content.getBytes());
+        ApiKeyData apiKeyData = new ApiKeyData();
+        assertThrows(HttpException.class, () -> ProxyUtil.collectAttachedFiles(tree, apiKeyData, link -> false));
```
Maxim-Gadalov, line 199: can you assert error message?

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
astsiapanay: it's already fixed.

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
astsiapanay: agreed to keep it as is

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
astsiapanay: fixed

src/test/java/com/epam/aidial/core/util/ProxyUtilTest.java
astsiapanay: I think it wouldn't be necessary. Let's skip it.

src/main/java/com/epam/aidial/core/security/AccessService.java
astsiapanay: yes, it's. fixed.

Maxim-Gadalov (approved):

PR #241, opened 2024-02-27
fix: unintentional sharing; clean up invitations and revoke shared access

Maxim-Gadalov:
#221


src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -98,14 +96,10 @@ void start() throws Exception {
...
 
-            if (redis != null) {
-                LockService lockService = new LockService(redis);
-                resourceService = new ResourceService(vertx, redis, storage, lockService, settings("resources"), storage.getPrefix());
-                invitationService = new InvitationService(resourceService, encryptionService, settings("invitations"));
-                shareService = new ShareService(resourceService, invitationService, encryptionService);
-            } else {
-                log.warn("Redis config is not found, some features may be unavailable");
-            }
+            LockService lockService = new LockService(redis);
```
astsiapanay, line 99: Are we going to make Redis mandatory?

src/main/java/com/epam/aidial/core/service/LockService.java
```diff
@@ -42,6 +46,21 @@ public Lock lock(String key) {
         return () -> unlock(id, owner);
     }
 
+    public <T> T underUserLock(Proxy proxy, ProxyContext context, Supplier<T> function) {
+        String userBucketLocation = BlobStorageUtil.buildInitiatorBucket(context);
+        String prefix = proxy.getStorage().getPrefix();
```
astsiapanay: The logic to handle the prefix case could be moved to BlobStorageUtil. Someone could forget add this logic.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -158,9 +164,12 @@ private Future<?> deleteResource(ResourceDescription descriptor) {
         return vertx.executeBlocking(() -> {
                     Set<ResourceLink> resourceLinks = new HashSet<>();
                     resourceLinks.add(new ResourceLink(descriptor.getUrl()));
-                    shareService.revokeSharedAccess(descriptor.getBucketName(), descriptor.getBucketLocation(),
-                            new ResourceLinkCollection(resourceLinks));
-                    return service.deleteResource(descriptor);
+                    return lockService.underUserLock(proxy, context, () -> {
```
astsiapanay: We're going to use multiple level locks:

1. User level lock everything will be locked under /Users/<user>
2. Resource level lock

Could you explain why we need both lock levels?
User level lock is a strong lock which could prevent a concurrent access to different resources of the same user. We need to understand reasons of doing that.

src/main/java/com/epam/aidial/core/service/InvitationService.java
```diff
@@ -125,6 +126,30 @@ public InvitationCollection getMyInvitations(String bucket, String location) {
         return new InvitationCollection(new HashSet<>(invitationMap.getInvitations().values()));
     }
 
+    public void cleanUpResourceLinks(String bucket, String location, Set<ResourceLink> resourcesToCleanUp) {
+        ResourceDescription resource = ResourceDescription.fromDecoded(ResourceType.INVITATION, bucket, location, INVITATION_RESOURCE_FILENAME);
+        resourceService.computeResource(resource, state -> {
+            InvitationsMap invitations = ProxyUtil.convertToObject(state, InvitationsMap.class);
+            if (invitations == null) {
+                invitations = new InvitationsMap(new HashMap<>());
```
astsiapanay: I think we don't need to create an empty map and proceed the request further.
Just make a return here.

src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -130,7 +124,7 @@ void start() throws Exception {
     private RedissonClient openRedis() throws IOException {
         JsonObject conf = settings("redis");
         if (conf.isEmpty()) {
-            return null;
+            throw new IllegalStateException("Redis configuration not found");
```
artsiomkorzun: replace with IllegalArgumentException

src/main/java/com/epam/aidial/core/AiDial.java
Maxim-Gadalov: yes

src/main/java/com/epam/aidial/core/service/LockService.java
```diff
@@ -42,6 +46,21 @@ public Lock lock(String key) {
         return () -> unlock(id, owner);
     }
 
+    public <T> T underUserLock(Proxy proxy, ProxyContext context, Supplier<T> function) {
```
artsiomkorzun: underBucketLock is better name?

src/main/java/com/epam/aidial/core/controller/DeleteFileController.java
```diff
@@ -35,15 +41,14 @@ protected Future<?> handle(ResourceDescription resource) {
...
         Future<Void> result = proxy.getVertx().executeBlocking(() -> {
             try {
-                // clean shared access
-                // TODO remove check when redis become mandatory
-                if (shareService != null) {
-                    Set<ResourceLink> resourceLinks = new HashSet<>();
-                    resourceLinks.add(new ResourceLink(resource.getUrl()));
+                Set<ResourceLink> resourceLinks = new HashSet<>();
+                resourceLinks.add(new ResourceLink(resource.getUrl()));
+                return lockService.underUserLock(proxy, context, () -> {
```
artsiomkorzun: Please take the bucket from the resource not from the context. We will add logic for admins, so admin with users/admin will have permissions to delete users/any resources.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -158,9 +164,12 @@ private Future<?> deleteResource(ResourceDescription descriptor) {
         return vertx.executeBlocking(() -> {
                     Set<ResourceLink> resourceLinks = new HashSet<>();
                     resourceLinks.add(new ResourceLink(descriptor.getUrl()));
```
artsiomkorzun, line 166: use resource bucket not user's.

src/main/java/com/epam/aidial/core/service/LockService.java
```diff
@@ -42,6 +46,21 @@ public Lock lock(String key) {
...
+        String userBucketLocation = BlobStorageUtil.buildInitiatorBucket(context);
+        String prefix = proxy.getStorage().getPrefix();
+        if (prefix != null) {
+            userBucketLocation = prefix + BlobStorageUtil.PATH_SEPARATOR + userBucketLocation;
+        }
+        return underUserLock(userBucketLocation, function);
+    }
+
+    private <T> T underUserLock(String userBucketLocation, Supplier<T> function) {
+        try (var ignored = lock(userBucketLocation)) {
```
artsiomkorzun: underBucketLock is better name?

src/main/java/com/epam/aidial/core/controller/InvitationController.java
```diff
@@ -40,14 +44,20 @@ public Future<?> getInvitations() {
...
                         String bucketLocation = BlobStorageUtil.buildInitiatorBucket(context);
                         String bucket = encryptionService.encrypt(bucketLocation);
-                        shareService.acceptSharedResources(bucket, bucketLocation, invitationId);
-                        return null;
+                        ResourceDescription invitationResource = invitationService.getInvitationResource(invitationId);
+                        if (invitationResource == null) {
+                            throw new ResourceNotFoundException();
+                        }
+                        return lockService.underUserLock(proxy, context, () -> {
+                            shareService.acceptSharedResources(bucket, bucketLocation, invitationId);
```
artsiomkorzun, line 58: It uses two buckets: my (target) and a different user (source). So you need to lock source user bucket to defend against concurrent deletion of resources in source user's invitations.

src/main/java/com/epam/aidial/core/controller/InvitationController.java
```diff
@@ -40,14 +44,20 @@ public Future<?> getInvitations() {
...
                     .executeBlocking(() -> {
                         String bucketLocation = BlobStorageUtil.buildInitiatorBucket(context);
                         String bucket = encryptionService.encrypt(bucketLocation);
-                        shareService.acceptSharedResources(bucket, bucketLocation, invitationId);
-                        return null;
+                        ResourceDescription invitationResource = invitationService.getInvitationResource(invitationId);
+                        if (invitationResource == null) {
+                            throw new ResourceNotFoundException();
+                        }
+                        return lockService.underUserLock(proxy, context, () -> {
```
artsiomkorzun: take lock of invitation owner

artsiomkorzun (approved):

astsiapanay (approved):

Issue #239, opened 2024-02-26
Support available credentials providers for GCP blob storage


astsiapanay:
GKE - https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity

Workload identity federation for GKE

astsiapanay:
JClouds doesn't support workload identity federation for GKE using metadata server to obtain a token for access to the blob store. See https://issues.apache.org/jira/browse/JCLOUDS-1585

There's only one way: use private key and client email. See https://jclouds.apache.org/guides/google/

astsiapanay:
We found a workaround to overcome the issue by supplying access token directly to JClouds and choosing the credential type `bearerTokenCredentials`

PR #245, opened 2024-02-28
feat: Support available credentials providers for GCP blob storage #239

astsiapanay:


src/main/java/com/epam/aidial/core/storage/credential/CredentialProviderFactory.java
```diff
@@ -0,0 +1,17 @@
...
+import com.epam.aidial.core.storage.StorageProvider;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class CredentialProviderFactory {
+    public static CredentialProvider create(String providerName, String identity, String credential) {
+        StorageProvider provider = StorageProvider.from(providerName);
+        return switch (provider) {
+            case S3, AZURE_BLOB -> new DefaultCredentialProvider(identity, credential);
+            case GOOGLE_CLOUD_STORAGE -> new GcpCredentialProvider(credential);
```
Maxim-Gadalov: why you remove the "old" one with static credentials for gcp?

astsiapanay: It's not removed but replaced with new one.

Maxim-Gadalov (approved):

PR #260, opened 2024-03-05
feat: publication part 1

artsiomkorzun:


src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -69,6 +73,12 @@ public class AiDial {
 
     private BlobStorage storage;
     private ResourceService resourceService;
+    private InvitationService invitationService;
+    private ShareService shareService;
+    private PublicationService publicationService;
```
Maxim-Gadalov: I see no usage for this services

src/test/java/com/epam/aidial/core/ResourceBaseTest.java
```diff
@@ -226,6 +236,13 @@ Response send(HttpMethod method, String path, String queryParams, String body, S
         }
     }
 
+    @SneakyThrows
+    private static String pretty(String json) {
```
Maxim-Gadalov: no usages

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -19,8 +19,10 @@
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class ResourceDescription {
 
-    private static final Set<Character> INVALID_FILE_NAME_CHARS = Set.of('/', '{', '}');
+    public static final String PUBLIC_BUCKET = "public";
+    public static final String PUBLIC_LOCATION = PUBLIC_BUCKET + BlobStorageUtil.PATH_SEPARATOR;
```
Maxim-Gadalov: I would keep public location along with user/key location patterns in BlobStorageUtil

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -165,6 +167,30 @@ public static ResourceDescription fromLink(String link, EncryptionService encryp
...
+
+        String relativePath = link.substring(parts[0].length() + parts[1].length() + 2);
+        return fromEncoded(resourceType, bucketEncoded, bucketDecoded, relativePath);
+    }
+
+    public static ResourceDescription fromBucketLink(String link, ResourceDescription bucket) {
+        return fromLink(link, bucket.getBucketName(), bucket.getBucketLocation());
+    }
+
+    public static ResourceDescription fromPublicLink(String link) {
```
Maxim-Gadalov, line 196: Can you add few tests for this methods? see ResourceDescriptionTest

src/main/java/com/epam/aidial/core/data/Publication.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+import java.util.List;
+
+@Data
+@Accessors(chain = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Publication {
```
Maxim-Gadalov, line 12: Can you add javadocs to the fields? This class represent publication_create body. What if my request contain createdAt or status

src/main/java/com/epam/aidial/core/data/ResourceUrl.java
```diff
@@ -0,0 +1,67 @@
+package com.epam.aidial.core.data;
+
+import com.epam.aidial.core.storage.BlobStorageUtil;
+import com.epam.aidial.core.util.UrlUtil;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class ResourceUrl {
```
Maxim-Gadalov, line 9: what's the value of this class? Can you use ResourceDescription instead?

artsiomkorzun: I need a more generic url without enforcing resource type and bucket. Because I do not have them in some places.

src/main/java/com/epam/aidial/core/data/Publication.java
artsiomkorzun: It is used as publication request, that's true. Nothing will happen because, they are reset. A bit tricky, but otherwise we need to create lots of dto classes for each type of request.

artsiomkorzun: What is not clear? I mean I can add java doc like url - is publication url, but does it really help?

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: added

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: moved

src/test/java/com/epam/aidial/core/ResourceBaseTest.java
artsiomkorzun: a victim of merge

src/main/java/com/epam/aidial/core/AiDial.java
artsiomkorzun: a victim of merge

src/main/java/com/epam/aidial/core/data/ResourceUrl.java
Maxim-Gadalov: We need to think how to consolidate those classes, right now you validate only path separator in segment name but we added more chars recently, and same things may happen again and again.

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -0,0 +1,359 @@
...
+        if (publication.getSourceUrl() == null) {
+            throw new IllegalArgumentException("Publication \"sourceUrl\" is missing");
+        }
+
+        if (publication.getTargetUrl() == null) {
+            throw new IllegalArgumentException("Publication \"targetUrl\" is missing");
+        }
+
+        if (publication.getResources() == null) {
+            publication.setResources(List.of());
```
Maxim-Gadalov: probably if resources are empty we need to throw an exception

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -0,0 +1,359 @@
...
+
+        for (Publication.Resource resource : publication.getResources()) {
+            ResourceDescription source = ResourceDescription.fromBucketLink(resource.getSourceUrl(), bucket);
+            String sourceUrl = source.getUrl();
+
+            if (source.isFolder()) {
+                throw new IllegalArgumentException("Resource folder is not allowed: " + sourceUrl);
+            }
+
+            if (source.getType() != ResourceType.FILE && source.getType() != ResourceType.PROMPT && source.getType() != ResourceType.CONVERSATION) {
```
Maxim-Gadalov: make allowed resource set

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -0,0 +1,359 @@
...
+                throw new IllegalArgumentException("Resource type is not supported: " + sourceUrl);
+            }
+
+            String sourceSuffix = sourceUrl.substring(source.getType().getGroup().length() + 1);
+
+            if (!sourceSuffix.startsWith(publication.getSourceUrl())) {
```
Maxim-Gadalov: Are you sure about this condition? all auto-generated attachments will not pass this check, it's require move/copy files to the source folder first

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -0,0 +1,359 @@
...
+    private String encodeReviewBucket(ResourceDescription bucket, String id) {
+        String path = bucket.getBucketLocation()
+                + PUBLICATIONS_NAME + PATH_SEPARATOR
+                + id + PATH_SEPARATOR;
+
+        return encryption.encrypt(path);
+    }
+
+    private static void leaveDescription(Publication publication) {
+        publication.setResources(null).setResources(null);
```
Maxim-Gadalov: redundant call, could you leave a comment? I don't understand this method. Why you clean resources on create/list operation?

src/main/java/com/epam/aidial/core/controller/PublicationController.java
```diff
@@ -0,0 +1,142 @@
...
+            status = e.getStatus();
+            body = e.getMessage();
+        } else if (error instanceof IllegalArgumentException e) {
+            status = HttpStatus.BAD_REQUEST;
+            body = e.getMessage();
+        } else {
+            log.warn(message, error);
+        }
+
+        context.respond(status, body == null ? "" : body);
```
Maxim-Gadalov: redundant null check, see `content.respond`

src/main/java/com/epam/aidial/core/controller/PublicationController.java
```diff
@@ -0,0 +1,142 @@
...
+                    checkAccess(resource);
+                    return vertx.executeBlocking(() -> publicationService.deletePublication(resource));
+                })
+                .onSuccess(deleted -> context.respond(deleted ? HttpStatus.OK : HttpStatus.NOT_FOUND))
+                .onFailure(error -> respond("Can't delete publication", error));
+
+        return Future.succeededFuture();
+    }
+
+    private void respond(String message, Throwable error) {
```
Maxim-Gadalov: rename to respondError/respondeWithError/handleError

src/main/java/com/epam/aidial/core/data/Publication.java
artsiomkorzun: Added

src/main/java/com/epam/aidial/core/service/PublicationService.java
artsiomkorzun: There is a case when you want to change just rules.

Maxim-Gadalov: create operation responsible for update as well?

src/main/java/com/epam/aidial/core/service/PublicationService.java
artsiomkorzun: only header is stored for admin.

src/main/java/com/epam/aidial/core/controller/PublicationController.java
artsiomkorzun: removed

src/main/java/com/epam/aidial/core/controller/PublicationController.java
artsiomkorzun: renamed

Maxim-Gadalov (dismissed):

Maxim-Gadalov (approved):

src/main/java/com/epam/aidial/core/controller/AccessControlBaseController.java
```diff
@@ -54,12 +54,16 @@ public Future<?> handle(String resourceType, String bucket, String path) {
                             return true;
                         }
 
+                        if (proxy.getAccessService().isReviewResource(resource, context)) {
```
astsiapanay, line 57: why can we not replace 
 ```
"reviewUrl" : "conversations/2CZ9i2bcBACFts8JbBu3MdTHfU5imDZBmDVomBuDCkbhEstv1KXNzCiw693js8BLmo/conversation"
 ```
to
 ```
"reviewUrl" : "conversations/3CcedGxCx23EwiVbVmscVktScRyf46KypuBQ65miviST/publications/id/conversation"
 ```
and remove 

``` 
if (accessService.hasReviewAccess(resource, context)) {
return true;
}

```
?

PR #269, opened 2024-03-11
feat: publications rules

artsiomkorzun:


src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
```diff
@@ -328,4 +310,18 @@ private Matcher match(Pattern pattern, String path) {
         Matcher matcher = pattern.matcher(path);
         return matcher.find() ? matcher : null;
     }
+
+    private String resourcePath(String url) {
+        String prefix = "/v1/";
+
+        if (!url.startsWith(prefix)) {
```
astsiapanay, line 325: Nit. It's unnecessary to check the condition. Regexp matcher doesn't pass a wrong url.

astsiapanay (dismissed):

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
artsiomkorzun: Yes, but it is more safe approach of doing so, plus it is more obvious for the reader.

src/main/resources/aidial.settings.json
```diff
@@ -56,5 +56,10 @@
     "syncBatch": 4096,
     "cacheExpiration": 300000,
     "compressionMinSize": 256
+  },
+  "access": {
```
Maxim-Gadalov, line 60: describe settings in readme

src/main/java/com/epam/aidial/core/service/LockService.java
```diff
@@ -20,13 +20,15 @@
...
-    private static final long PERIOD = TimeUnit.SECONDS.toMicros(60);
+    private static final long PERIOD = TimeUnit.SECONDS.toMicros(300);
     private static final long WAIT_MIN = TimeUnit.MILLISECONDS.toNanos(1);
     private static final long WAIT_MAX = TimeUnit.MILLISECONDS.toNanos(128);
 
+    private final String prefix;
     private final RScript script;
 
-    public LockService(RedissonClient redis) {
+    public LockService(RedissonClient redis, @Nullable BlobStorage storage) {
```
Maxim-Gadalov: take prefix from storage config, lock service should not require storage

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
Maxim-Gadalov: it's hard to maintain in future, when you have v2, v3 or so. Obviously I don't see a real value of checking this

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -146,22 +154,68 @@ public static ResourceDescription fromDecoded(ResourceDescription description, S
         return fromDecoded(description.getType(), description.getBucketName(), description.getBucketLocation(), relativePath);
     }
 
-    public static ResourceDescription fromLink(String link, EncryptionService encryptionService) {
-        String[] parts = link.split(BlobStorageUtil.PATH_SEPARATOR);
+    public static ResourceDescription fromPublicUrl(String url) {
+        return fromUrl(url, BlobStorageUtil.PUBLIC_BUCKET, BlobStorageUtil.PUBLIC_LOCATION, null);
+    }
+
+    public static ResourceDescription fromPrivateUrl(String url, ResourceDescription bucket) {
```
Maxim-Gadalov: resource, not bucket

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -277,7 +277,7 @@ private ResourceDescription getResourceDescription(String url) {
             // skip public resource
             return null;
         }
-        return ResourceDescription.fromLink(url, proxy.getEncryptionService());
+        return ResourceDescription.fromPrivateUrl(url, proxy.getEncryptionService());
```
Maxim-Gadalov: what if conversation contains public links - whole chat completion will fail. Please write a test for this case

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
```diff
@@ -23,6 +23,7 @@ protected Future<?> handle(ResourceDescription resource) {
             try {
                 MetadataBase metadata = storage.listMetadata(resource);
                 if (metadata != null) {
+                    proxy.getAccessService().filterForbidden(context, resource, metadata);
```
Maxim-Gadalov, line 26: filter metadata if provided resource is public

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
artsiomkorzun: The future is not yet to come. v11 won't work as well.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -85,6 +85,7 @@ private Future<?> getMetadata(ResourceDescription descriptor) {
                     if (result == null) {
                         context.respond(HttpStatus.NOT_FOUND, "Not found: " + descriptor.getUrl());
                     } else {
+                        proxy.getAccessService().filterForbidden(context, descriptor, result);
```
Maxim-Gadalov, line 88: same

src/main/resources/aidial.settings.json
artsiomkorzun: added

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: We take only bucket information from this resource. So name "bucket" highlights it.

src/main/java/com/epam/aidial/core/service/LockService.java
artsiomkorzun: Now it takes String prefix.

src/main/java/com/epam/aidial/core/controller/ControllerSelector.java
artsiomkorzun: At least it will fail, when you will add a new version. But not silently fail.

src/main/java/com/epam/aidial/core/controller/PublicationController.java
```diff
@@ -132,10 +168,19 @@ private ResourceDescription decodePublication(String path) {
         return resource;
     }
 
-    private void checkAccess(ResourceDescription resource) {
-        String bucket = BlobStorageUtil.buildInitiatorBucket(context);
+    private void checkAccess(ResourceDescription resource, boolean onlyAdmin, boolean onlyPrivate) {
```
Maxim-Gadalov: This method require java doc or proper naming.
This 2 flags are controversial. I expect either admin access required or user access. In your case you have both true.

`
checkAccess(ResourceDescription resource, boolean requireAdminAccess);

// if user has admin access - immediately return 
if (accessService.hasAdminAccess(context)) {
return;
}

// if private resource - check bucket
if (!requireAdminAccess && resource.getBucketLocation().equals(bucket)) {
 return;
}

throw new FobiddenException();
`

src/main/java/com/epam/aidial/core/security/RuleMatcher.java
```diff
@@ -0,0 +1,81 @@
...
+import com.epam.aidial.core.ProxyContext;
+import com.epam.aidial.core.data.Rule;
+import lombok.experimental.UtilityClass;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.regex.Pattern;
+
+@UtilityClass
+public class RuleMatcher {
```
Maxim-Gadalov, line 12: can you write some unit tests to cover this class?

src/main/java/com/epam/aidial/core/controller/PublicationController.java
```diff
@@ -91,22 +92,57 @@ public Future<?> deletePublication() {
...
+    }
+
+    public Future<?> rejectPublication() {
+        context.getRequest()
+                .body()
+                .compose(body -> {
+                    String url = ProxyUtil.convertToObject(body, ResourceLink.class).url();
+                    ResourceDescription resource = decodePublication(url);
+                    checkAccess(resource, true, true);
+                    return vertx.executeBlocking(() -> publicationService.rejectPublication(resource));
```
Maxim-Gadalov, line 128: I think reject also require publication lock

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
Maxim-Gadalov: there is no bucket of ResourceDescription - bucket is a part of each resource.This method has only 1 usage, maybe we should take bucket and location explicitly.

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -146,22 +154,68 @@ public static ResourceDescription fromDecoded(ResourceDescription description, S
...
+        }
+
+        return description;
+    }
+
+    public static ResourceDescription fromAnyUrl(String url, EncryptionService encryption) {
+        return fromUrl(url, null, null, encryption);
+    }
+
+    private static ResourceDescription fromUrl(String url,
```
Maxim-Gadalov, line 190: can you add Nullable annotations for the arguments? It seems this method become a key point for the future usage (extension).
As far as I understand we should provide either expected bucket and location or encryptionService - not all at the same time

src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -1,52 +1,117 @@
...
-        String expectedAppDataBucket = BlobStorageUtil.buildAppDataBucket(context);
-        if (expectedAppDataBucket != null && expectedAppDataBucket.equals(decryptedBucket)) {
-            return filePath.startsWith(BlobStorageUtil.APPDATA_PATTERN.formatted(UrlUtil.encodePath(context.getSourceDeployment())));
+
+        if (resource.isPublic()) {
+            return hasPublicAccess(resource, context);
         }
-        return false;
+
+        return isMyResource(resource, context) || isAppResource(resource, context)
```
astsiapanay, line 43: please merge if statements into a single one
```
if (resource.isPublic()) {
             return hasPublicAccess(resource, context);
         }
return hasAdminAccess(context) || isAutoShared(resource, context) ||  isMyResource(resource, context) || isAppResource(resource, context)
                 || hasReviewAccess(resource, context) || isSharedResource(resource, context);
```

src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -1,52 +1,117 @@
...
+                         PublicationService publicationService,
+                         JsonObject settings) {
+        this.encryptionService = encryptionService;
+        this.shareService = shareService;
+        this.publicationService = publicationService;
+        this.adminRules = adminRules(settings);
+    }
+
+    public boolean hasReadAccess(ResourceDescription resource, ProxyContext context) {
+        if (hasAdminAccess(context)) {
```
astsiapanay: I think it may break user data privacy. it should be
```
if (resource.isPublic() && hasAdminAccess(context))
```

artsiomkorzun: Reworked Admin has write access to public/review resources

src/main/java/com/epam/aidial/core/security/AccessService.java
artsiomkorzun: Why?

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: Added java doc.

src/main/java/com/epam/aidial/core/controller/PublicationController.java
artsiomkorzun: Why?

src/main/java/com/epam/aidial/core/controller/PublicationController.java
artsiomkorzun: refactored

src/main/java/com/epam/aidial/core/controller/FileMetadataController.java
artsiomkorzun: check the method

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: same

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
artsiomkorzun: fixed

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: renamed, added javadoc

src/main/java/com/epam/aidial/core/security/RuleMatcher.java
artsiomkorzun: added

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
Maxim-Gadalov: should we add a check when all 3 are null?

src/main/java/com/epam/aidial/core/controller/PublicationController.java
Maxim-Gadalov: I see, you have a double check

src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
artsiomkorzun: It will handle public resources if all of them are null

src/main/java/com/epam/aidial/core/security/AccessService.java
artsiomkorzun: Changed to have admin access only to public + review resources.

Maxim-Gadalov (approved):

src/main/java/com/epam/aidial/core/security/AccessService.java
astsiapanay: It's a developer preference. I think the statement 
```
if (isAutoShared(resource, context)) {
             return true;
         }
```
could be merge in to the last one in the method body.

astsiapanay (approved):

Issue #265, opened 2024-03-06
Support auto-sharing of folders


adubovik:
Required for RAG-like applications:
1. User attaches a whole folder from one's userspace
2. Calls a RAG application with this attachment
3. RAG application must have a access to every single file in the given directory to be able to index them

PR #272, opened 2024-03-11
feat: Support auto-sharing of folders #265

astsiapanay:


src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -49,4 +51,19 @@ public boolean isReviewResource(ResourceDescription resource, ProxyContext conte
         String actualUserBucket = encryptionService.encrypt(actualUserLocation);
         return publicationService != null && publicationService.hasReviewAccess(resource, actualUserBucket, actualUserLocation);
     }
+
+    public boolean isAutoSharedResource(ResourceDescription resource, ProxyContext context) {
```
Maxim-Gadalov, line 55: can you write a test workflow for auto-shared folder?

astsiapanay: yes. added

Maxim-Gadalov (approved):

PR #274, opened 2024-03-12
fix: run processing request body async in order to prevent event thread blocking

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -250,6 +253,11 @@ void handleRequestBody(Buffer requestBody) {
     }
 
     private void processAttachedFile(String url) {
+        try {
+            Thread.sleep(5000);
```
artsiomkorzun: Leftover after testing

astsiapanay: my bad. fixed.

artsiomkorzun (approved):

PR #281, opened 2024-03-14
feat: cache public rules

artsiomkorzun:


src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -53,6 +56,7 @@ public class PublicationService {
 
     private static final Set<ResourceType> ALLOWED_RESOURCES = Set.of(ResourceType.FILE, ResourceType.CONVERSATION, ResourceType.PROMPT);
 
+    private final AtomicReference<Pair<Long, Map<String, List<Rule>>>> cachedRules = new AtomicReference<>();
```
astsiapanay, line 63: Please add javadoc, what the key and the value are.

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -473,6 +477,24 @@ private String encodeReviewBucket(ResourceDescription bucket, String id) {
...
+        long key = (meta == null) ? Long.MIN_VALUE : meta.getUpdatedAt();
+        Pair<Long, Map<String, List<Rule>>> current = cachedRules.get();
+
+        if (current == null || current.getKey() != key) {
+            Pair<ResourceItemMetadata, String> resource = resources.getResourceWithMetadata(PUBLIC_RULES);
+            Pair<Long, Map<String, List<Rule>>> next = (resource == null)
+                    ? Pair.of(Long.MIN_VALUE, decodeRules(null))
+                    : Pair.of(resource.getKey().getUpdatedAt(), decodeRules(resource.getValue()));
+
+            cachedRules.compareAndSet(current, next);
```
astsiapanay, line 516: you should evaluate a result of the CAS operation to make sure it's successful

artsiomkorzun: Not needed.

astsiapanay (approved):

Issue #277, opened 2024-03-13
Implement rate limits for requests


astsiapanay:
Support request limits in DIAL config:
```
"roles": {
        "default": {
            "limits": {
                "chat-gpt-35-turbo": {
                    "minute": "100000",
                    "day": "10000000",
                    "requestHour": "5",
                    "requestDay": "20"
                }
```

PR #282, opened 2024-03-14
feat: Implement rate limits for requests #277

astsiapanay:


src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -96,71 +94,125 @@ public Future<LimitStats> getLimitStats(String deploymentName, ProxyContext cont
...
-        long timestamp = System.currentTimeMillis();
         return rateLimit.update(timestamp, limit);
     }
 
-    private Void updateLimit(ResourceDescription resourceDescription, long totalUsedTokens) {
-        resourceService.computeResource(resourceDescription, json -> updateLimit(json, totalUsedTokens));
+    private RateLimitResult checkRequestLimit(ProxyContext context, Limit limit, long timestamp) {
+        String tokensPath = getPathToRequests(context.getDeployment().getName());
+        ResourceDescription resourceDescription = getResourceDescription(context, tokensPath);
+        RateLimitResult[] result = new RateLimitResult[1];
```
Maxim-Gadalov, line 194: you can make this class mutable to avoid passing array in lambda

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
```diff
@@ -96,71 +94,125 @@ public Future<LimitStats> getLimitStats(String deploymentName, ProxyContext cont
...
+        RateLimitResult tokenResult = checkTokenLimit(context, limit, timestamp);
+        if (tokenResult.status() != HttpStatus.OK) {
+            return tokenResult;
+        }
+        return checkRequestLimit(context, limit, timestamp);
+    }
+
+    private RateLimitResult checkTokenLimit(ProxyContext context, Limit limit, long timestamp) {
+        String tokensPath = getPathToTokens(context.getDeployment().getName());
+        ResourceDescription resourceDescription = getResourceDescription(context, tokensPath);
```
Maxim-Gadalov, line 181: you update this state later but never save it back to the storage or I missing something?

astsiapanay: The state will be updated later by calling the method increase.

src/main/java/com/epam/aidial/core/limiter/RateLimiter.java
astsiapanay: added a comment to explain the reason for doing so

Maxim-Gadalov (approved):

Issue #271, opened 2024-03-11
Support ability to share file in already shared conversation


sdryapko:
Let's add API that allows owner to copy access list from one resource to another. For instance from chat to newly added file.

PR #283, opened 2024-03-14
feat: implement copy shared access operation

Maxim-Gadalov:
#271

### DIFF
--- a/src/main/java/com/epam/aidial/core/controller/ShareController.java
+++ b/src/main/java/com/epam/aidial/core/controller/ShareController.java
@@ -2,17 +2,22 @@ package com.epam.aidial.core.controller;
 
 import com.epam.aidial.core.Proxy;
 import com.epam.aidial.core.ProxyContext;
+import com.epam.aidial.core.data.CopySharedAccessRequest;
 import com.epam.aidial.core.data.ListSharedResourcesRequest;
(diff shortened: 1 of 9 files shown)


src/main/java/com/epam/aidial/core/controller/ShareController.java
```diff
@@ -137,6 +147,49 @@ public Future<?> revokeSharedResources() {
...
+                .compose(buffer -> {
+                    CopySharedAccessRequest request;
+                    try {
+                        request = ProxyUtil.convertToObject(buffer, CopySharedAccessRequest.class);
+                    } catch (Exception e) {
+                        log.error("Invalid request body provided", e);
+                        throw new IllegalArgumentException("Can't initiate copy shared access request. Incorrect body provided");
+                    }
+
+                    String sourceUrl = request.sourceUrl();
```
astsiapanay, line 162: It could be easier to set constraint on JSON fields instead of writing checks in the controller

Maxim-Gadalov: I prefer a user friendly error messages in this case (only 2 checks compare to conversation validation)

astsiapanay: ok. let's keep as is

astsiapanay (approved):

Issue #240, opened 2024-02-27
Broadcast DIAL version through API


nepalevov:
We need a text field in the core configuration, where the DIAL version (as a software system) can be placed.

This version text field must be available at some endpoint, so ai-dial-core clients can get it and e.g. show it to the end user (ai-dial-chat).

P.S. I don't have any proposal for designing this system metadata configuration block, but in general, it may be later extended with some other metadata fields, e.g. global announcements.

PR #290, opened 2024-03-25
feat: implement version endpoint

Maxim-Gadalov:
#240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,7 @@ build/
 bin/
 /data/
 *.log
+# gradle distribution task will write a version from build.gradle to the version file.
+# since CI bumps version only in build.gradle and we are copying this value on fly we don't want to track version file changes
+# even if developer/user create own version file - distribution task overwrite it
 src/main/resources/version
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/
 bin/
 /data/
 *.log
+src/main/resources/version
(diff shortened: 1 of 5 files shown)


Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-290.nightly-test.deltixhub.io
>E2E tests status: success

.gitignore
```diff
@@ -4,4 +4,5 @@
 build/
 bin/
 /data/
-*.log
\ No newline at end of file
+*.log
+src/main/resources/version
```
astsiapanay, line 11: Please add a comment why we don't check-in the resource

build.gradle
```diff
@@ -74,6 +74,16 @@ application {
     mainClass = 'com.epam.aidial.core.AiDial'
 }
 
+distributions {
+    // write current version to the version file
+    beforeEvaluate {
+        def versionFile = file("$projectDir/src/main/resources/version")
+        versionFile.getParentFile().mkdirs()
```
astsiapanay, line 81: what happens if resources or file already exists?

.gitignore
Maxim-Gadalov: added

build.gradle
Maxim-Gadalov: according to the documentation, file and parent directory(s) will be created if not exists

astsiapanay (approved):

PR #302, opened 2024-04-09
feat: implement publication deletion workflow

Maxim-Gadalov:
#295

### DIFF
--- a/src/main/java/com/epam/aidial/core/service/PublicationService.java
+++ b/src/main/java/com/epam/aidial/core/service/PublicationService.java
@@ -231,38 +231,37 @@ public class PublicationService {
         return publication;
     }
 
-    public Publication deletePublication(ResourceDescription resource) {
+    public Publication deletePublication(ResourceDescription resource, boolean isAdmin) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -322,12 +321,13 @@ public Publication rejectPublication(ResourceDescription resource) {
...
 
-            if (publication.getStatus() != Publication.Status.PENDING) {
+            boolean isPending = publication.getStatus() == Publication.Status.PENDING;
+            if (!isPending && publication.getStatus() != Publication.Status.REQUESTED_FOR_DELETION) {
                 throw new ResourceNotFoundException("Publication is already finalized: " + resource.getUrl());
             }
 
             reference.setValue(publication);
-            publication.setStatus(Publication.Status.REJECTED);
+            publication.setStatus(isPending ? Publication.Status.REJECTED : Publication.Status.APPROVED);
```
astsiapanay, line 332: Please add a comment why we set the status to approved. It's not clear.

astsiapanay (dismissed):

astsiapanay (approved):

PR #303, opened 2024-04-10
fix: improve resource name validation

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/storage/ResourceDescription.java
```diff
@@ -11,15 +11,15 @@
...
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 @Data
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class ResourceDescription {
 
-    private static final Set<Character> INVALID_FILE_NAME_CHARS = Set.of('/', '{', '}');
+    private static final Pattern INVALID_FILE_NAME_CHARS_PATTERN = Pattern.compile(".*[\\x00-\\x1f/}{].*");
```
astsiapanay: it's easy to use range instead of writing regexp
```
private static boolean isValidFilename(String value) {
        for (int i = 0; i < value.length(); i++) {
            char c = value.charAt(i);
            if (c <= 0x1F || INVALID_FILE_NAME_CHARS.contains(c)) {
                return false;
            }
        }
        return true;
    }
```

Maxim-Gadalov: changed

astsiapanay (approved):

Issue #306, opened 2024-04-12
Add custom claim check support for publications


sdryapko:

Maxim-Gadalov:
For the first iteration, we agreed to support string and array of strings.

PR #309, opened 2024-04-17
feat: support custom claims check for publications

Maxim-Gadalov:
#306

### DIFF
--- a/src/main/java/com/epam/aidial/core/security/ExtractedClaims.java
+++ b/src/main/java/com/epam/aidial/core/security/ExtractedClaims.java
@@ -1,6 +1,7 @@
 package com.epam.aidial.core.security;
 
 import java.util.List;
+import java.util.Map;
 
-public record ExtractedClaims(String sub, List<String> userRoles, String userHash) {
+public record ExtractedClaims(String sub, List<String> userRoles, String userHash, Map<String, List<String>> userClaims) {
 }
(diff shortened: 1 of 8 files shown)


Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-309.nightly-test.deltixhub.io
>E2E tests status: success

src/main/java/com/epam/aidial/core/security/ExtractedClaims.java
```diff
@@ -1,6 +1,7 @@
 package com.epam.aidial.core.security;
 
 import java.util.List;
+import java.util.Map;
 
-public record ExtractedClaims(String sub, List<String> userRoles, String userHash) {
+public record ExtractedClaims(String sub, List<String> userRoles, String userHash, Map<String, List<String>> userClaims) {
```
astsiapanay, line 6: Perhaps we can put user claims  reference and later a caller may extract some claims if needed.
What do you think?

astsiapanay: Moreover I would move the logic of some claims extraction to the class ExtractedClaims.
and leave just userHash since it's computed.

astsiapanay: as agreed. let's postpone the suggested implementation

astsiapanay (approved):

Issue #183, opened 2024-01-29
Implement per request API keys through stateful layer


astsiapanay:
Store/verify per request API keys of app/user into Redis

PR #312, opened 2024-04-19
feat: Implement per request API keys through stateful layer #183

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -619,8 +625,14 @@ private Future<Void> respond(HttpStatus status, Object result) {
...
-        if (context.getProxyApiKeyData() != null) {
-            proxy.getApiKeyStore().invalidateApiKey(context.getProxyApiKeyData());
+        ApiKeyData proxyApiKeyData = context.getProxyApiKeyData();
+        if (proxyApiKeyData != null) {
+            proxy.getApiKeyStore().invalidateApiKey(proxyApiKeyData)
+                    .onSuccess(invalidated -> {
+                        if (!invalidated) {
+                            log.warn("Per request is not removed: {}", proxyApiKeyData.getPerRequestKey());
+                        }
+                    }).onFailure(error -> log.error("error occurred on invalidating per-request key: {}", error));
```
Maxim-Gadalov: onFailure(error -> log.error("error occurred on invalidating per-request key: {}", error.getMessage()));

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -125,19 +125,22 @@ private void handleRateLimitSuccess(String deploymentId) {
...
-
-        proxy.getTokenStatsTracker().startSpan(context);
-
-        context.getRequest().body()
-                .onSuccess(body -> proxy.getVertx().executeBlocking(() -> {
-                    handleRequestBody(body);
-                    return null;
-                }))
-                .onFailure(this::handleRequestBodyError);
+        proxy.getVertx().executeBlocking(() -> {
```
Maxim-Gadalov: maybe it's better to rewrite below code to highlight 2 steps explicitly
```proxy.getVertx().executeBlocking(() -> {
                    ApiKeyData proxyApiKeyData = new ApiKeyData();
                    context.setProxyApiKeyData(proxyApiKeyData);
                    ApiKeyData.initFromContext(proxyApiKeyData, context);
                    proxy.getApiKeyStore().assignApiKey(proxyApiKeyData);
                    proxy.getTokenStatsTracker().startSpan(context);
                    return null;
                })
                .onFailure(this::handleError)
                .map(mapper -> context.getRequest().body()
                        .onSuccess(body -> proxy.getVertx().executeBlocking(() -> {
                            handleRequestBody(body);
                            return null;
                        }))
                        .onFailure(this::handleRequestBodyError));
```

src/main/java/com/epam/aidial/core/security/ApiKeyStore.java
```diff
@@ -2,57 +2,117 @@
...
+            return Future.succeededFuture(apiKeyData);
+        }
+        ResourceDescription resource = toResource(key);
+        return vertx.executeBlocking(() -> ProxyUtil.convertToObject(resourceService.getResource(resource), ApiKeyData.class));
     }
 
-    public synchronized void invalidateApiKey(ApiKeyData apiKeyData) {
-        if (apiKeyData.getPerRequestKey() != null) {
-            keys.remove(apiKeyData.getPerRequestKey());
+    public Future<Boolean> invalidateApiKey(ApiKeyData apiKeyData) {
```
Maxim-Gadalov: does invalidation require bucket lock as well?

src/main/java/com/epam/aidial/core/security/ApiKeyStore.java
```diff
@@ -2,57 +2,117 @@
...
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
+import javax.annotation.concurrent.GuardedBy;
 
 import static com.epam.aidial.core.security.ApiKeyGenerator.generateKey;
+import static com.epam.aidial.core.storage.BlobStorageUtil.PATH_SEPARATOR;
 
 @Slf4j
+@AllArgsConstructor
```
Maxim-Gadalov, line 29: Can you add javadocs for this class or method with some hints how to use it?
For example ApiKeyData stored locally in `keys` and related per-request-keys stored in redis only

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: I think printing just an error message might be useless as it doesn’t reveal a root cause.

src/main/java/com/epam/aidial/core/security/ApiKeyStore.java
astsiapanay: I think it doesn't unless you give an example where that might be required.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: I might be wrong. But this code leads to the same error might be processed twice:
.onFailure(this::handleError) and onFailure(this::handleRequestBodyError)).
Doesn't it?
I will take some time the code can be improved.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: your error message expect 1 argument, but 0 provided

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: map function only called if future succeeded

src/main/java/com/epam/aidial/core/security/ApiKeyStore.java
Maxim-Gadalov: You are right, my initial intention was to block access to this key during finalization. but now I realized that it can't be solved by just adding bucket lock for invalidation, because getApiKey does not hold any lock. let's leave it as is

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: you're right the function map is called when the future is succeeded but the callback onFailure is called when an error happens in the chain.
For example,
`assign api key is failed -> onFailure is called -> map is not called so no new feature is created -> onFailure is called again`

src/main/java/com/epam/aidial/core/security/ApiKeyStore.java
astsiapanay: added

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-312.nightly-test.deltixhub.io
>E2E tests status: failed

Maxim-Gadalov (approved):

PR #333, opened 2024-05-08
fix: publication deletion as request; change empty rules behavior 

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -206,67 +206,78 @@ public Publication createPublication(ResourceDescription bucket, Publication pub
         checkSourceResources(publication);
         checkTargetResources(publication);
 
-        copySourceToReviewResources(publication);
+        // check target location access
+        boolean hasAccess = hasPublicAccess(context, ResourceDescription.fromPublicUrl(publication.getResources().get(0).getTargetUrl()));
```
artsiomkorzun: Check all sub folders?

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -524,6 +532,26 @@ private void checkTargetResources(Publication publication) {
...
+            if (!checkResource(descriptor)) {
+                throw new IllegalArgumentException("Target resource does not exist: " + descriptor.getUrl());
+            }
+        }
+    }
+
+    private void checkTargetResourceAccess(ProxyContext context, Publication publication) {
+        for (Publication.Resource resource : publication.getResources()) {
+            String url = resource.getTargetUrl();
+            if (!hasPublicAccess(context, ResourceDescription.fromPublicUrl(url))) {
```
artsiomkorzun: Please optimize it to check all resources at once.

src/main/java/com/epam/aidial/core/service/PublicationService.java
```diff
@@ -505,6 +510,9 @@ private void checkSourceResources(Publication publication) {
     private void checkReviewResources(Publication publication) {
         for (Publication.Resource resource : publication.getResources()) {
             String url = resource.getReviewUrl();
+            if (url == null) {
```
artsiomkorzun: You do not need to check it. It is always not null for Publication.PENDING.

src/main/java/com/epam/aidial/core/service/PublicationService.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/service/PublicationService.java
Maxim-Gadalov: added

src/main/java/com/epam/aidial/core/service/PublicationService.java
Maxim-Gadalov: removed

artsiomkorzun (approved):

PR #343, opened 2024-05-16
fix: refactor publication API according to the approved design

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/controller/PublicationController.java
```diff
@@ -93,10 +94,10 @@ public Future<?> deletePublication() {
         context.getRequest()
                 .body()
                 .compose(body -> {
-                    Publication publication = ProxyUtil.convertToObject(body, Publication.class);
-                    ResourceDescription resource = decodePublication(publication.getUrl(), false);
+                    String url = ProxyUtil.convertToObject(body, ResourceLink.class).url();
```
astsiapanay, line 97: Is permission check not required any longer?

src/main/java/com/epam/aidial/core/data/Publication.java
```diff
@@ -19,19 +19,20 @@ public class Publication {
     /**
      * Target directory url without resource prefix to publish to: public/ or public/folder/.
      */
-    String targetUrl;
+    String targetFolder;
```
astsiapanay, line 22: I may confuse because a folder is just a singe name. However the description states that's an URL

src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -52,6 +53,28 @@ public boolean hasWriteAccess(ResourceDescription resource, ProxyContext context
         return isMyResource(resource, context) || isAppResource(resource, context) || hasReviewAccess(resource, context, false);
     }
 
+    public boolean hasPublicAccess(List<ResourceDescription> resources, ProxyContext context) {
+        boolean isAllPublic = resources.stream().allMatch(ResourceDescription::isPublic);
+        if (!isAllPublic) {
+            throw new IllegalArgumentException("Provided resources must be public");
```
astsiapanay: Why not to return just `false`?

src/main/java/com/epam/aidial/core/security/AccessService.java
```diff
@@ -52,6 +53,28 @@ public boolean hasWriteAccess(ResourceDescription resource, ProxyContext context
...
+        }
+
+        return ruleService.hasPublicAccess(context, resources);
+    }
+
+    private boolean hasPublicAccess(ResourceDescription resource, ProxyContext context, boolean readOnly) {
```
astsiapanay: Could you please add a javadoc?
These two methods have the same name but do the job differently.
Why can we not marge the methods something like that?
```
  hasPublicAccess(List<Resource> resources, ProxyContext context) {
    for (var resource : resources) {
         if (!hasPublicAcess(resource, context) {
             return false;
         }
   }
   return true;
}
```

src/main/java/com/epam/aidial/core/service/PublicationUtil.java
```diff
@@ -0,0 +1,73 @@
...
+import com.epam.aidial.core.storage.BlobStorageUtil;
+import com.epam.aidial.core.storage.ResourceDescription;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import lombok.experimental.UtilityClass;
+
+import java.util.Map;
+
+@UtilityClass
+public class PublicationUtil {
```
astsiapanay, line 12: Please add javadoc.

src/main/java/com/epam/aidial/core/controller/PublicationController.java
Maxim-Gadalov: url is no longer required

src/main/java/com/epam/aidial/core/data/Publication.java
Maxim-Gadalov: I changed description

src/main/java/com/epam/aidial/core/security/AccessService.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/security/AccessService.java
Maxim-Gadalov: refactored

src/main/java/com/epam/aidial/core/service/PublicationUtil.java
Maxim-Gadalov: added

astsiapanay (approved):

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-343.nightly-test.deltixhub.io
>E2E tests status: success

PR #345, opened 2024-05-28
feat: notification API

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/service/NotificationService.java
```diff
@@ -0,0 +1,82 @@
...
+        });
+
+        return notification;
+    }
+
+    public Set<Notification> listNotification(ProxyContext context) {
+        ResourceDescription notificationResource = getNotificationResource(context, encryptionService);
+        Map<String, Notification> notifications = decodeNotifications(resourceService.getResource(notificationResource));
+
+        return new TreeSet<>(notifications.values());
```
astsiapanay, line 51: `LinkedHashMap` preserves the insertion order. I think we can just return values.

src/main/java/com/epam/aidial/core/service/NotificationService.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public void deleteNotification(ProxyContext context, DeleteNotificationRequest request) {
+        List<String> ids = request.ids();
+        if (ids.isEmpty()) {
+            throw new IllegalArgumentException("Notification IDs cannot be empty");
+        }
+        ResourceDescription notificationResource = getNotificationResource(context, encryptionService);
+        resourceService.computeResource(notificationResource, state -> {
```
astsiapanay, line 60: I think it's worth deleting resource if the object is empty. That would require code changes in the method `computeResource`

src/main/java/com/epam/aidial/core/service/NotificationService.java
Maxim-Gadalov: there is no guarantee that notifications will be inserted in correct sort order

src/main/java/com/epam/aidial/core/service/NotificationService.java
Maxim-Gadalov: notifications are going to be pulled frequently, so it's better to return empty body from redis compare to blob storage look ups

src/main/java/com/epam/aidial/core/service/NotificationService.java
astsiapanay: core returns time to  create notifications to front-end. Front-end expects the notifications to be sorted by time precisely.

src/main/java/com/epam/aidial/core/service/NotificationService.java
astsiapanay: let's keep as is.

astsiapanay (approved):

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-345.nightly-test.deltixhub.io
>E2E tests status: success

Issue #351, opened 2024-06-04
Support defaults for `chat/completions` request parameters in config


adubovik:
There are models which require `max_tokens` request parameter as a mandatory.
However, this field is optional in DIAL API (as well as in OpenAI `chat/completions` API call).

Currently, we set certain reasonable `max_tokens` default directly in the adapters: [bedrock](https://github.com/epam/ai-dial-adapter-bedrock/blob/714f3bb57cbed7af9667cb23a5ed954f193e08f7/aidial_adapter_bedrock/llm/model/conf.py#L1-L5) and [openai](https://github.com/epam/ai-dial-adapter-openai/blob/ce4a06b308123fa5d3d07ccbc87be4ee252d349d/aidial_adapter_openai/gpt4_multi_modal/chat_completion.py#L41).

We propose to support the default values for `chat/completions` request parameters on the DIAL Core side.

- [x] Add `defaults` dictionary to the deployment configuration
```json
"deployment-name": {
    "endpoint": "...",
    "defaults": {
        "max_tokens": 2000, 
        "temperature": 0.38, 
        "top_p": 0.0001,
        "frequency_penalty": 1.12
    },
    "features": ...,
    "limits": ...
}
```
- [x] Implement filling-the-blanks logic:
```python
for field, default_value in deployment.defaults.items():
    if field not in request:
        request[field] = default_value
```
- [x] Expose the `defaults` in the listing
- [x] Document the feature in README

PR #357, opened 2024-06-07
feat: Support defaults for chat/completions request parameters in config #351

astsiapanay:
There are models which require `max_tokens` request parameter as a mandatory.
However, this field is optional in DIAL API (as well as in OpenAI `chat/completions` API call).

Currently, we set certain reasonable `max_tokens` default directly in the adapters: [bedrock](https://github.com/epam/ai-dial-adapter-bedrock/blob/714f3bb57cbed7af9667cb23a5ed954f193e08f7/aidial_adapter_bedrock/llm/model/conf.py#L1-L5) and [openai](https://github.com/epam/ai-dial-adapter-openai/blob/ce4a06b308123fa5d3d07ccbc87be4ee252d349d/aidial_adapter_openai/gpt4_multi_modal/chat_completion.py#L41).

We propose to support the default values for `chat/completions` request parameters on the DIAL Core side.

- [x] Add `defaults` dictionary to the deployment configuration
```json
"deployment-name": {
    "endpoint": "...",
    "defaults": {
        "max_tokens": 2000, 
        "temperature": 0.38, 
        "top_p": 0.0001,
        "frequency_penalty": 1.12
    },
    "features": ...,
    "limits": ...
}
```
- [ ] Implement filling-the-blanks logic:
```python
for field, default_value in deployment.defaults.items():
    if field not in request:
        request[field] = default_value
```
- [x] Expose the `defaults` in the listing
- [x] Document the feature in README


Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-357.nightly-test.deltixhub.io
>E2E tests status: success

sample/aidial.config.json
```diff
@@ -59,7 +59,10 @@
                     "folderAttachmentsSupported": false
                 },
                 "maxInputAttachments": 10,
-                "inputAttachmentTypes": ["type1", "type2"]
+                "inputAttachmentTypes": ["type1", "type2"],
+                "defaults": {
+                    "param1": "value"
```
adubovik: Please add tests not only for string parameters, but for floats and integers as well, because this is what we would actually need to be configured:

```json
    "defaults": {
        "max_tokens": 2000, 
        "temperature": 0.38, 
        "top_p": 0.0001,
        "frequency_penalty": 1.12
    }
```

src/main/java/com/epam/aidial/core/config/Deployment.java
```diff
@@ -22,4 +23,11 @@ public abstract class Deployment {
     private Features features;
     private List<String> inputAttachmentTypes;
     private Integer maxInputAttachments;
+    /**
+     * Default parameters are applied if a request doesn't contain them in OpenAI chat/completions API call.
+     * <p>
+     *     Note. Parameters' values are converted to strings, e.g. 0.25 -> "0.25" or true -> "true"
+     * </p>
+     */
+    private Map<String, String> defaults = Map.of();
```
adubovik: We should generalize from `String` to an arbitrary atomic Object - strings, floats, ints.

astsiapanay: @adubovik is that crucial to map exactly data type, eg float -> float?
the code does float to string and the parameter always comes to the deployment as a string value.

src/main/java/com/epam/aidial/core/function/CollectAttachmentsFn.java
```diff
@@ -0,0 +1,76 @@
...
+import com.epam.aidial.core.util.HttpStatus;
+import com.epam.aidial.core.util.ProxyUtil;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+
+import java.net.URI;
+
+@Slf4j
+public class CollectAttachmentsFn extends BaseFunction<ObjectNode> {
```
Maxim-Gadalov, line 24: please add a javadoc since this class not only collect attachments, but also assign per-request api-key

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -260,26 +226,6 @@ void handleRequestBody(Buffer requestBody) {
...
-            } else {
-                destApiKeyData.getAttachedFiles().add(resourceUrl);
-            }
-        } else {
-            throw new HttpException(HttpStatus.FORBIDDEN, "Access denied to the file %s".formatted(url));
-        }
-    }
-
     @SneakyThrows
     private ResourceDescription getResourceDescription(String url) {
```
Maxim-Gadalov: is it still needed?

sample/aidial.config.json
astsiapanay: fixed

src/main/java/com/epam/aidial/core/config/Deployment.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: No, it's not. removed

src/main/java/com/epam/aidial/core/function/CollectAttachmentsFn.java
astsiapanay: fixed

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-357.nightly-test.deltixhub.io
>E2E tests status: success

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-357.nightly-test.deltixhub.io
>E2E tests status: success

Maxim-Gadalov (approved):

PR #369, opened 2024-06-13
feat: add name attribute to the publication request

Maxim-Gadalov:


src/main/java/com/epam/aidial/core/data/Publication.java
```diff
@@ -19,6 +19,7 @@ public class Publication {
     /**
      * Target directory prefix to publish; for example: public/ or public/folder/.
      */
+    String name;
```
astsiapanay: Javadoc description is wrong. It points out to the field name instead of targetFolder

Maxim-Gadalov: fixed

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-369.nightly-test.deltixhub.io
>E2E tests status: success

astsiapanay (approved):

PR #384, opened 2024-06-21
chore: improve logging

astsiapanay:


src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -61,8 +61,8 @@ protected Future<?> handle(ResourceDescription descriptor) {
         if (context.getRequest().method() == HttpMethod.DELETE) {
             return deleteResource(descriptor);
         }
-
-        return context.respond(HttpStatus.BAD_GATEWAY, "No route");
+        log.warn("Unsupported HTTP method for accessing resource {}", descriptor.getAbsoluteFilePath());
```
Maxim-Gadalov: it's better to log resource url

astsiapanay: fixed

Maxim-Gadalov (approved):

PR #394, opened 2024-07-16
feat: subscription for resources

artsiomkorzun:


src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
```diff
@@ -13,25 +19,35 @@
...
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
 @Slf4j
 public class ResourceOperationController {
 
+    private static final Set<ResourceType> SUBSCRIPTION_ALLOWED_TYPES = Set.of(
```
Maxim-Gadalov, line 34: files is not supported

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
```diff
@@ -84,9 +100,75 @@ public Future<?> move() {
         return Future.succeededFuture();
     }
 
+    public Future<?> subscribe() {
+        HttpServerResponse response = context.getResponse();
+        Consumer<ResourceEvent> subscriber = event -> {
```
Maxim-Gadalov: consider moving to the separate method

src/main/java/com/epam/aidial/core/service/ResourceTopic.java
```diff
@@ -0,0 +1,87 @@
...
+            });
+        }
+    }
+
+    private void handle(ResourceEvent event) {
+        for (Subscription subscription : urlToSubscriptions.getOrDefault(event.getUrl(), Set.of())) {
+            try {
+                subscription.subscriber.accept(event);
+            } catch (Throwable e) {
+                log.warn("Can't notify about resource event", e);
```
Maxim-Gadalov: Failed to send event?

src/test/java/com/epam/aidial/core/ResourceBaseTest.java
```diff
@@ -231,6 +241,19 @@ static void verifyJsonNotExact(Response response, int status, String body) {
         }
     }
 
+    static void verifyJsonNotExact(String expected, String actual) {
```
Maxim-Gadalov, line 235: cleanup above method to:
 ```
static void verifyJsonNotExact(Response response, int status, String body) {
        assertEquals(status, response.status());
        verifyJsonNotExact(body, response.body());
    }
 ```

src/main/java/com/epam/aidial/core/service/ResourceTopic.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+
+
+@Slf4j
+public class ResourceTopic {
+
+    private final Map<String, Set<Subscription>> urlToSubscriptions = new ConcurrentHashMap<>();
```
Maxim-Gadalov, line 23: keys are never evicted

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
artsiomkorzun: Disabled for now, but they will be in the next MR.

src/main/java/com/epam/aidial/core/service/ResourceTopic.java
artsiomkorzun: It takes an interface subscriber. It has nothing to do with sending events. Plus this one is unreachable code, cause the sending code catches exceptions.

src/test/java/com/epam/aidial/core/ResourceBaseTest.java
artsiomkorzun: done

src/main/java/com/epam/aidial/core/service/ResourceTopic.java
artsiomkorzun: Check Map contract and ConcurrentHashMap implementation.

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
artsiomkorzun: I will leave it as is, it is a pretty small piece of code.

artsiomkorzun: After considering I moved it into a method and added permission check since the permissions could change during long sse connection

src/main/java/com/epam/aidial/core/service/ResourceTopic.java
Maxim-Gadalov: add a comment about this catch block

Maxim-Gadalov (dismissed):

Maxim-Gadalov (approved):

PR #397, opened 2024-07-18
fix: use the result of AccessControlBaseController#handle

artsiomkorzun:


Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-397.nightly-test.deltixhub.io
>E2E tests status: success

src/main/java/com/epam/aidial/core/Proxy.java
```diff
@@ -101,8 +102,19 @@ public void handle(HttpServerRequest request) {
     }
 
     private void handleError(Throwable error, HttpServerRequest request) {
-        log.error("Can't handle request", error);
-        respond(request, HttpStatus.INTERNAL_SERVER_ERROR);
+        if (!request.response().ended()) {
+            HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
+            String message = null;
```
Maxim-Gadalov, line 107: this may lead to NPE

src/main/java/com/epam/aidial/core/ProxyContext.java
```diff
@@ -124,7 +124,8 @@ public Future<Void> respond(HttpStatus status, String body) {
                     body.length() > LOG_MAX_ERROR_LENGTH ? body.substring(0, LOG_MAX_ERROR_LENGTH) : body);
         }
 
-        return response.setStatusCode(status.getCode()).end(body);
+        response.setStatusCode(status.getCode()).end(body);
+        return Future.succeededFuture();
```
Maxim-Gadalov: why so? why do we need to succeed future before response body actually written

src/main/java/com/epam/aidial/core/controller/DownloadFileController.java
```diff
@@ -29,40 +29,38 @@ protected Future<?> handle(ResourceDescription resource, boolean hasWriteAccess)
...
+                        return Future.succeededFuture();
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                }).onFailure(error -> context.respond(HttpStatus.INTERNAL_SERVER_ERROR,
+                        "Failed to fetch file with path %s/%s".formatted(resource.getBucketName(),
+                                resource.getOriginalPath())));
 
-        return result.future();
+        return Future.succeededFuture();
```
Maxim-Gadalov, line 64: same here, previous implementation succeed future only after sending file content to the response

src/main/java/com/epam/aidial/core/Proxy.java
artsiomkorzun: fixed

src/main/java/com/epam/aidial/core/controller/DownloadFileController.java
artsiomkorzun: Check the contract of handle.

Maxim-Gadalov (approved):

Issue #393, opened 2024-07-16
TokenUsage stats is not calculated properly for apps when Core have multiple instances


astsiapanay:
We added scalability support to DIAL Core a while ago by implementing per-request API Keys are hosted in Redis.

However the code to calculate cost stats for apps still relies on a single Core instance.

PR #402, opened 2024-07-23
fix: TokenUsage stats is not calculated properly for apps when Core have multiple instances #393

astsiapanay:


astsiapanay:
/deploy-review
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/497768)

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-402.nightly-test.deltixhub.io
>E2E tests status: success

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+    private final Vertx vertx;
+
+    private final LongSupplier clock;
+
+    private final String trackerId;
+
+    public TokenStatsTracker(RedissonClient redis, Vertx vertx, LongSupplier clock) {
+        this.clock = clock;
+        this.vertx = vertx;
+        this.topic = redis.getTopic("deployment-cost-stats-event", new StringCodec());
```
artsiomkorzun: Please use prefix for the key as the other components do.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
 public class TokenStatsTracker {
+
+
     private final Map<String, TraceContext> traceIdToContext = new ConcurrentHashMap<>();
 
-    public void startSpan(ProxyContext context) {
+    /**
+     * List of registered trackers in the cluster
+     */
+    private final Map<String, TrackerData> registeredTrackers = new ConcurrentHashMap<>();
```
artsiomkorzun: In case one backend failed, not ending span and not publishing end of span event, how do we clean this map?

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.LongSupplier;
 
+@Slf4j
 public class TokenStatsTracker {
+
+
     private final Map<String, TraceContext> traceIdToContext = new ConcurrentHashMap<>();
```
artsiomkorzun: In case one of backends fails not publishing an end of span event, how do we clean this map?

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+            default -> log.warn("Unsupported event {}", event);
+        }
+    }
+
+    private void handlePing(PingEvent event) {
+        if (this.trackerId.equals(event.trackerId)) {
+            return;
+        }
+        TrackerData trackerData = getTrackerData(event.trackerId);
+        trackerData.lastUpdateTs = clock.getAsLong();
```
artsiomkorzun, line 114: concurrent access here and cleanupDeadTrackers

artsiomkorzun: seems like not a problem

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+
+    private void cleanupDeadTrackers() {
+        long currentTime = clock.getAsLong();
+        for (Map.Entry<String, TrackerData> entry : registeredTrackers.entrySet()) {
+            TrackerData trackerData = entry.getValue();
+            String trackerId = entry.getKey();
+            if (currentTime - trackerData.lastUpdateTs > 60_000) {
+                // complete all promises forcibly
+                trackerData.completePromises();
+                registeredTrackers.remove(trackerId);
```
artsiomkorzun, line 85: race? You can complete/clear all promises, then addSpan can add one more, only then you will remove and the promise will never end

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+        }
+    }
+
+    private void handleStartSpan(StartSpanEvent event) {
+        TraceContext traceContext = traceIdToContext.get(event.traceId);
+        if (traceContext == null) {
+            return;
+        }
+        Promise<TokenUsage> promise = traceContext.updateSpan(event);
+        if (promise != null && !trackerId.equals(event.trackerId)) {
```
artsiomkorzun, line 157: Seems like we can add global filter to onEvent to skip events from "me"

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -28,45 +140,79 @@ public Future<TokenUsage> getTokenStats(ProxyContext context) {
...
+            TokenStats tokenStats = spans.get(event.parentSpanId);
+            if (tokenStats == null) {
+                return null;
+            }
+            Promise<TokenUsage> promise = Promise.promise();
+            tokenStats.children.put(event.spanId, promise);
+            return promise;
+        }
+
+        synchronized Promise<TokenUsage> updateSpan(EndSpanEvent event) {
```
artsiomkorzun: Span could end with error (for some reason even though we get successful event?), how do we know when to end with error?

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-402.nightly-test.deltixhub.io
>E2E tests status: success

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -28,45 +140,79 @@ public Future<TokenUsage> getTokenStats(ProxyContext context) {
...
-            TokenStats tokenStats = new TokenStats(new TokenUsage(), Promise.promise(), new ArrayList<>());
+            TokenStats tokenStats = new TokenStats(new TokenUsage(), new HashMap<>());
             spans.put(spanId, tokenStats);
             if (parentSpanId != null) {
-                TokenStats parent = spans.get(parentSpanId);
-                if (parent != null) {
-                    parent.children.add(tokenStats.promise.future());
-                }
+                StartSpanEvent event = new StartSpanEvent(trackerId, context.getTraceId(), parentSpanId, context.getSpanId());
+                return publish(Event.START_SPAN, event);
```
artsiomkorzun, line 230: With all this future code it is hard to tell if this is possible:
1) Flow: 1 -> 2 -> 3. -> means edge, so 1 is parent span for 2.
2) Real publication order: 1 -> 3 -> 2.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: No, we can’t.  We should accept events in case of a single core or when parent and child spans are on the same instance

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: Every tracker has own unique ID if the tracker is dead the code completes all promises. A new instance with a **different** ID can add more span.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+        }
+        TrackerData trackerData = getTrackerData(event.trackerId);
+        trackerData.lastUpdateTs = clock.getAsLong();
+    }
+
+    private TrackerData getTrackerData(String trackerId) {
+        return registeredTrackers.computeIfAbsent(trackerId, key -> new TrackerData(clock.getAsLong()));
+    }
+
+    private void handleEndSpan(EndSpanEvent event) {
```
Maxim-Gadalov: please add javadoc/comment how this method behaves for caller (producer) and consumer (other core instances)

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,22 +2,134 @@
...
+            return;
+        }
+        Promise<TokenUsage> promise = traceContext.updateSpan(event);
+        if (promise != null && !trackerId.equals(event.trackerId)) {
+            TrackerData trackerData = getTrackerData(event.trackerId);
+            trackerData.removePromise(promise);
+        }
+    }
+
+    private void handleStartSpan(StartSpanEvent event) {
```
Maxim-Gadalov: please add javadoc/comment how this method behaves for caller (producer) and consumer (other core instances)

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: Right. The block should be synchronized. I'll fix it.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -28,45 +140,79 @@ public Future<TokenUsage> getTokenStats(ProxyContext context) {
...
             }
+            String parentSpanId = context.getParentSpanId();
+            if (parentSpanId != null) {
+                EndSpanEvent event = new EndSpanEvent(trackerId, context.getTraceId(), parentSpanId, context.getSpanId(), context.getTokenUsage());
+                return publish(Event.END_SPAN, event);
+            }
+            return Future.succeededFuture();
         }
 
         synchronized Future<TokenUsage> getStats(ProxyContext context) {
```
Maxim-Gadalov: please add javadoc/comment how and when this method should be called

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: In case of error the core send an empty tokenUsage: e.g. bug in the cost calculator. So the code treats all results as successful. This is how it works now. We always complete a promise successfully.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: I can't imagine the case.
Core aways startSpan parent span before child span. So events will come in the order they were sent to the topic.
The order of End span events doesn't matter.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: You mean a case when the Core is alive but publishing doesn't work for some reason, e.g. bug in the code or Redis is not functional. Right?

We don't clean the map in this case.

In my opinion that's a bug in Core or infrastructure issue. Those traces will stay forever.

If you believe that's a strong case we can introduce a job that scans the map to find obsolete traces and remove them after some timeout, e.g. 1 min.

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: will fix

astsiapanay: fixed

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: fixed

astsiapanay:
The main issue of that approach is timeline of receiving events from Redis topic might be no-determenistic. For example
startSpanEvent might arrive after parentSpan collected stats from dependent spans. That will cause the issue that parent span doesn't have any dependent spans because of even came later.

PR #407, opened 2024-07-25
fix: TokenUsage stats is not calculated properly for apps when Core have multiple instances #393

astsiapanay:


src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -404,13 +415,21 @@ void handleResponse() {
...
             }
         } else {
             tokenUsageFuture = proxy.getTokenStatsTracker().getTokenStats(context).andThen(result -> context.setTokenUsage(result.result()));
         }
 
         tokenUsageFuture.onComplete(ignore -> {
 
+            HttpServerResponse response = context.getResponse();
+            if (responseStream.getLastChunk() != null) {
+                response.end(responseStream.getLastChunk());
```
artsiomkorzun: hide into responseStream.finish() or end()

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,89 +2,147 @@
...
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
+import static com.epam.aidial.core.storage.BlobStorageUtil.PATH_SEPARATOR;
+
+@Slf4j
+@RequiredArgsConstructor
 public class TokenStatsTracker {
```
artsiomkorzun, line 23: side track: seems like current algo can leave junk on storage

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -69,6 +69,13 @@ public DeploymentPostController(Proxy proxy, ProxyContext context) {
         this.context = context;
         this.applicationService = proxy.getCustomApplicationService();
         this.enhancementFunctions = List.of(new CollectAttachmentsFn(proxy, context),
+                new BaseFunction<>(proxy, context) {
+                    @Override
+                    public Throwable apply(ObjectNode tree) {
+                        context.setStreamingRequest(tree.get("stream").asBoolean(false));
```
artsiomkorzun: Do you really need this flag?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
```diff
@@ -111,7 +118,33 @@ public synchronized ReadStream<Buffer> endHandler(Handler<Void> handler) {
 
     private synchronized void handleChunk(Buffer chunk) {
         content.appendBuffer(chunk);
-        notifyOnChunk(chunk);
+        if (isStreaming && isLastChunk(content)) {
+            lastChunk = chunk;
```
artsiomkorzun: We should stop streaming, but still accumulating buffer. At the end we should send the tail. This approach will work for bad response like:
```
data: [DONE]
error: NOT DONE, ha ha
```

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -340,8 +350,9 @@ private void handleProxyResponse(HttpClientResponse proxyResponse) {
             return;
         }
 
+        boolean isStreaming = context.getDeployment() instanceof Model && context.isStreamingRequest();
```
artsiomkorzun: Why only model?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
artsiomkorzun: IF you really need it, please extract a class

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: we need to collect last chunk for models only

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,89 +2,147 @@
...
-        if (traceContext == null) {
-            return Future.succeededFuture();
-        }
-        return traceContext.getStats(context);
+        return vertx.executeBlocking(() -> {
+            ResourceDescription resource = toResource(context.getTraceId());
+            String json = resourceService.getResource(resource);
+            TraceContext traceContext = ProxyUtil.convertToObject(json, TraceContext.class);
+            if (traceContext == null) {
+                return null;
```
artsiomkorzun, line 57: Do you handle it correctly everywhere?

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
```diff
@@ -2,89 +2,147 @@
...
+    /**
+     * Ends current span.
+     */
+    public Future<Void> endSpan(ProxyContext context) {
         ApiKeyData apiKeyData = context.getApiKeyData();
         if (apiKeyData.getPerRequestKey() == null) {
-            traceIdToContext.remove(context.getTraceId());
+            return vertx.executeBlocking(() -> {
+                ResourceDescription resource = toResource(context.getTraceId());
+                resourceService.deleteResource(resource, EtagHeader.ANY);
```
artsiomkorzun, line 71: Does it work for this case:

Application uses two models?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: yes, we need it when constructing buffered stream for response

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
astsiapanay: fixed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
artsiomkorzun: Why application streaming is so different?

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: yes, we do

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: stats for applications are derived from model stats we don't need to wait for last chunk of the app response

artsiomkorzun: Sounds like a good comment at this place

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: right. that may happen if root span is not completed for some reason: pod restart or something else

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/token/TokenStatsTracker.java
astsiapanay: yes, we delete trace for root requests only, e.g. chat back end -> app. Core knows that request doesn't have any per request API key with that request so we are safe to delete. Otherwise app may make calls to two model A and B and we have to wait for the request to the app is completed and then delete the trace.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
```diff
@@ -116,7 +117,19 @@ public synchronized ReadStream<Buffer> endHandler(Handler<Void> handler) {
...
+        } else {
+            response.end();
+        }
+    }
+
     private synchronized void handleChunk(Buffer chunk) {
+        if (lastChunk != null) {
+            // stop streaming
+            return;
+        }
```
artsiomkorzun, line 136: stop streaming  == stop calling handler

artsiomkorzun: remember position at which your stopped streaming and send the tail in end method

astsiapanay: fixed

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
```diff
@@ -24,7 +27,7 @@ public class BufferingReadStream implements ReadStream<Buffer> {
     private boolean ended;
     private boolean reset;
     private final boolean isStreaming;
-    private Buffer lastChunk;
+    private List<Buffer> lastChunks;
```
artsiomkorzun: just remember position and make a slice from content buffer

astsiapanay: fixed

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
```diff
@@ -109,9 +118,49 @@ public synchronized ReadStream<Buffer> endHandler(Handler<Void> handler) {
...
+        } else {
+            notifyOnChunk(chunk);
+        }
+    }
+
+    private static boolean isLastChunk(Buffer content) {
+        int i = skipWhitespaces(content, content.length() - 1);
+        String lastMessage = "data: [DONE]";
+        int j = lastMessage.length() - 1;
+        for (; i >= 0 && j >= 0; i--, j--) {
```
artsiomkorzun, line 148: Gives true for partially matched content: e.g. ], ONE]

artsiomkorzun: Check that the content has enough bytes to match

astsiapanay: fixed

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
```diff
@@ -109,9 +118,49 @@ public synchronized ReadStream<Buffer> endHandler(Handler<Void> handler) {
...
+        if (isStreaming && isLastChunk(content)) {
+            lastChunkPos = pos;
+        } else {
+            notifyOnChunk(chunk);
+        }
+    }
+
+    private static boolean isLastChunk(Buffer content) {
+        int i = skipWhitespaces(content, content.length() - 1);
+        String lastMessage = "data: [DONE]";
```
Maxim-Gadalov, line 146: Nit; extract a constant

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-407.nightly-test.deltixhub.io
>E2E tests status: failed

src/main/java/com/epam/aidial/core/util/BufferingReadStream.java
astsiapanay: Let’s keep as is

PR #418, opened 2024-08-01
feat: return createdAt/updatedAt for files

### LINKED ISSUE #417: Return createdAt/updatedAt time for files

Oleksii-Klimov:


src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -740,6 +757,19 @@ public String getEtag(ResourceDescription descriptor) {
         return metadata.getEtag();
     }
 
+    private static Map<String, String> toUserMetadata(String etag, Long createdAt, Long updatedAt) {
+        Map<String, String> metadata = new HashMap<>();
+        metadata.put(ResourceUtil.ETAG_ATTRIBUTE, etag);
+        if (createdAt != null) {
+            metadata.put(ResourceUtil.CREATED_AT_ATTRIBUTE, Long.toString(createdAt));
```
artsiomkorzun, line 764: Seems like they are always non nullable.

artsiomkorzun (approved):

src/main/java/com/epam/aidial/core/service/ResourceService.java
Oleksii-Klimov: I can't guarantee it. The 'updatedAt' field seems unlikely to be null, but 'createdAt' may be null. For instance, 'createdAt' is not returned by the local filesystem blob store, and possibly by some other blob stores as well.

Issue #292, opened 2024-03-27
feature request: advanced load balancing


justrp:
Current load balancing is using round robin strategy. I'd like to have more intelligent balancing strategies:
* weights-based
* latency-based (technically, it might be somehow implemented via weights controlled by external infrastructure)
* tokens capacity/demand based (different endpoints might have different capacity - but that also might be solved by weights-based)

Some alternative solutions already claim to have this (e.g. https://github.com/microsoft/AICentral)

PR #422, opened 2024-08-05
feat: implement advanced load balancing

Maxim-Gadalov:
#292


Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-422.nightly-test.deltixhub.io
>E2E tests status: failed

README.md
```diff
@@ -221,7 +221,7 @@ Dynamic settings can include the following parameters:
...
... Model endpoint.<br />`key`: Your API key.<br />`weight`: Weight for upstream endpoint; positive number represents an endpoint weight, zero or negative disables this enpoint from routing. Default value: 1.<br />`tier`: Specifies tier group for the endpoint. Only positive numbers allowed. All requests will be routed to the endpoints with hieghest tier, other endpoints (with lower tier) may be used if and only if hiehest tier endpoints are unavailable. Default value: 1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
```
artsiomkorzun: Any chance to make it backward compatible to work with the previous config?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -329,17 +329,27 @@ void handleProxyRequest(HttpClientRequest proxyRequest) {
...
+        if (isRetriableError(responseStatusCode)) {
+            upstreamRoute.failed(HttpStatus.fromStatusCode(responseStatusCode), UpstreamRoute.calculateRetryAfterSeconds(proxyResponse));
+            // get next upstream
+            upstreamRoute.next();
             sendRequest(); // try next
             return;
         }
 
+        if (responseStatusCode == 200) {
+            context.getUpstreamRoute().succeed();
```
artsiomkorzun: Use upstreamRoute variable

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -329,17 +329,27 @@ void handleProxyRequest(HttpClientRequest proxyRequest) {
...
+        }
+
```
artsiomkorzun, line 360: What do you do when error occurred but it is not retriable?

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+    public synchronized UpstreamRoute get(UpstreamProvider provider) {
+        String deploymentName = provider.getName();
+        List<Upstream> upstreams = provider.getUpstreams();
+        TieredBalancer balancer = cache.getIfPresent(deploymentName);
+        if (balancer == null) {
+            balancer = new TieredBalancer(deploymentName, upstreams);
+            cache.put(deploymentName, balancer);
+        } else {
+            // update upstreams (rebuild balancer if needed) in case of upstreams changed
+            balancer.updateUpstreams(upstreams);
```
artsiomkorzun: It really should be done on config change, maybe it is worth investing time to add some code into ConfigProvider to notify this component about a change.

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import com.epam.aidial.core.config.Upstream;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class LoadBalancerProvider {
+
+    private final Cache<String, TieredBalancer> cache = CacheBuilder.newBuilder()
```
artsiomkorzun: Can be evicted before some time-related state?

src/main/java/com/epam/aidial/core/upstream/TieredBalancer.java
```diff
@@ -0,0 +1,82 @@
...
+            // rebuild balancer
+            originalUpstreams = upstreams;
+            tiers = buildTiers(this.deploymentName, upstreams);
+
+        } finally {
+            writeLock.unlock();
+        }
+    }
+
+    private static List<WeightedRoundRobinBalancer> buildTiers(String deploymentName, List<Upstream> upstreams) {
```
artsiomkorzun, line 44: Is it an adaptive rebuild and just let's start from scratch?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
     public Upstream next() {
-        if (hasNext()) {
-            prev = next++;
-            int index = (offset + prev) % upstreams.size();
-            upstream = upstreams.get(index);
-            return upstream;
-        }
-
-        return null;
+        retries++;
```
artsiomkorzun: Do you need to check for retries < maxRetries here?

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
+     * Fail current upstream due to error
+     *
+     * @param status - response http status; typically, 5xx or 429
+     * @param retryAfterSeconds - the amount of seconds after which upstream should be available; if status 5xx this value ignored
      */
-    public void retry() {
-        if (prev < next && retries <= maxRetries) {
-            retries++;
-            next = prev;
+    public void failed(HttpStatus status, long retryAfterSeconds) {
```
artsiomkorzun: fail or onFailure or onFailed

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
+        if (upstreamState == null) {
+            return;
         }
+        upstreamState.failed(status, retryAfterSeconds);
+    }
+
+    public void succeed() {
```
artsiomkorzun, line 113: to match with succeed (cause failed <-> succeeded)

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
+        }
+        upstreamState.succeeded();
+    }
+
+    /**
+     * @param response http response from upstream
+     * @return the amount of seconds after which upstream should be available
+     */
+    public static long calculateRetryAfterSeconds(HttpClientResponse response) {
+        long retryAfterSeconds = 30;
```
artsiomkorzun: extract a constant or a config setting

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
+        String retryAfterHeaderValue = response.getHeader("Retry-After");
+        if (retryAfterHeaderValue != null) {
+            retryAfterSeconds = Long.parseLong(retryAfterHeaderValue);
```
artsiomkorzun: Catch error

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,112 @@
...
+    private int currentWeight;
+
+    public WeightedRoundRobinBalancer(String deploymentName, List<Upstream> upstreams) {
+        this.deploymentName = deploymentName;
+        if (upstreams == null || upstreams.isEmpty()) {
+            throw new IllegalArgumentException("Upstream list is null or empty");
+        }
+        int tier = upstreams.get(0).getTier();
+        for (Upstream upstream : upstreams) {
+            if (upstream.getTier() != tier) {
```
artsiomkorzun, line 31: Maybe I am missing something, but why must they be the same?

src/main/java/com/epam/aidial/core/upstream/TieredBalancer.java
```diff
@@ -0,0 +1,82 @@
...
+        List<WeightedRoundRobinBalancer> balancers = new ArrayList<>();
+        Map<Integer, List<Upstream>> groups = upstreams.stream()
+                .collect(Collectors.groupingBy(Upstream::getTier));
+
+        for (Map.Entry<Integer, List<Upstream>> entry : groups.entrySet()) {
+            balancers.add(new WeightedRoundRobinBalancer(deploymentName, entry.getValue()));
+        }
+
+        balancers.sort(Comparator.reverseOrder());
```
artsiomkorzun: Usually tiered something means from hottest to coldest. For example, tiered storage:

0) CPU CACHE
1) RAM
2) NVME
3) SSD
4) HHD

That means that the tiers with the lowest number is going to be used first. But here it is different, any reason behind it?

README.md
Maxim-Gadalov: if no weight and tier provided it will work as previous one

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,112 @@
...
+            if (currentIndex == 0) {
+                currentWeight -= gcdWeight;
+                if (currentWeight <= 0) {
+                    currentWeight = maxWeight;
+                    if (currentWeight == 0) {
+                        return null;
+                    }
+                }
+            }
+            if (upstreams.get(currentIndex).getUpstream().getWeight() >= currentWeight) {
```
astsiapanay: It doesn't make sense to keep iterating over the list: just jump to the first element.

10, 8, 6, 4, 2

we stay at index = 1, with current weight = 10. So we can jump directly to the begining:

currentIndex = -1;

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -0,0 +1,69 @@
...
+import lombok.Getter;
+
+public class UpstreamState implements Comparable<UpstreamState> {
+
+    @Getter
+    private final Upstream upstream;
+    private final int errorsThreshold;
+
+    private static final long INITIAL_BACKOFF_DELAY_MS = 1000;
+    private static final int BACKOFF_MULTIPLIER = 2;
```
astsiapanay: BACKOFF_BASE

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param retryAfterSeconds time in seconds when upstream may become available; only take into account with 429 status code
+     */
+    public synchronized void failed(HttpStatus status, long retryAfterSeconds) {
+        if (status == HttpStatus.TOO_MANY_REQUESTS) {
+            retryAfter = System.currentTimeMillis() + retryAfterSeconds * 1000;
+        }
+
+        if (status.is5xx()) {
+            if (++errorCount >= errorsThreshold) {
+                retryAfter = System.currentTimeMillis() + (long) (INITIAL_BACKOFF_DELAY_MS * Math.pow(BACKOFF_MULTIPLIER, errorCount));
```
astsiapanay: as an option

(INITIAL_BACKOFF_DELAY_MS * (1 << errorCount))

src/main/java/com/epam/aidial/core/upstream/LoadBalancer.java
```diff
@@ -0,0 +1,5 @@
+package com.epam.aidial.core.upstream;
+
+public interface LoadBalancer<T> {
+    T get();
```
astsiapanay: The method name is not suitable for the current use case. ‘Next’ sounds better

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
astsiapanay: The balancer state is not synchronised across Core instances. Should we keep it in the storage?

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+            .expireAfterAccess(5, TimeUnit.MINUTES)
```
astsiapanay: We shouldn’t evict entries after 5 mins: consider back off is greater than 5 mins

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
astsiapanay: More over custom application config might have changed

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,32 +1,53 @@
...
+        this.maxRetries = maxRetries;
+        this.upstreamState = balancer.get();
+    }
 
     /**
      * @return the number of upstreams which returned any http response.
      */
     public int used() {
-        return next;
+        return retries + 1;
```
astsiapanay: I think the method returns a wrong result or the description should be changed

README.md
astsiapanay: It's still unclear what's a meaning of endpoint's weight. How a user should treat a weight value.

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import javax.annotation.Nullable;
+
+@Getter
+@Slf4j
+public class WeightedRoundRobinBalancer implements Comparable<WeightedRoundRobinBalancer>, LoadBalancer<UpstreamState> {
```
astsiapanay, line 15: It's worth adding description of balancer logic with examples.

src/main/java/com/epam/aidial/core/config/Upstream.java
```diff
@@ -1,9 +1,18 @@
...
 @Data
+@AllArgsConstructor
+@NoArgsConstructor
 public class Upstream {
+
+    public static final int ERROR_THRESHOLD = 3;
+
     private String endpoint;
     private String key;
+    private int weight = 1;
```
astsiapanay, line 16: Is a weight value something relative or absolute? How should we treat them?

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,112 @@
...
+                .sorted(Comparator.reverseOrder())
+                .toList();
+        if (this.upstreams.isEmpty()) {
+            log.warn("No available upstreams for deployment %s and tier %d".formatted(deploymentName, tier));
+        }
+    }
+
+    @Nullable
+    @Override
+    public synchronized UpstreamState get() {
```
astsiapanay: Is there somewhere a reference of the algorithm?

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Comparator;
+import java.util.List;
+import javax.annotation.Nullable;
+
+@Getter
+@Slf4j
+public class WeightedRoundRobinBalancer implements Comparable<WeightedRoundRobinBalancer>, LoadBalancer<UpstreamState> {
+
+    private final int tier;
+    private final List<UpstreamState> upstreams;
```
astsiapanay, line 18: Should we distribute load across upstreams in proportion to their weights?

For example there are 3 upstreams with weights: 2, 3 and 5. That means if there are 10 requests each upstream will take 2, 3 and 5 requests respectively. Right?

So we treat each weight as 20%, 30% and 50% in the example above.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -36,41 +57,65 @@ public int retries() {
...
-    @Override
-    public boolean hasNext() {
-        return next < upstreams.size() && retries <= maxRetries;
+    /**
+     * Checks if upstream present (not null) and retry count does not exceed max retry attempts
+     *
+     * @return true if upstream available, false otherwise
+     */
+    public boolean available() {
+        return upstreamState != null && retries <= maxRetries;
```
astsiapanay: retries are monotonically grown and after some time upstream becomes unavailable due to the condition: `retries <= maxRetries` is false

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
astsiapanay: The also doesn't work properly if weights are prime numbers. See the test below

```
public void test() {
        Upstream upstream1 = new Upstream();
        upstream1.setWeight(59);
        upstream1.setTier(1);
        upstream1.setEndpoint("1");
        Upstream upstream2 = new Upstream();
        upstream2.setWeight(53);
        upstream2.setTier(1);
        upstream2.setEndpoint("2");
        Upstream upstream3 = new Upstream();
        upstream3.setWeight(47);
        upstream3.setTier(1);
        upstream3.setEndpoint("3");
        WeightedRoundRobinBalancer balancer = new WeightedRoundRobinBalancer("app", List.of(upstream1, upstream2, upstream3));
        Map<Upstream, Integer> map = new HashMap<>();
        for (int i = 0; i < 10; i++) {
            UpstreamState upstreamState = balancer.get();
            assertNotNull(upstreamState, "failed at " + i);
            map.put(upstreamState.getUpstream(), map.getOrDefault(upstreamState.getUpstream(), 0) + 1);
        }
        for (var e : map.entrySet()) {
            System.out.println(e.getKey().getEndpoint() + " -> " + e.getValue());
        }
    }
```

The  actual output is
```
1 -> 8
2 -> 2
```

but should be
```
1 -> 4
2 -> 3
3 -> 3
```

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -329,17 +329,29 @@ void handleProxyRequest(HttpClientRequest proxyRequest) {
...
                 context.getTraceId(), context.getSpanId(),
                 context.getProject(), context.getDeployment().getName(),
-                context.getDeployment().getEndpoint(), context.getUpstreamRoute().get().getEndpoint(),
+                context.getDeployment().getEndpoint(), currentUpstream == null ? "N/A" : currentUpstream.getEndpoint(),
                 proxyResponse.statusCode(), proxyResponse.headers().size());
 
-        if (context.getUpstreamRoute().hasNext() && isRetriableError(proxyResponse.statusCode())) {
+        int responseStatusCode = proxyResponse.statusCode();
+        if (isRetriableError(responseStatusCode)) {
+            upstreamRoute.fail(HttpStatus.fromStatusCode(responseStatusCode), UpstreamRoute.calculateRetryAfterSeconds(proxyResponse));
```
astsiapanay: Can we move the logic to calculate retry time to the method ‘fail’?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -458,13 +473,15 @@ private void handleProxyConnectionError(Throwable error) {
...
     private void handleProxyResponseError(Throwable error) {
+        UpstreamRoute upstreamRoute = context.getUpstreamRoute();
         log.warn("Proxy failed to receive response header from origin. Trace: {}. Span: {}. Key: {}. Deployment: {}. Address: {}. Error:",
                 context.getTraceId(), context.getSpanId(),
                 context.getProject(), context.getDeployment().getName(),
                 context.getProxyRequest().connection().remoteAddress(),
                 error);
 
-        context.getUpstreamRoute().retry();
+        upstreamRoute.fail(HttpStatus.BAD_GATEWAY, -1);
```
astsiapanay: It seems we need explanation here: why we put -1

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        if (previous != null && isUpstreamsTheSame(upstream, previous)) {
+            newState.put(name, previous);
+        } else {
+            newState.put(name, new TieredBalancer(name, upstream.getUpstreams()));
+        }
+    }
+
+    private static boolean isUpstreamsTheSame(UpstreamProvider upstreamProvider, TieredBalancer balancer) {
+        return new HashSet<>(upstreamProvider.getUpstreams()).equals(new HashSet<>(balancer.getOriginalUpstreams()));
```
astsiapanay, line 117: upstreamProvider.getUpstreams().containsAll(new HashSet<>(balancer.getOriginalUpstreams()))?

src/main/java/com/epam/aidial/core/upstream/LoadBalancer.java
```diff
@@ -0,0 +1,5 @@
+package com.epam.aidial.core.upstream;
+
+public interface LoadBalancer<T> {
```
astsiapanay, line 3: Please javadoc

src/main/java/com/epam/aidial/core/upstream/TieredBalancer.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.Getter;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+public class TieredBalancer implements LoadBalancer<UpstreamState> {
```
astsiapanay, line 17: Please add javadoc

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,76 +1,124 @@
...
     }
 
     /**
-     * @return the number of retries due to connection errors.
+     * @return the number of used upstreams.
      */
-    public int retries() {
-        return Math.min(retries, maxRetries);
+    public int used() {
+        return used + 1;
```
astsiapanay: I still don't get it why we return used + 1 instead of just the actual number of used?

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
```diff
@@ -0,0 +1,111 @@
...
+        String deploymentName = provider.getName();
+        List<Upstream> upstreams = provider.getUpstreams();
+
+        TieredBalancer balancer = balancers.get(deploymentName);
+        if (balancer == null) {
+            // if no state found for upstream, it's probably custom application
+            balancer = new TieredBalancer(deploymentName, upstreams);
+        }
+
+        return new UpstreamRoute(balancer, 5);
```
astsiapanay: Where does the number 5 come from? Shouldn't we use the number of upstreams instead?

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -329,17 +329,29 @@ void handleProxyRequest(HttpClientRequest proxyRequest) {
...
+            upstreamRoute.fail(HttpStatus.fromStatusCode(responseStatusCode), UpstreamRoute.calculateRetryAfterSeconds(proxyResponse));
+            // get next upstream
+            upstreamRoute.next();
             sendRequest(); // try next
             return;
         }
 
+        if (responseStatusCode == 200) {
+            upstreamRoute.succeed();
+        } else if (isFailedStatusCode(responseStatusCode)) {
```
astsiapanay, line 357: So you want to remove the upstream from the list of available upstreams? Shouldn't we try to make one more try instead?

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-422.nightly-test.deltixhub.io
>E2E tests status: failed

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@Slf4j
+public class UpstreamRouteProvider {
```
astsiapanay, line 27: Please javadoc. It's worth mentioning that if upstream is changed due to config change then the balancer state is reset.

src/main/java/com/epam/aidial/core/upstream/LoadBalancerProvider.java
astsiapanay: As agreed with the team we decided not to keep the balancers' state across Core instances.

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
Maxim-Gadalov: because we already call next in constructor

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: if upstream failed with non-retryable status code we report used upstream anyway

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
Maxim-Gadalov: no, this is basically max retry attempts, we don't want to stick this number to upstream count, because applications usually have only 1 upstream

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: see javadoc description

astsiapanay: Perhaps it’s worth adding one more method in this case.
Since that confuses

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
astsiapanay: The previous concept Iterator was more clear.
You could increase the var used in the constructor as well?

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
astsiapanay: Please put the value to constant with description

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: fail5xx fail429? I think this may lead to more complex error handling
We can add a comment in all places where we pass -1

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
Maxim-Gadalov: previous iterator and balancer was simpler, we can initialize with 1 in constructor

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: not really, you don't have response in all cases

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
Maxim-Gadalov: getOriginalUpstreams may be subset

Maxim-Gadalov:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-422.nightly-test.deltixhub.io
>E2E tests status: failed

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: I see all cases use http proxy response

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
astsiapanay: I'm for clear methods where arguments make sense. You pass -1 just to pass something. That's not good. You could make an arg as optional like in Python but Java doesn't support that. I would overload the methods that's more simple to me.

```
public synchronized void failed(HttpStatus status) {
        if (status.is5xx()) {
            if (++errorCount >= errorsThreshold) {
                retryAfter = System.currentTimeMillis() + (long) (INITIAL_BACKOFF_DELAY_MS * Math.pow(BACKOFF_BASE, errorCount));
            }
        }
    }

    public synchronized void failed(HttpClientResponse response) {
        HttpStatus status = HttpStatus.fromStatusCode(response.statusCode());
        if (status == HttpStatus.TOO_MANY_REQUESTS) {
            long retryAfterSeconds = UpstreamRoute.calculateRetryAfterSeconds(response);
            retryAfter = System.currentTimeMillis() + retryAfterSeconds * 1000;
        } else {
            failed(status);
        }
    }
```

Makes sense?

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
astsiapanay: You create a new instance of HashSet. I suggest reusing
```
(new HashSet<>(upstreamProvider.getUpstreams())).containsAll(balancer.getOriginalUpstreams())
```

Eventually the code of HashSet `equals` does:
```
Collection<?> c = (Collection<?>) o;
        if (c.size() != size())
            return false;
        try {
            return containsAll(c);
        } catch (ClassCastException | NullPointerException unused) {
            return false;
        }
```

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,76 +1,130 @@
...
-        return next;
+    @Nullable
+    private UpstreamState upstreamState;
+    private int used;
+
+    public UpstreamRoute(LoadBalancer<UpstreamState> balancer, int maxUpstreamsToUse) {
+        this.balancer = balancer;
+        this.maxUpstreamsToUse = maxUpstreamsToUse;
+        this.upstreamState = balancer.next();
+        this.used = 1;
```
astsiapanay: `this.used = upstreamState == null ? 0 : 1;` ?

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
Maxim-Gadalov: This is not a hot path of the code, so creating some garbage here - not so critical
You suggest to rewrite hashset equals with reusing 1 party, which means:
instead of simple calling
`new HashSet<>(upstreamProvider.getUpstreams()).equals(new HashSet<>(balancer.getOriginalUpstreams()));`
we should call
```
if (upstreamProvider.getUpstreams().size() != balancer.getOriginalUpstreams().size()) {
return false;
}
return upstreamProvider.getUpstreams().containsAll(new HashSet<>(balancer.getOriginalUpstreams()));
```

or merge those 2 into a single condition which is more complicated to read

1 prefer to have a simple single line condition here

src/main/java/com/epam/aidial/core/AiDial.java
```diff
@@ -95,7 +95,7 @@ void start() throws Exception {
             client = vertx.createHttpClient(new HttpClientOptions(settings("client")));
 
             LogStore logStore = new GfLogStore(vertx);
-            UpstreamBalancer upstreamBalancer = new UpstreamBalancer();
+            UpstreamRouteProvider upstreamBalancer = new UpstreamRouteProvider();
```
artsiomkorzun: change variable accordingly

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
artsiomkorzun: agree

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,76 +1,130 @@
...
+        // if max attempts reached - do not call balancer
+        if (used + 1 > maxUpstreamsToUse) {
+            this.upstreamState = null;
+            return null;
         }
-
-        return null;
+        used++;
+        UpstreamState upstreamState = balancer.next();
+        this.upstreamState = upstreamState;
```
artsiomkorzun: you can simpify `upstreamState = balancer.next()`

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,76 +1,130 @@
...
+     *
+     * @param status - response http status; typically, 5xx or 429
+     * @param retryAfterSeconds - the amount of seconds after which upstream should be available; if status 5xx this value ignored
      */
-    public void retry() {
-        if (prev < next && retries <= maxRetries) {
-            retries++;
-            next = prev;
+    public void fail(HttpStatus status, long retryAfterSeconds) {
+        if (upstreamState == null) {
```
artsiomkorzun: `if (upsteamState != null)` is much better flow with only one jump

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
```diff
@@ -1,76 +1,130 @@
...
+            return;
         }
+        upstreamState.failed(status, retryAfterSeconds);
+    }
+
+    public void succeed() {
+        if (upstreamState == null) {
```
artsiomkorzun: same

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
```diff
@@ -0,0 +1,122 @@
...
+            String name = deployment.getName();
+
+            DeploymentUpstreamProvider endpointProvider = new DeploymentUpstreamProvider(deployment);
+            TieredBalancer previous = oldState.get(name);
+            updateDeployment(endpointProvider, previous, uniqueDeployments, newState);
+        }
+    }
+
+    private static void updateDeployment(UpstreamProvider upstream, TieredBalancer previous, Set<String> uniqueDeployments,
+                                         Map<String, TieredBalancer> newState) {
```
artsiomkorzun: Why you can't just use Map<String, TieredBalancer> newState to check for duplicates?

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -0,0 +1,69 @@
...
+import lombok.Getter;
+
+public class UpstreamState implements Comparable<UpstreamState> {
+
+    @Getter
+    private final Upstream upstream;
+    private final int errorsThreshold;
+
+    private static final long INITIAL_BACKOFF_DELAY_MS = 1000;
+    private static final int BACKOFF_BASE = 2;
```
artsiomkorzun: Let's add MAX_BACKOFF_DELAY to limit the max delay.

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Register upstream failure. Supported error codes are 429 and 5xx.
+     *
+     * @param status response status code from upstream
+     * @param retryAfterSeconds time in seconds when upstream may become available; only take into account with 429 status code
+     */
+    public synchronized void failed(HttpStatus status, long retryAfterSeconds) {
+        if (status == HttpStatus.TOO_MANY_REQUESTS) {
+            retryAfter = System.currentTimeMillis() + retryAfterSeconds * 1000;
```
artsiomkorzun: Is retryAfterSeconds guaranteed to be always non-negative?

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (status.is5xx()) {
+            if (++errorCount >= errorsThreshold) {
```
artsiomkorzun, line 48: I hope we won't hit the overflow.

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        }
+        try {
+            int size = upstreams.size();
+            for (int i = 0; i < size; i++) {
+                UpstreamState upstreamState = upstreams.get(i);
+                double actualUsageRate = upstreamsUsage[i] == 0 ? 0 : (double) upstreamsUsage[i] / totalUsage;
+                double expectedUsageRate = (double) upstreamsWeights[i] / totalWeight;
+                double delta = expectedUsageRate - actualUsageRate;
+                upstreamPriority.offer(new UpstreamUsage(upstreamState, i, delta));
```
artsiomkorzun, line 62: You add N but remove only 1.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: what about `handleProxyResponseError`

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
Maxim-Gadalov: yes, check Retry-After header specification

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
Maxim-Gadalov: almost impossible

src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java
Maxim-Gadalov: we build full queue, and get closest available upstream

src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java
astsiapanay: ok. let's do it.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
Maxim-Gadalov: my main point - call single method in case of any error. Method itself responsible for the logic how to dial with specific errors.
with your approach you push every developers (who will use this class) to handle this logic by themselves and check every time what's the type of error

We can introduce 2 methods, but they must follow single logic, for example
```
public synchronized void failed(HttpStatus status, long retryAfterSeconds) {
        if (status == HttpStatus.TOO_MANY_REQUESTS) {
            retryAfter = System.currentTimeMillis() + retryAfterSeconds * 1000;
        }
                if (status.is5xx()) {
            if (++errorCount >= errorsThreshold) {
                retryAfter = System.currentTimeMillis() + (long) (INITIAL_BACKOFF_DELAY_MS * Math.pow(BACKOFF_BASE, errorCount));
            }
        }
    }

public synchronized void failed(HttpStatus status) {
failed(status, 30);
    }
    ```

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java (hunk 34 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java]: added

File: src/main/java/com/epam/aidial/core/AiDial.java (hunk 35 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/AiDial.java]: fixed

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java (hunk 36 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java]: fixed

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java (hunk 37 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java]: fixed

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java (hunk 38 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java]: fixed

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java (hunk 39 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java]: fixed

Comment (artsiomkorzun) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamRouteProvider.java]: I see Long.parseLong() which can return negatives.

File: src/main/java/com/epam/aidial/core/upstream/UpstreamState.java (hunk 41 above)
Comment (artsiomkorzun) [reply src/main/java/com/epam/aidial/core/upstream/UpstreamState.java]: Long.parseLong could return negatives. Please make it more safe without context of Retry-After. Use Math.max(0, retrySeconds) or similar code.

File: src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java (hunk 43 above)
Comment (artsiomkorzun) [reply src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java:62]: You add N on each next call. Did you forget to clear the queue before filling it out?

File: src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java (hunk 25 above)
Comment (astsiapanay) [reply src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java]: Does it look more simple?

```
public synchronized void failed(HttpStatus status) {
        if (status.is5xx()) {
            if (++errorCount >= errorsThreshold) {
                retryAfter = System.currentTimeMillis() + (long) (INITIAL_BACKOFF_DELAY_MS * Math.pow(BACKOFF_BASE, errorCount));
            }
        }
    }

    public synchronized void failed(HttpClientResponse response) {
        HttpStatus status = HttpStatus.fromStatusCode(response.statusCode());
        if (status == HttpStatus.TOO_MANY_REQUESTS) {
            long retryAfterSeconds = UpstreamRoute.calculateRetryAfterSeconds(response);
            retryAfter = System.currentTimeMillis() + retryAfterSeconds * 1000;
        } else {
            failed(status);
        }
    }
```

File: src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java (hunk 43 above)
Comment (astsiapanay) [reply src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java:62]: I think it's worth adding a comment why we build the queue here instead of find an element with the max delta.
The reason is that we want to keep good load distribution in case of some upstreams might be unavailable. We increase usage for candidates and keep searching if the current one is not available.

Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java:62]: @artsiomkorzun 
        } finally {
            upstreamPriority.clear();
        }
        I'll add a comment

Comment (artsiomkorzun) [reply src/main/java/com/epam/aidial/core/upstream/WeightedRoundRobinBalancer.java:62]: Oh, missed this piece of code

File: src/main/java/com/epam/aidial/core/upstream/TieredBalancer.java (hunk 13 above)
Comment (Maxim-Gadalov) [reply src/main/java/com/epam/aidial/core/upstream/TieredBalancer.java]: changed

**Review by astsiapanay — DISMISSED** [review]

**Comment by Maxim-Gadalov** [comment]
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-422.nightly-test.deltixhub.io
>E2E tests status: success

**Review by astsiapanay — DISMISSED** [review]

File: src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java [hunk 44]
```diff
@@ -126,6 +126,7 @@ private static long calculateRetryAfterSeconds(HttpClientResponse response) {
             if (retryAfterHeaderValue != null) {
                 return Long.parseLong(retryAfterHeaderValue);
             }
+            log.info("Retry-after header not found, available headers: {}", response.headers());
```
astsiapanay: Please don’t print all headers. There might be sensitive info like api key or jet

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
```diff
@@ -37,6 +41,8 @@ public UpstreamState(Upstream upstream, int errorsThreshold) {
     public synchronized void failed(HttpStatus status, long retryAfterSeconds) {
         if (status == HttpStatus.TOO_MANY_REQUESTS) {
             retryAfter = System.currentTimeMillis() + Math.max(retryAfterSeconds, 0) * 1000;
+
+            log.info("Upstream limit hit: retry after {} millis {}", retryAfter, Instant.ofEpochMilli(retryAfter).toString());
```
astsiapanay:
```
log.info("Upstream limit hit: retry after {}", Instant.ofEpochMilli(retryAfter).toString());
```
?

astsiapanay: please change to warn level

astsiapanay (changes requested):

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/upstream/UpstreamRoute.java
Maxim-Gadalov: fixed

src/main/java/com/epam/aidial/core/upstream/UpstreamState.java
Maxim-Gadalov: chnaged

astsiapanay (dismissed):

Maxim-Gadalov:
/deploy-review

>Environment URL: https://chat-ai-dial-core-pr-422.nightly-test.deltixhub.io
>E2E tests status: failed

astsiapanay (approved):

PR #456, opened 2024-08-23
Convert openapi request to dial format

AliakseiBuhakou:
New feature: conversion of the openapi request to dial old format to support both request types


ai-dial-actions:
Hey there and thank you for opening this pull request! 👋🏼

We require pull request titles to follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/) and it looks like your proposed title needs to be adjusted.

Details:

```
No release type found in pull request title "Convert openapi request to dial format". Add a prefix to indicate what kind of release this pull request corresponds to. For reference, see https://www.conventionalcommits.org/

Available types:
 - fix
 - hotfix
 - feat
 - feature
 - chore
```

astsiapanay:
Hi @AliakseiBuhakou,

Could you please link the PR to an issue to be reported in DIAL Core project?

If the issue doesn't exist please create it.

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
```diff
@@ -271,24 +272,25 @@ void handleRequestBody(Buffer requestBody) {
                 context.getTraceId(), context.getSpanId(),
                 context.getProject(), deployment.getName(), requestBody.length());
 
-        context.setRequestBody(requestBody);
-        context.setRequestBodyTimestamp(System.currentTimeMillis());
+        try {
+            ObjectNode json = (ObjectNode) ProxyUtil.MAPPER.readTree(requestBody.getBytes());
+            ObjectNode dialFormat = RequestUtil.convertToDialFormat(json);
```
astsiapanay, line 277: It seems like the conversion function is a good candidate to put it to the function's chain.
Please see `ProxyUtil.processChain(tree, enhancementFunctions)` for more details.

src/main/java/com/epam/aidial/core/util/RequestUtil.java
```diff
@@ -0,0 +1,82 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import lombok.experimental.UtilityClass;
+
+
+@UtilityClass
+public class RequestUtil {
```
astsiapanay, line 11: Please a javadoc explaining the purpose of the class

astsiapanay (changes requested):

astsiapanay:
@sdryapko 
Should the feature be planned and included into the current milestone?
I don't remember we discussed that on meetings.

AliakseiBuhakou:
> Hi @AliakseiBuhakou,
> 
> Could you please link the PR to an issue to be reported in DIAL Core project?
> 
> If the issue doesn't exist please create it.

Unfortunately, I don't have access to it.

src/main/java/com/epam/aidial/core/util/RequestUtil.java
AliakseiBuhakou: Sure, I will

src/main/java/com/epam/aidial/core/controller/DeploymentPostController.java
AliakseiBuhakou: The purpose of this feature is to handle new format of the request. We agreed that the best way to do that is to convert new format to the old one. Such conversion should be done at the real beginning of request handling to not modify any processing downstream. 

I think making this as enhancement function in theory may lead to some processing errors in future if something is changed. For example the order of the functions execution or introducing a new function that is called before request transformation.

I can wrap my code as an enhancement function but I don't think It is safe. Any thoughts?

astsiapanay: Right. It makes sense to run the conversion as the first function anyway. However that might be an unsafe way to move the logic to the functions' chain.

astsiapanay:
> > Hi @AliakseiBuhakou,
> > Could you please link the PR to an issue to be reported in DIAL Core project?
> > If the issue doesn't exist please create it.
> 
> Unfortunately, I don't have access to it.

@sdryapko 
Please give access @AliakseiBuhakou to report an issue to DIAL Core project.

astsiapanay:
According to @adubovik the feature is already supported by adapters.
the issue can be closed

Issue #465, opened 2024-09-02
Add resource type to user metadata of a blob


PMitrafanau:
## Problem statement

Currently, it's impossible to find all resources of a specific type in storage without using wildcard search.
That can limit functionality on some back-end storages. For example, AWS S3 doesn't support a retention policy
with a wildcard in the file prefix condition.

Some of the platform users want to know the `ResourceType` of a particular object stored in back-end storage.
Storing that data with object metadata can help properly identify the `ResourceType` of the stored object.

PR #466, opened 2024-09-03
feat: Add resource type to user metadata of a blob #465

PMitrafanau:
Issue: #465

### DIFF
--- a/src/main/java/com/epam/aidial/core/service/ResourceService.java
+++ b/src/main/java/com/epam/aidial/core/service/ResourceService.java
@@ -52,14 +52,15 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import static com.epam.aidial.core.util.ResourceUtil.CREATED_AT_ATTRIBUTE;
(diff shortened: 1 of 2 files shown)


src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -762,6 +774,9 @@ private static Map<String, String> toUserMetadata(String etag, Long createdAt, L
         if (updatedAt != null) {
             metadata.put(ResourceUtil.UPDATED_AT_ATTRIBUTE, Long.toString(updatedAt));
         }
+        if (resourceType != null) {
```
akurnosau, line 777: Question: is it possible that Resource will not have ResourceType? I thought it is used everywhere at least to build a path.

akurnosau: Discussed offline. That check makes sense.

akurnosau (approved):

src/main/java/com/epam/aidial/core/service/ResourceService.java
PMitrafanau: For the context: I've traced all the scenarios I could and indeed the `resourceType` seems to be non-nullable here. And so is `updatedAt`. But as there is still a check for the `updatedAt` it means I could miss something or it's there just-in-case. With this in mind, I've decided to put the same check for the `resourceType`.

sdryapko:
@astsiapanay  please review

astsiapanay (approved):

PR #474, opened 2024-09-06
feat: move resources across buckets

artsiomkorzun:


src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
```diff
@@ -76,29 +76,28 @@ public Future<?> move() {
...
-                    if (!destinationResource.getBucketName().equals(bucket)) {
-                        throw new IllegalArgumentException("destinationUrl do not belong to the user");
+                    if (source.getUrl().equals(destination.getUrl())) {
+                        throw new IllegalArgumentException("source and destination resources cannot be the same");
                     }
 
-                    if (!sourceResource.getType().equals(destinationResource.getType())) {
-                        throw new IllegalArgumentException("source and destination resources must be the same type");
-                    }
+                    Set<ResourceDescription> resources = Set.of(source, destination);
```
astsiapanay, line 91: The permission check is not quite correct. A requester should have read access to source and write access to destination.

src/main/java/com/epam/aidial/core/service/InvitationService.java
```diff
@@ -170,22 +170,35 @@ public void cleanUpPermissions(
         });
     }
 
-    public void moveResource(String bucket, String location, ResourceDescription source, ResourceDescription destination) {
-        ResourceDescription resource = ResourceDescription.fromDecoded(ResourceType.INVITATION, bucket, location, INVITATION_RESOURCE_FILENAME);
-        resourceService.computeResource(resource, state -> {
+    public void moveResource(ResourceDescription source, ResourceDescription destination) {
+        if (!source.isPrivate()) {
```
astsiapanay: Should we add a warning message?

src/main/java/com/epam/aidial/core/service/ShareService.java
```diff
@@ -379,11 +386,16 @@ public void copySharedAccess(String bucket, String location, ResourceDescription
...
     }
 
-    public void moveSharedAccess(String bucket, String location, ResourceDescription source, ResourceDescription destination) {
-        // copy shared access from source to destination
-        copySharedAccess(bucket, location, source, destination);
-        // revoke shared access from source
-        revokeSharedAccess(bucket, location, Map.of(source, ResourceAccessType.ALL));
+    public void moveSharedAccess(ResourceDescription source, ResourceDescription destination) {
+        if (!source.isPrivate()) {
+            return; // only private resources are sharable
```
astsiapanay: the same

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
artsiomkorzun: Move removes the source resource (not just reads)

src/main/java/com/epam/aidial/core/service/InvitationService.java
artsiomkorzun: Why? Now it is a normal path when we move resources from public bucket.

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
artsiomkorzun: Do you want me to check that there is read/write access to source and just write access to destination?

src/main/java/com/epam/aidial/core/service/InvitationService.java
artsiomkorzun: Reworked to make it more clear.

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
astsiapanay: Yes, you're right. Let's check permissions in this way.

src/main/java/com/epam/aidial/core/service/ResourceOperationService.java
```diff
@@ -51,11 +53,23 @@ public void moveResource(String bucket, String location, ResourceDescription sou
...
+            String bucketName = source.getBucketName();
+            String bucketLocation = source.getBucketLocation();
+            boolean isSameBucket = source.getBucketName().equals(destination.getBucketName());
+
+            if (isSameBucket) {
+                invitationService.moveResource(bucketName, bucketLocation, source, destination);
+                shareService.moveSharedAccess(bucketName, bucketLocation, source, destination);
+            } else {
+                Map<ResourceDescription, Set<ResourceAccessType>> resources =
+                        Map.of(source, Set.of(ResourceAccessType.READ, ResourceAccessType.WRITE));
```
Oleksii-Klimov:
```java
ResourceType.ALL
```

src/main/java/com/epam/aidial/core/controller/ResourceOperationController.java
artsiomkorzun: changed

src/main/java/com/epam/aidial/core/service/ResourceOperationService.java
artsiomkorzun: changed

astsiapanay (approved):

PR #498, opened 2024-09-27
feat: initial application refactoring

artsiomkorzun:
Publication code is not changed in this pr


src/main/java/com/epam/aidial/core/config/Application.java
```diff
@@ -10,4 +10,41 @@
...
 public class Application extends Deployment {
+
+    private Function function;
+
+    @Data
+    @Accessors(chain = true)
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public static class Function {
+
+        private String sources;
```
astsiapanay: why is the field name plural?

src/main/java/com/epam/aidial/core/controller/ApplicationController.java
```diff
@@ -22,98 +22,67 @@ public class ApplicationController {
...
-    private final CustomApplicationService customApplicationService;
-    private final boolean includeCustomApplications;
+    private final ApplicationService applications;
 
     public ApplicationController(ProxyContext context, Proxy proxy) {
         this.context = context;
         this.vertx = proxy.getVertx();
-        this.customApplicationService = proxy.getCustomApplicationService();
-        this.includeCustomApplications = customApplicationService.includeCustomApplications();
+        this.applications = proxy.getApplicationService();
```
astsiapanay: I wouldn't recommend to have the field in a plural form. It may confuse a reviewer.
Let's add a suffix `Service`.
`applications` are usually treated as a set or list of applications.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -117,30 +122,39 @@ private Future<?> getResource(ResourceDescription descriptor, boolean hasWriteAc
...
-                            Application application = ProxyUtil.convertToObject(body, Application.class, true);
-                            ApplicationData applicationData = ApplicationUtil.mapApplication(application);
-                            context.respond(HttpStatus.OK, ProxyUtil.convertToString(applicationData));
-                        } else {
-                            context.respond(HttpStatus.OK, body);
-                        }
-                    }
+        Future<Pair<ResourceItemMetadata, String>> responseFuture;
+
+        if (descriptor.getType() == ResourceType.APPLICATION) {
```
astsiapanay: It's better to move the block to a single method.

src/main/java/com/epam/aidial/core/controller/ResourceController.java
```diff
@@ -117,30 +122,39 @@ private Future<?> getResource(ResourceDescription descriptor, boolean hasWriteAc
...
+                Application application = result.getValue();
+                String body = hasWriteAccess
+                        ? ProxyUtil.convertToString(application, true)
+                        : ProxyUtil.convertToString(ApplicationUtil.mapApplication(application));
+
+                return Pair.of(meta, body);
+
+            }, false);
+        } else {
+            responseFuture = vertx.executeBlocking(() -> {
```
astsiapanay: the same thing

src/main/java/com/epam/aidial/core/service/ResourceService.java
```diff
@@ -442,24 +442,47 @@ public FileMetadata finishFileUpload(
     }
 
     public void computeResource(ResourceDescription descriptor, Function<String, String> fn) {
+        computeResource(descriptor, EtagHeader.ANY, fn);
+    }
+
+    public ResourceItemMetadata computeResource(ResourceDescription descriptor, EtagHeader etag, Function<String, String> fn) {
```
astsiapanay, line 448: I would add a Unit test. The logic has more branches need to be verified.

src/main/java/com/epam/aidial/core/config/Application.java
artsiomkorzun: renamed to sourceFolder

src/main/java/com/epam/aidial/core/controller/ApplicationController.java
artsiomkorzun: renamed

src/main/java/com/epam/aidial/core/controller/ResourceController.java
artsiomkorzun: Please collaborate. I prefer not having complex inner ifs.

src/main/java/com/epam/aidial/core/service/ResourceService.java
artsiomkorzun: All branches are already covered.

astsiapanay: Where are the branches covered? Could you please point out to a place?

src/main/java/com/epam/aidial/core/controller/ResourceController.java
astsiapanay:
```
private Future<Pair<ResourceItemMetadata, String>> getApplication(ResourceDescription descriptor, boolean hasWriteAccess) {
            Pair<ResourceItemMetadata, Application> result = applicationService.getApplication(descriptor);
            ResourceItemMetadata meta = result.getKey();

            Application application = result.getValue();
            String body = hasWriteAccess
                    ? ProxyUtil.convertToString(application, true)
                    : ProxyUtil.convertToString(ApplicationUtil.mapApplication(application));

            return Pair.of(meta, body);

        }, false);
    }
```

makes sense?

astsiapanay:
@artsiomkorzun 
I think we should eliminate cycle dependencies  like this `ApplicationService` <-> `ResourceService`.

src/main/java/com/epam/aidial/core/service/ApplicationService.java
```diff
@@ -0,0 +1,282 @@
...
+                            || function.getStatus() == Application.Function.Status.STARTED
+                            || function.getStatus() == Application.Function.Status.STOPPING)) {
+                throw new HttpException(HttpStatus.NOT_FOUND, "Application must be stopped: " + resource.getUrl());
+            }
+
+            return null;
+        });
+    }
+
+    public void copyApplication(ResourceDescription source, ResourceDescription description, boolean overwrite) {
```
astsiapanay: Let's rename the 2nd arg to `destination`

artsiomkorzun: renamed

astsiapanay (approved):

PR #507, opened 2024-10-04
feat: supported auto-sharing for files in content parts

### LINKED ISSUE #506: Support auto-sharing for files defined in message content parts
The content of messages could have references to image URLs at the path: 
`request.messages[*].content[*].image_url.url`. 

We need to collect and auto-share those like it's already done for 
`request.messages[*].custom_content.attachments[*].url`.

Required for 
* https://github.com/epam/ai-dial-adapter-vertexai/issues/129
* https://github.com/epam/ai-dial-adapter-bedrock/issues/156
* https://github.com/epam/ai-dial-adapter-openai/issues/155

adubovik:
Resolves https://github.com/epam/ai-dial-core/issues/506

### DIFF
--- a/src/main/java/com/epam/aidial/core/util/ProxyUtil.java
+++ b/src/main/java/com/epam/aidial/core/util/ProxyUtil.java
@@ -161,6 +161,54 @@ public class ProxyUtil {
         }
     }
 
+    private static void collectAttachedFilesFromContent(JsonNode content, Consumer<String> consumer) {
+        if (content == null || !(content instanceof ArrayNode)) {
+            return;
+        }
+        ArrayNode contentParts = (ArrayNode) content;
+
+        for (int i = 0; i < contentParts.size(); i++) {
+            JsonNode partNode = contentParts.get(i);
+            JsonNode partTypeNode = partNode.get("type");
+            if (partTypeNode == null || !partTypeNode.textValue().equals("image_url")) {
+                continue;
+            }
+            JsonNode imageNode = partNode.get("image_url");
+            if (imageNode == null) {
+                continue;
+            }
+            JsonNode urlNode = imageNode.get("url");
+            collectURL(urlNode, null, consumer);
+        }
+    }
+
+    private static void collectAttachedFilesFromCustomContent(JsonNode customContent, Consumer<String> consumer) {
+        if (customContent == null) {
+            return;
+        }
+        ArrayNode attachments = (ArrayNode) customContent.get("attachments");
+        if (attachments != null) {
+            for (int i = 0; i < attachments.size(); i++) {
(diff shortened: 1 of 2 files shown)


adubovik:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-507.nightly-test.deltixhub.io
>E2E tests status: success

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -161,50 +161,80 @@ private static void collectAttachedFilesCustomInput(JsonNode input, Consumer<Str
         }
     }
 
-    private static void collectAttachedFilesChatCompletion(ObjectNode tree, Consumer<String> consumer) {
-        ArrayNode messages = (ArrayNode) tree.get("messages");
-        if (messages == null) {
+    private static void collectAttachedFilesFromContent(JsonNode content, Consumer<String> consumer) {
+        if (content == null || !(content instanceof ArrayNode)) {
```
astsiapanay: remove redundant check `content == null`

astsiapanay: it could be replaced with

```
if (!(content instanceof ArrayNode contentParts)) {
            return;
        }
```

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
```diff
@@ -161,50 +161,80 @@ private static void collectAttachedFilesCustomInput(JsonNode input, Consumer<Str
...
+
+    private static void collectAttachedFilesFromUrl(JsonNode urlNode, JsonNode typeNode, Consumer<String> consumer) {
         if (urlNode == null) {
             return;
         }
 
         String url = urlNode.textValue();
 
-        JsonNode typeNode = attachment.get("type");
+        if (url == null) {
```
astsiapanay, line 233: the check could be unnecessary.

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
adubovik: fixed

src/main/java/com/epam/aidial/core/util/ProxyUtil.java
adubovik: `JsonNode` is null when the field is **missing** from the dictionary.
When the field is **present** and equal to JSON `null`, this `JsonNode` isn't null.

So the check is required for this case:

```json
{"type": "image_url", "image_url": {"url": null}},
```

Otherwise, `consumer.accept(null);` will be called.

astsiapanay (approved):

PR #539, opened 2024-10-22
chore: re-factor ResourceService

astsiapanay:


README.md
```diff
@@ -79,6 +79,7 @@ Priority order:
 | resources.syncBatch                           |        4096        |    No    |How many resources to sync in one go.
 | resources.cacheExpiration                     |       300000       |    No    |Expiration in milliseconds for synced resources in Redis.
 | resources.compressionMinSize                  |        256         |    No    |Compress a resource with gzip if its size in bytes more or equal to this value.
+| resources.redisPrefix                         |      resource      |    No    |Prefix name is used to form Redis key for resource's queue and topic.
```
artsiomkorzun, line 82: Default should be used from storage, otherwise it is breaking change

server/src/main/java/com/epam/aidial/core/server/service/ScheduledService.java
```diff
@@ -0,0 +1,23 @@
+package com.epam.aidial.core.server.service;
+
+/**
+ * A service that can schedule commands to run after a given delay, or to execute periodically.
+ * The schedule methods create tasks with various delays and return a task object that can be used to cancel or check execution.
+ */
+public interface ScheduledService {
```
artsiomkorzun, line 7: TimerService is better?

README.md
astsiapanay: it's not breaking change. the default value is taken from settings.json

artsiomkorzun: Blob prefix should be the default in this case. People already set some values to blob prefix. If they run the old config it won't work.

astsiapanay: discussed privately. Let blob storage prefix is used to make a different names for topic and queue in Redis

server/src/main/java/com/epam/aidial/core/server/service/ScheduledService.java
astsiapanay: Agree

astsiapanay:
closed as invalid

PR #540, opened 2024-10-22
feat: Ability to split roles by Delimiter

akurnosau:
That is to allow split roles if they are presented as a single String in JWT token.
If `rolesDelimiter` is set to `" "`, then following claim
```
...
roles: "r1 r2 r3",
...
```
will be extracted as `["r1","r2","r3"]`

- Doesn't affect existing behaviour if `rolesDelimiter` is not set.
- Doesn't affect existing behaviour if roles are presented as json array


server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -153,6 +157,9 @@ private List<String> extractUserRoles(Map<String, Object> map) {
                 if (next instanceof List) {
                     return (List<String>) next;
                 } else if (next instanceof String) {
+                    if (rolesDelimiter != null) {
+                        return List.of(((String) next).split(rolesDelimiter));
```
alekseyvdovenko: Maybe it makes sense to trim split role values. For example, when the delimiter is "," and the roles string contains a whitespace: "r1, r2". The result will be "r1", " r2". The second role value contains a leading whitespace.

alekseyvdovenko: btw, if the roles string is "r1  r2  r3" (two whitespaces) and the delimiter is " " (single whitespace), then the roles list will contain 5 values (2 of them are empty strings)

alekseyvdovenko:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-540.nightly-test.deltixhub.io
>E2E tests status: success

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
akurnosau: First example: your delimiter should be ", " then.
Second: that is true. However that shouldn't be the case with JWT tokens, I barely can't imagine IDP that issues roles with unpredictable number of delimiters between them. 
Anyway, let's maybe filter empty strings from the list of extracted roles. 
Thank you for that case.

akurnosau:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-540.nightly-test.deltixhub.io
>E2E tests status: failed

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -40,6 +41,9 @@ public class IdentityProvider {
     // path to the claim of user roles in JWT
     private final String[] rolePath;
 
+    // Delimiter to split the roles is they are set as a single String
```
astsiapanay: typo `is` -> `if`

alekseyvdovenko (dismissed):

astsiapanay (dismissed):

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
akurnosau: Nice catch! Fixed

alekseyvdovenko (approved):

PR #541, opened 2024-10-22
chore: re-factor ResourceService

astsiapanay:
- make resource type extensible via introducing interface
- spit ResourceDescription into two classes


server/src/main/java/com/epam/aidial/core/server/resource/ResourceDescriptor.java
```diff
@@ -0,0 +1,158 @@
...
+    public String getParentPath() {
+        return parentFolders.isEmpty() ? null : String.join(PATH_SEPARATOR, parentFolders);
+    }
+
+    public ResourceDescriptor resolveByUrl(String url) {
+        String prefix = type.getGroup() + PATH_SEPARATOR + bucketName  + PATH_SEPARATOR;
+        return resolve(prefix, url);
+    }
+
+    public ResourceDescriptor resolveByPath(String url) {
```
artsiomkorzun: rename parameter

server/src/main/java/com/epam/aidial/core/server/resource/ResourceDescriptor.java
```diff
@@ -0,0 +1,158 @@
...
+        if (!isFolder) {
+            throw new IllegalStateException("Resource must be folder");
+        }
+        if (!url.startsWith(prefix)) {
+            throw new IllegalArgumentException("Incompatible description and absolute path");
+        }
+
+        String relativePath = url.substring(prefix.length());
+
+        List<String> paths = Stream.of(relativePath.split(PATH_SEPARATOR)).map(UrlUtil::decodePath).toList();
```
artsiomkorzun: Why do you decode each segment when called for a path?

astsiapanay: We want to put a decoded path to blob store

server/src/main/java/com/epam/aidial/core/server/resource/ResourceDescriptor.java
astsiapanay: I changed the method name

server/src/main/java/com/epam/aidial/core/server/controller/FileMetadataController.java
```diff
@@ -55,7 +55,7 @@ protected Future<?> handle(ResourceDescriptor resource, boolean hasWriteAccess)
             } catch (Exception ex) {
                 log.error("Failed to list files", ex);
                 context.respond(HttpStatus.INTERNAL_SERVER_ERROR,
-                        "Failed to list files by path %s/%s".formatted(resource.getBucketName(), resource.getOriginalPath()));
+                        "Failed to list files by path %s".formatted(resource.getUrl()));
```
artsiomkorzun, line 58: Why to change?

astsiapanay: I removed originalPath since it's used just here and very similar to what has been changed

artsiomkorzun: Sorry, misread what was changed. This is good change.

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-541.nightly-test.deltixhub.io
>E2E tests status: success

PR #543, opened 2024-10-23
fix: span names 

### LINKED ISSUE #542: Fix span names
Fix span names that cannot be updated when jwt auth happens and jwk result is cached

alekseyvdovenko:
Closes https://github.com/epam/ai-dial-core/issues/542

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
+++ b/server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
@@ -182,7 +182,13 @@ public class IdentityProvider {
     }
 
     private Future<JwkResult> getJwk(String kid) {
-        return cache.computeIfAbsent(kid, key -> vertx.executeBlocking(() -> {
+        /* The result of vertx.executeBlocking is a future that contains Vert.x context which is valid during a request
+         * execution. So, if we put that future in a cache, it will contain a context from the initial request, that
+         * may be invalid for further requests. For this reason, when we retrieve the future from the cache, we must
+         * extract the value and put it into another future (Promise) which holds a valid context of a current request.
+         * */
+        Promise<JwkResult> promise = Promise.promise();
+        cache.computeIfAbsent(kid, key -> vertx.executeBlocking(() -> {
             JwkResult jwkResult;
             long currentTime = System.currentTimeMillis();
             try {
@@ -192,7 +198,8 @@ public class IdentityProvider {
                 jwkResult = new JwkResult(null, e, currentTime + negativeCacheExpirationMs);
             }
             return jwkResult;
-        }, false));
+        }, false)).onSuccess(promise::complete).onFailure(promise::fail);
+        return promise.future();
     }
 
     private Future<DecodedJWT> verifyJwt(DecodedJWT jwt) {


alekseyvdovenko:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-543.nightly-test.deltixhub.io
>E2E tests status: failed

alekseyvdovenko:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-543.nightly-test.deltixhub.io
>E2E tests status: failed

akurnosau (dismissed):

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -182,7 +182,8 @@ public static DecodedJWT decodeJwtToken(String encodedToken) {
     }
 
     private Future<JwkResult> getJwk(String kid) {
-        return cache.computeIfAbsent(kid, key -> vertx.executeBlocking(() -> {
+        Promise<JwkResult> promise = Promise.promise();
+        cache.computeIfAbsent(kid, key -> vertx.executeBlocking(() -> {
```
astsiapanay, line 191: Please add a comment why we don’t reuse the cached result.

server/src/main/java/com/epam/aidial/core/server/Proxy.java
```diff
@@ -280,6 +281,11 @@ private void respond(HttpServerRequest request, HttpStatus status) {
     }
 
     private void respond(HttpServerRequest request, HttpStatus status, String body) {
+        if (status == HttpStatus.OK) {
+            SpanUtil.updateName(request.method().name(), request.path());
+        } else if (status.is4xx() || status.is5xx()) {
```
astsiapanay: Why not use if/else?

what if http 3xx comes?

server/src/main/java/com/epam/aidial/core/server/Proxy.java
```diff
@@ -280,6 +281,11 @@ private void respond(HttpServerRequest request, HttpStatus status) {
     }
 
     private void respond(HttpServerRequest request, HttpStatus status, String body) {
+        if (status == HttpStatus.OK) {
+            SpanUtil.updateName(request.method().name(), request.path());
```
astsiapanay: Why do we update span name again? Moreover we set a raw path.

So we do have different span names based on http codes. Why?

server/src/main/java/com/epam/aidial/core/server/util/SpanUtil.java
```diff
@@ -12,7 +12,15 @@ public class SpanUtil {
     private static final List<String> GROUPS = List.of("id", "bucket", "path");
 
     public static void updateName(Pattern pathPattern, String path, String httpMethod) {
-        String spanName = RegexUtil.replaceNamedGroups(pathPattern, path, GROUPS);
-        Span.current().updateName(httpMethod + " " + spanName);
+        String httpPath = RegexUtil.replaceNamedGroups(pathPattern, path, GROUPS);
```
astsiapanay: We can do code re-factoring and move the methods to SelectorController class.

So the class is no longer needed in this case.

alekseyvdovenko: Will be fixed in scope of a separate issue

server/src/main/java/com/epam/aidial/core/server/Proxy.java
alekseyvdovenko: Will be fixed in scope of a separate issue

server/src/main/java/com/epam/aidial/core/server/Proxy.java
alekseyvdovenko: Will be fixed in scope of a separate issue

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
alekseyvdovenko: added

astsiapanay (approved):

PR #548, opened 2024-10-25
chore: make module resource-service

astsiapanay:


resource-service/build.gradle
```diff
@@ -0,0 +1,57 @@
+plugins {
```
artsiomkorzun: It is mostly same as server build.gradle. Could you move the similar parts to build.gradle at the root  project directory?

resource-service/src/main/java/com/epam/aidial/core/resourceservice/service/ResourceTopic.java
```diff
@@ -39,7 +38,7 @@ public Subscription subscribe(Collection<ResourceDescriptor> resources, Consumer
             String url = resource.getUrl();
             urlToSubscriptions.compute(url, (key, subs) -> {
                 if (subs == null) {
-                    subs = new ConcurrentHashSet<>();
+                    subs = ConcurrentHashMap.newKeySet();
```
artsiomkorzun: why?

server/src/main/java/com/epam/aidial/core/server/data/ResourceTypes.java
```diff
@@ -1,6 +1,6 @@
 package com.epam.aidial.core.server.data;
 
-import com.epam.aidial.core.server.resource.ResourceType;
+import com.epam.aidial.core.resourceservice.resource.ResourceType;
```
artsiomkorzun: A weird package name? com.epam.aidial.core.resource.service.ResourceType?

resource-service/src/main/java/com/epam/aidial/core/resourceservice/service/ResourceTopic.java
astsiapanay: we want to remove vertx dep from the module

server/src/main/java/com/epam/aidial/core/server/data/ResourceTypes.java
astsiapanay: right the package name `resourceservice` is a bit weird but
if you suggest `resource.service`
The rest packages will be `resource.service.resource` or `resource.service.service`

resource-service/src/main/java/com/epam/aidial/core/resourceservice/service/ResourceTopic.java
artsiomkorzun: It comes from java.util.*

resource-service/build.gradle
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/data/ResourceTypes.java
astsiapanay: let's rename to `storage`

resource-service/src/main/java/com/epam/aidial/core/resourceservice/service/ResourceTopic.java
artsiomkorzun: Sorry, it comes from Vertx

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-548.nightly-test.deltixhub.io
>E2E tests status: success

PR #550, opened 2024-10-25
feat: store http headers through an interceptors invocation chain

### LINKED ISSUE #531: Support automatic propagation of DIAL-specific headers in the interceptor call chain: X-JOB-TITLE and X-CONVERSATION-ID
Currently, unless the interceptor developer takes care of it, the request headers aren't automatically propagated from interceptor to interceptor to the final endpoint.

These headers are important for the analytics:

https://github.com/epam/ai-dial-core/blob/f171827525909b6b03875ddf67c01191575d5a75/src/main/java/com/epam/aidial/core/Proxy.java#L62-L63

https://github.com/epam/ai-dial-core/blob/f171827525909b6b03875ddf67c01191575d5a75/src/main/java/com/epam/aidial/core/log/GfLogStore.java#L71-L82

The chain of interceptor calls identifies the same conversation and is initiated by the same person, so it makes sense to always propagate these headers.

It would be reasonable to support the propagation in the DIAL Core directly to help the developers avoid mistakes.

alekseyvdovenko:
Closes https://github.com/epam/ai-dial-core/issues/531


server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
```diff
@@ -23,6 +27,10 @@
  */
 @Data
 public class ApiKeyData {
+    private static final List<String> HTTP_HEADERS_TO_STORE = Stream.of(
+            Proxy.HEADER_JOB_TITLE,
+            Proxy.HEADER_CONVERSATION_ID
+    ).map(String::toLowerCase).toList();
```
astsiapanay: I don't think we need to make header names to lower case.

You can see GfLogStore tries to get http headers by upper case

server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
```diff
@@ -67,9 +68,13 @@ private Void doSave(ProxyContext context) {
     private void append(ProxyContext context, LogEntry entry) throws JsonProcessingException {
         HttpServerRequest request = context.getRequest();
         HttpServerResponse response = context.getResponse();
+        ApiKeyData apiKeyData = Optional.ofNullable(context.getProxyApiKeyData())
```
astsiapanay: I would suggest moving the logic for extraction header to ProxyContext.

astsiapanay: ApiKeyData could be retrieved from `context.getApiKeyData` directly if headers are unavailable in http request.

server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
```diff
@@ -47,6 +55,8 @@ public class ApiKeyData {
     // deployment triggers interceptors
     private String initialDeployment;
     private String initialDeploymentApi;
+    // HTTP headers to store during an interceptor invocation chain
```
astsiapanay: I would rephrase the comment.
```
// Original HTTP headers to be stored during an interceptor invocation chain

```

astsiapanay (changes requested):

server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
alekseyvdovenko: updated

alekseyvdovenko: updated

server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
alekseyvdovenko: updated

server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
alekseyvdovenko: updated

astsiapanay (approved):

Issue #560, opened 2024-11-04
Core returns wrong http code if if non match condition fails


astsiapanay:
Core returns 412 instead of 304 when the condition fails for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) methods, then the server must return HTTP status code 304 (Not Modified). For methods that apply server-side changes, the status code 412 (Precondition Failed) is used. Note that the server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 (OK) response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.

PR #561, opened 2024-11-04
fix: Core returns wrong http code if if non match condition fails #560

astsiapanay:


storage/src/main/java/com/epam/aidial/core/storage/util/EtagHeader.java
```diff
@@ -38,21 +43,38 @@ public void validate(Supplier<String> etagSupplier) {
...
+        validateIfMatch(etag);
+
+        validateIfNoneMatch(etag);
+    }
+
+    private void validateIfNoneMatch(String etag) {
         // If-None-Match used with the * value can be used to save a file not known to exist,
         // guaranteeing that another upload didn't happen before
         if (ifNoneMatchTags == null) {
             throw new HttpException(HttpStatus.PRECONDITION_FAILED, "Resource already exists");
```
artsiomkorzun: Should throw 304 for GET/HEAD?

storage/src/main/java/com/epam/aidial/core/storage/http/HttpStatus.java
```diff
@@ -8,6 +8,7 @@
 public enum HttpStatus {
 
     OK(200),
+    NOT_MODIFIED(304),
```
artsiomkorzun, line 11: update fromStatusCode

server/src/main/java/com/epam/aidial/core/server/ProxyContext.java
```diff
@@ -157,10 +157,20 @@ public Future<?> respond(HttpStatus status, String body) {
 
     public Future<?> respond(Throwable error, String fallbackError) {
         return error instanceof HttpException exception
-                ? respond(exception.getStatus(), exception.getMessage())
+                ? respond(exception)
                 : respond(HttpStatus.INTERNAL_SERVER_ERROR, fallbackError);
     }
 
+    public Future<?> respond(HttpException exception) {
+        Map<String, String> headers = exception.getHeaders();
```
artsiomkorzun: nit: you can make it return non-nullable and use response.headers().putAll()

storage/src/main/java/com/epam/aidial/core/storage/util/EtagHeader.java
astsiapanay: right. fixed

storage/src/main/java/com/epam/aidial/core/storage/http/HttpStatus.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/ProxyContext.java
astsiapanay: fixed

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-561.nightly-test.deltixhub.io
>E2E tests status: failed

Issue #569, opened 2024-11-13
ADD_IF_ABSENT Action Saves Files to User Bucket Instead of Public Folder


Gimir:
Steps to Reproduce:

1. Create a publish request (version 1) with the following resource actions:
- Action: ADD for conversations/gpt-4-vision-preview__chameleon
- Action: ADD for files/cham.jpg
2. Publish this first version successfully.
3. Create a second publish request (version 2) with these actions:
- Action: ADD for a new target URL for gpt-4-vision-preview__chameleon
- Action: ADD_IF_ABSENT for files/cham.jpg

PR #570, opened 2024-11-14
fix: ADD_IF_ABSENT Action Saves Files to User Bucket Instead of Publi…

astsiapanay:
…c Folder #569

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
@@ -205,7 +205,7 @@ public class PublicationService {
 
         if (publication.getStatus() == Publication.Status.PENDING) {
             List<Publication.Resource> resourcesToAdd = publication.getResources().stream()
-                    .filter(i -> i.getAction() == Publication.ResourceAction.ADD)
+                    .filter(i -> i.getAction() == Publication.ResourceAction.ADD || i.getAction() == Publication.ResourceAction.ADD_IF_ABSENT)
                     .toList();
             deleteReviewResources(resourcesToAdd);
         }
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
```diff
@@ -223,6 +224,10 @@ public Publication approvePublication(ResourceDescriptor resource) {
                 .filter(i -> i.getAction() == Publication.ResourceAction.ADD)
                 .toList();
 
+        List<Publication.Resource> resourcesToAddIfAbsent = publication.getResources().stream()
```
artsiomkorzun: Please check with checkReviewResources(resourcesToAddIfAbsent).

And delete deleteReviewResources(resourcesToAddIfAbsent).

server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
```diff
@@ -251,7 +256,10 @@ public Publication approvePublication(ResourceDescriptor resource) {
 
         ruleService.storeRules(publication);
 
-        copyReviewToTargetResources(resourcesToAdd);
+        List<Publication.Resource> allResourcesToAdd = new ArrayList<>();
+        allResourcesToAdd.addAll(resourcesToAddIfAbsent);
+        allResourcesToAdd.addAll(resourcesToAdd);
+        copyReviewToTargetResources(allResourcesToAdd);
         deleteReviewResources(resourcesToAdd);
```
artsiomkorzun: Should you take all ADD and ADD_IF_ABSENT everywhere to make it consistent? Right now some methods take, some not.

server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/PublicationService.java
astsiapanay: fixed

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-570.nightly-test.deltixhub.io
>E2E tests status: success

PR #577, opened 2024-11-19
feat: Support multiple rolePath in IDP configuration

akurnosau:
Allow multiple `rolePath` IDP configuration. That is to support cases when roles should be extracted from multiple JWT token claims.
Supports both single String:
```json
{
  "rolePath": "groups"
}
```
or Array of Strings:
```json
{
  "rolePath": [
    "groups",
    "some.path.roles",
    "another.path.roles"
  ]
}
```


akurnosau:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-577.nightly-test.deltixhub.io
>E2E tests status: failed

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -149,32 +163,37 @@ private void evictExpiredJwks() {
...
-                    return (List<String>) next;
-                } else if (next instanceof String) {
-                    if (rolesDelimiter != null) {
-                        return Arrays.stream(((String) next)
-                                .split(rolesDelimiter))
-                                .filter(s -> !s.isBlank())
-                                .toList();
-                    }
-                    return List.of((String) next);
+        List<String> result = new ArrayList<>();
```
astsiapanay, line 167: Could we make two methods: `extractUserRoles(Map<String, Object> map)` and `extractUserRoles(Map<String, Object> map, String rolePath)`?

The current implementation has two nested loops. Instead we can keep the implementation of the original method as is with a minor change: add a new arg `rolePath`. And add the new method will call the original one in the loop. 

 I think that would be more simple. Do you agree?

akurnosau: Agree, refactored

astsiapanay (approved):

PR #582, opened 2024-11-22
chore: re-factor DeleteFileController

astsiapanay:


server/src/main/java/com/epam/aidial/core/server/controller/DeleteFileController.java
```diff
@@ -2,31 +2,19 @@
...
 import com.epam.aidial.core.storage.http.HttpStatus;
 import com.epam.aidial.core.storage.resource.ResourceDescriptor;
-import com.epam.aidial.core.storage.service.LockService;
-import com.epam.aidial.core.storage.service.ResourceService;
 import com.epam.aidial.core.storage.util.EtagHeader;
 import io.vertx.core.Future;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class DeleteFileController extends AccessControlBaseController {
```
artsiomkorzun: Seems like redundant to me. We can call resource controller to delete it.

server/src/main/java/com/epam/aidial/core/server/security/ApiKeyStore.java
```diff
@@ -114,7 +118,7 @@ public Future<Boolean> invalidatePerRequestApiKey(ApiKeyData apiKeyData) {
         String apiKey = apiKeyData.getPerRequestKey();
         if (apiKey != null) {
             ResourceDescriptor resource = toResource(apiKey);
-            return vertx.executeBlocking(() -> resourceService.deleteResource(resource, EtagHeader.ANY), false);
+            return vertx.executeBlocking(() -> resourceOperationService.deleteResource(resource, EtagHeader.ANY), false);
```
artsiomkorzun: Please do not mess it up. Let's only delete prompts/conversations/files/applications with it. Other things are localized and should be left untouched.

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -270,9 +270,11 @@ public void deleteApplication(ResourceDescriptor resource, EtagHeader etag) {
 
         Application application = reference.getValue();
 
+        boolean deleted = true;
         if (isPublicOrReview(resource) && application.getFunction() != null) {
-            deleteFolder(application.getFunction().getSourceFolder());
+            deleted = deleteFolder(application.getFunction().getSourceFolder());
         }
+        return deleted;
```
artsiomkorzun: Not correct. The code above will drop exception.

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
```diff
@@ -82,4 +85,30 @@ public void moveResource(ResourceDescriptor source, ResourceDescriptor destinati
             resourceService.deleteResource(source, EtagHeader.ANY);
         }
     }
+
+    public boolean deleteResource(ResourceDescriptor resource, EtagHeader etag) {
+        String bucketName = resource.getBucketName();
+        String bucketLocation = resource.getBucketLocation();
+        MutableObject<Boolean> deleted = new MutableObject<>();
+        if (resource.isPrivate() && !PublicationService.isReviewBucket(resource)) {
```
artsiomkorzun: && !com.epam.aidial.core.server.service.ApplicationService#hasDeploymentAccess

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
```diff
@@ -82,4 +85,30 @@ public void moveResource(ResourceDescriptor source, ResourceDescriptor destinati
             resourceService.deleteResource(source, EtagHeader.ANY);
         }
     }
+
+    public boolean deleteResource(ResourceDescriptor resource, EtagHeader etag) {
+        String bucketName = resource.getBucketName();
```
artsiomkorzun: Make sure it takes only 4 types: conversations/prompts/files/applications

server/src/main/java/com/epam/aidial/core/server/token/TokenStatsTracker.java
```diff
@@ -69,7 +71,7 @@ public Future<Void> endSpan(ProxyContext context) {
         if (apiKeyData.getPerRequestKey() == null) {
             return vertx.executeBlocking(() -> {
                 ResourceDescriptor resource = toResource(context.getTraceId());
-                resourceService.deleteResource(resource, EtagHeader.ANY);
+                resourceOperationService.deleteResource(resource, EtagHeader.ANY);
```
artsiomkorzun: Do not mess

server/src/main/java/com/epam/aidial/core/server/controller/DeleteFileController.java
astsiapanay: Right I have the same feelings

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
astsiapanay: I guess you want to throw exception instead of return a Boolean result or what?

artsiomkorzun: Line 259 throws an exception if an application does not exist. This code just deletes the app folder.

server/src/main/java/com/epam/aidial/core/server/controller/DeleteFileController.java
astsiapanay: deleted

server/src/main/java/com/epam/aidial/core/server/security/ApiKeyStore.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/token/TokenStatsTracker.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
```diff
@@ -76,10 +85,47 @@ public void moveResource(ResourceDescriptor source, ResourceDescriptor destinati
...
+        ResourceType type = resource.getType();
+        if (!(APPLICATION == type || FILE == type
+                || CONVERSATION == type || type == PROMPT)) {
+            throw new IllegalArgumentException("Unsupported resource type to delete: " + type.name());
+        }
+    }
+
+    private boolean deleteResourceInternally(ResourceDescriptor resource, EtagHeader etag) {
+        if (resource.getType() == APPLICATION) {
+            applicationService.deleteApplication(resource, etag);
```
artsiomkorzun: Should you catch ResourceNotFound and return false in this case?

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
```diff
@@ -225,31 +225,16 @@ private Future<?> deleteResource(ResourceDescriptor descriptor) {
...
-
-                        if (!deleted) {
-                            throw new ResourceNotFoundException();
-                        }
-
-                        return null;
-                    });
-                }, false)
-                .onSuccess(ignore -> context.respond(HttpStatus.OK))
+        vertx.executeBlocking(() -> proxy.getResourceOperationService().deleteResource(descriptor, etag, true), false)
```
artsiomkorzun: The previous version was better. This endpoint works with public and review resources. For which you do not need to cleanup links.

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
astsiapanay: I would make a consistent behavior either we throw exception or return false on resource not found.

it turns out resource service returns false but application service throws an exception for the same case

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
astsiapanay: The previous version required to pass ProxyContext to check a condition if resource links have to be deleted along with the resource.

I decided to pass an arg explicitly so a caller decides what to do.

server/src/main/java/com/epam/aidial/core/server/service/ResourceOperationService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
astsiapanay: reverted to previous version

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-582.nightly-test.deltixhub.io
>E2E tests status: success

Issue #585, opened 2024-11-26
Keep upstream's state in local cache


astsiapanay:
Core keeps upstream's state in local cache for deployments taken from DIAl config.

We want to maintain a consistent behavior for deployments from blob storage as well.

PR #590, opened 2024-11-26
feat: Keep upstream's state in local cache #585

astsiapanay:
Core keeps upstream's state in local cache for deployments taken from DIAl config.

We want to maintain a consistent behavior for deployments from blob storage as well.


server/src/main/java/com/epam/aidial/core/server/config/FileConfigStore.java
```diff
@@ -118,7 +118,6 @@ private void load(boolean fail) {
             }
 
             this.config = config;
-            upstreamRouteProvider.onUpdate(config);
```
artsiomkorzun, line 121: please remove the field

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
```diff
@@ -30,90 +27,88 @@ public class UpstreamRouteProvider {
...
+    public UpstreamRoute get(Route route) {
+        String key = getKey(route);
+        return get(key, route.getUpstreams());
     }
 
-    private static void updateRoutes(Map<String, TieredBalancer> newState, Map<String, TieredBalancer> oldState, Collection<Route> routes) {
-        for (Route route : routes) {
-            String name = route.getName();
+    private UpstreamRoute get(String key, List<Upstream> upstreams) {
+        TieredBalancer balancer = balancers.merge(key, new TieredBalancer(key, upstreams), (prev, next) -> {
```
artsiomkorzun: use compute to not create TieredBalancer all the time, it is quite expensive. Please match the given list with the current tiered balancer

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
```diff
@@ -30,90 +27,88 @@ public class UpstreamRouteProvider {
...
-        if (previousBalancer != null) {
-            log.warn("Duplicate deployment name: {}", name);
+        return prefix + ":" + deployment.getName();
+    }
+
+    private void evictExpiredBalancers() {
+        long currentTime = System.currentTimeMillis();
+        for (Map.Entry<String, TieredBalancer> entry : balancers.entrySet()) {
+            TieredBalancer balancer = entry.getValue();
+            if (currentTime - balancer.getLastAccessTime() > IDLE_PERIOD_IN_MS) {
```
artsiomkorzun: should be updated with the same memory-semantic as you set value. So map.compute and check there.

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/config/FileConfigStore.java
astsiapanay: fixed

artsiomkorzun (dismissed):

config/src/main/java/com/epam/aidial/core/config/Upstream.java
```diff
@@ -14,9 +15,15 @@ public class Upstream {
     public static final int ERROR_THRESHOLD = 3;
 
     private String endpoint;
+
+    @EqualsAndHashCode.Exclude
     private String key;
+
     @JsonDeserialize(using = JsonToStringDeserializer.class)
+    @EqualsAndHashCode.Exclude
     private String extraData;
```
artsiomkorzun: These two fields are used in DeploymentController. So if they are updated and the rest fields are not, you won't update a balancer and you will use the old values of theses fields forever.

artsiomkorzun (changes requested):

config/src/main/java/com/epam/aidial/core/config/Upstream.java
astsiapanay: Right the router may return a cached value of Upstream class. If it returned an upstream's index there wouldn't be any issue.

astsiapanay: fixed

artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-590.nightly-test.deltixhub.io
>E2E tests status: success

PR #605, opened 2024-12-09
feat: Get the Project name from JWT token claims

akurnosau:
Add ability to set Project name using JWT token claims.


server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -248,6 +252,25 @@ private static String extractUserSub(Map<String, Object> userContext) {
         return (String) userContext.get("sub");
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
astsiapanay, line 248: There is a common part of the claim extraction logic for user role and project:

```
for (int i = 0; i < projectPath.length - 1; i++) {
            if (claims.get(projectPath[i]) instanceof Map next) {
                claims = next;
            } else {
                return null;
            }
        }
        Object field = claims.get(projectPath[projectPath.length - 1]);
```

I suppose we can add a new method like

```
private static Object extractClaim(Map<String, Object> claims, String[] claimPath, Object defaultValue)

```
```

astsiapanay:
Please notify @justrp about changes in Core logs. AFAIK they built RegExps to extract info from Core logs for own purposes. 
Your changes will break them.

akurnosau:
> Please notify @justrp about changes in Core logs. AFAIK they built RegExps to extract info from Core logs for own purposes. Your changes will break them.

Thanks for raising this. Done

akurnosau:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-605.nightly-test.deltixhub.io
>E2E tests status: success

astsiapanay (approved):

PR #606, opened 2024-12-10
chore: Improve logging of LockService

astsiapanay:


storage/src/main/java/com/epam/aidial/core/storage/service/LockService.java
```diff
@@ -37,6 +37,7 @@ public LockService(RedissonClient redis, @Nullable String prefix) {
 
     public Lock lock(String key) {
         String id = id(key);
+        log.info("Thread {} acquires a lock the resource {}", Thread.currentThread().getName(), id);
```
artsiomkorzun: please add owner here and in unlock

astsiapanay: fixed

artsiomkorzun (approved):

PR #610, opened 2024-12-11
fix: execution path must contain deployment's name only

astsiapanay:
At the moment the first element of execution path is always either user hash or project name.

Execution path must contain deployment's names only.


artsiomkorzun (approved):

astsiapanay:
/deploy-review
>Environment URL: https://chat-ai-dial-core-pr-610.nightly-test.deltixhub.io
>E2E tests status: failed

Issue #612, opened 2024-12-12
Cut off request/response body if its size exceeds limit


astsiapanay:
Cut off request/response body if its size exceeds limit before saving them to prompt logs

PR #613, opened 2024-12-12
feat: Cut off request/response body if its size exceeds limit #612

astsiapanay:


server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
```diff
@@ -189,10 +192,22 @@ private void append(ProxyContext context, LogEntry entry, String assembledStream
...
     private static void append(LogEntry entry, Buffer buffer) {
-        if (buffer != null) {
-            byte[] bytes = buffer.getBytes();
-            String chars = new String(bytes, StandardCharsets.UTF_8); // not efficient, but ok for now
-            append(entry, chars, true);
+        if (buffer == null) {
+            return;
+        }
+        boolean largeBuffer;
+        if (exceedLimit(buffer)) {
```
artsiomkorzun: simplify with  largeBuffer = exceedLimit(buffer)

server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
```diff
@@ -189,10 +192,22 @@ private void append(ProxyContext context, LogEntry entry, String assembledStream
...
+            buffer = buffer.slice(0, MAX_BODY_SIZE_BYTES);
```
artsiomkorzun, line 204: You can break utf-8 encoded tail. One character can be encoded as 4 bytes in utf-8.

artsiomkorzun: String constructor uses decoder that fixes this broken tail. Could you write a test to see what will be in this case?

server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/log/GfLogStore.java
astsiapanay: good catch!

artsiomkorzun: The decoder will replace with character (65533). So it won't drop an exception at least.

astsiapanay: The decoder replaces broken bytes with 65533 character. 

It won't throw an exception.

artsiomkorzun (approved):

Issue #621, opened 2024-12-19
Setting for tweaking merging strategy for JSON arrays read during configuration load


PMitrafanau:
## What needs to be done
Introduce a setting for tweaking merging strategy for JSON arrays read during configuration load

## Why
Configuration may be presented in multiple JSON files. Files which come later in the list have precedence over those which come before. There may be use cases when something (e.g. model) is configured twice in different files (e.g. configurations come from different independent sources). Current behavior, which is "concatenate arrays", leads to arrays being concatenated, not replaced, which depending on the client may lead to potential problems (e.g. `userRoles` are concatenated instead of replacement).

PR #622, opened 2024-12-19
feat: Setting for tweaking merging strategy for JSON arrays in config…

PMitrafanau:
Issue: #621


server/src/main/java/com/epam/aidial/core/server/config/FileConfigStore.java
```diff
@@ -195,4 +198,17 @@ private static void setMissingFeatures(Deployment model, Features features) {
...
+
+    private JsonMapper buildJsonMapper(JsonObject settings) {
+        JsonMapper mapper = ProxyUtil.MAPPER.copy();
+
+        boolean overwriteArrays = settings
+                .getJsonObject("jsonMergeStrategy", new JsonObject())
+                .getBoolean("overwriteArrays", false);
+
+        mapper.configOverride(ArrayNode.class)
+                .setMergeable(!overwriteArrays);
```
akurnosau, line 212: It is a matter of preference. But I would rename the variable to `mergeArrays`, invert the logic and get rid of that negation. For me it just sounds more naturally. 
I don't insist.

server/src/test/resources/com/epam/aidial/core/server/config/first.config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "models": {
+    "testModel": {
+      "userRoles": [
+        "first_role1"
+      ]
+    }
+  }
+}
```
akurnosau: new line

server/src/test/resources/com/epam/aidial/core/server/config/second.config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "models": {
+    "testModel": {
+      "userRoles": [
+        "second_role1"
+      ]
+    }
+  }
+}
```
akurnosau: one more

akurnosau (dismissed):

server/src/main/java/com/epam/aidial/core/server/config/FileConfigStore.java
PMitrafanau: While I do agree with you as a developer, from user perspective, imho, `overwriteArrays` would be better as it gives more precise information about what will happen (merge is too abstract in my opinion, I didn't even know prior to this case that "merging" in terms of arrays is "concat").

PMitrafanau: Let's wait for @astsiapanay and his opinion, and then decide. I don't have any strong objections about renaming just in case.

astsiapanay: I agree with @PMitrafanau let's make a favor of the user perspective. That makes the property's meaning more simple.
`mergeArrays` may confuse an user: at least what strategy the Core will apply in two cases keep unanswered.

server/src/main/java/com/epam/aidial/core/server/config/FileConfigStore.java
```diff
@@ -195,4 +198,17 @@ private static void setMissingFeatures(Deployment model, Features features) {
             modelFeatures.setContentPartsSupported(features.getContentPartsSupported());
         }
     }
+
+    private JsonMapper buildJsonMapper(JsonObject settings) {
+        JsonMapper mapper = ProxyUtil.MAPPER.copy();
```
astsiapanay: I wouldn't copy the default mapper instance but rather create a new one from scratch. Who knows what settings were set up in that mapper may lead to unpredictable behavior in the config mapper.

PMitrafanau: Yep, that would be better, thx. Fixed

server/src/test/resources/com/epam/aidial/core/server/config/second.config.json
PMitrafanau: Fixed

server/src/test/resources/com/epam/aidial/core/server/config/first.config.json
PMitrafanau: Fixed

akurnosau (approved):

astsiapanay (approved):

PR #633, opened 2025-01-09
feat: code interpreter

artsiomkorzun:


server/build.gradle
```diff
@@ -36,14 +36,14 @@ dependencies {
     implementation 'org.hibernate.validator:hibernate-validator:8.0.0.Final'
     implementation 'org.glassfish:jakarta.el:4.0.2'
     implementation 'jakarta.validation:jakarta.validation-api:3.0.2' // Ensure you have Jakarta Validation API dependency
+    implementation 'org.apache.httpcomponents.client5:httpclient5:5.4'
```
astsiapanay, line 39: We eventually move to blocking http client. ;)

Using NIO http client makes more troubles rather than benefits

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterError.java
```diff
@@ -0,0 +1,13 @@
+package com.epam.aidial.core.server.service.codeinterpreter;
+
+import lombok.Getter;
+
+@Getter
+public class CodeInterpreterError extends RuntimeException {
```
astsiapanay: I think HttpException can replace the class. It has all members and the same purpose.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+            session.setDeploymentUrl(null);
+            session.setUsedAt(null);
+        }
+
+        context.respond(HttpStatus.OK, data);
+    }
+
+    private void respondError(Throwable error) {
+        HttpServerResponse response = context.getResponse();
+        if (response.headWritten()) {
```
astsiapanay, line 184: Why do we do this check?

astsiapanay: Please leave a comment

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+                    return vertx.executeBlocking(() -> service.transferOutputFile(context, data), false);
+                })
+                .onSuccess(this::respondJson)
+                .onFailure(this::respondError);
+
+        return Future.succeededFuture();
+    }
+
+    private void respondJson(Object data) {
+        if (data instanceof CodeInterpreterSession session) {
```
astsiapanay, line 173: Please move the block to the openSession and let's keep the method simple.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+                .onFailure(this::respondError);
+
+        return Future.succeededFuture();
+    }
+
+    Future<?> executeCode() {
+        context.getRequest()
+                .body()
+                .compose(body -> {
+                    CodeInterpreterExecute data = convertJson(body, CodeInterpreterExecute.class);
```
astsiapanay: Do we need to check max body size?

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+        try {
+            T result = ProxyUtil.convertToObject(body, clazz);
+
+            if (result == null) {
+                throw new IllegalArgumentException("No JSON body");
+            }
+
+            return result;
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Not valid JSON body");
```
astsiapanay, line 212: Let's log the error here

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+
+        return Future.succeededFuture();
+    }
+
+    Future<?> uploadFile() {
+        context.getRequest()
+                .setExpectMultipart(true)
+                .uploadHandler(upload -> {
+                    // do not move inside execute blocking, otherwise you can miss the beginning of file
+                    InputStream stream = new InputStreamAdapter(upload);
```
astsiapanay: Stream is never closed

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
```diff
@@ -0,0 +1,215 @@
...
+
+        if (fileName == null) {
+            throw new IllegalArgumentException("Missing filename in multipart upload");
+        }
+
+        return service.uploadFile(context, sessionId, fileName, stream);
+    }
+
+    Future<?> downloadFile() {
+        context.getRequest().body()
```
astsiapanay, line 113: Are we absolutely sure the file always fits to RAM?

Perhaps we want to pipe file's stream to response as we do for a regular file.

server/src/main/java/com/epam/aidial/core/server/data/codeinterpreter/CodeInterpreterExecute.java
```diff
@@ -0,0 +1,15 @@
+package com.epam.aidial.core.server.data.codeinterpreter;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CodeInterpreterExecute {
```
astsiapanay: Let's change the class name to a noun

astsiapanay: As an option we can use names like ExecuteCodeInterpeterRequest or ExecuteCodeEnterpeterResponse

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterClient.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@RequiredArgsConstructor
+public class CodeInterpreterClient {
+
+    // Vertx HttpClient does not support multipart upload, Vertx WebClient supports only Buffer as body for multipart upload
+    private final HttpClient client = HttpClients.createDefault();
+    private final long timeout;
```
astsiapanay: Let's rename to response timeout or add a comment

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+        long footprint = memory.addAndGet(delta);
+        if (footprint <= LOW_MEMORY_BYTES) {
+            stream.fetch(HIGH_MEMORY_BYTES - footprint);
+        } else if (footprint >= HIGH_MEMORY_BYTES) {
+            stream.pause();
+        }
+    }
+
+    @Override
+    public synchronized void close() {
```
astsiapanay: The field is marked as volatile. The keyword `synchronized` is not needed any longer.

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+public class InputStreamAdapter extends InputStream {
+
+    private static final Buffer END_PILL = Buffer.buffer();
+    private static final int LOW_MEMORY_BYTES = 1024 * 1024;
+    private static final int HIGH_MEMORY_BYTES = 4 * 1024 * 1024;
+
+    private final BlockingQueue<Buffer> queue = new LinkedBlockingQueue<>();
+    private final AtomicLong memory = new AtomicLong();
```
astsiapanay: I believe it's a consumed memory of read bytes from the input stream.
Right? Let's make a right name

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public int read() throws IOException {
+        // so dumb - not really used
+        byte[] array = new byte[1];
+        int size = read(array, 0, 1);
+        return (size <= 0) ? -1 : (array[0] & 0xFF);
+    }
+
+    @Override
+    public synchronized int read(byte[] array, int offset, int length) throws IOException {
```
astsiapanay, line 63: Please add Unit test to cover this logic

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
```diff
@@ -0,0 +1,380 @@
...
+            CodeInterpreterSession session = convertToObject(json, CodeInterpreterSession.class);
+
+            if (session != null && predicate.test(session)) {
+                operatorService.deleteCodeInterpreterDeployment(session.getDeploymentId());
+                resourceService.deleteResource(resource, EtagHeader.ANY, false);
+                session = null;
+            }
+
+            if (session == null) {
+                activeSessions.remove(resource.getUrl());
```
astsiapanay, line 119: we can move the line to the block above

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
```diff
@@ -0,0 +1,380 @@
...
+            log.warn("Failed to cleanup active session", e);
+        }
+    }
+
+    public CodeInterpreterSession touchSession(ProxyContext context, String sessionId) {
+        verifyActive();
+        verifySessionId(sessionId);
+
+        ResourceDescriptor resource = sessionResource(context, sessionId);
+        try (LockService.Lock lock = resourceService.lockResource(resource)) {
```
astsiapanay, line 131: Why can we not use resourceService.compute here?

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
```diff
@@ -0,0 +1,380 @@
...
+
+        if (request.getOutputFiles() != null) {
+            for (CodeInterpreterOutputFile output : request.getOutputFiles()) {
+                output.setSessionId(session.getSessionId());
+                transferOutputFile(context, output);
+            }
+        }
+
+        if (anonymous) {
+            closeSession(context, session.getSessionId());
```
astsiapanay: it has to be wrapped by try/finally block

server/build.gradle
artsiomkorzun: For now, the reason is Vertx does not support multipart upload from streaming source.

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterError.java
artsiomkorzun: changed

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: left

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: I can't move it because I need it in all places except this one. Only a client does not need these fields.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: Why?

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: Why? We log all not 200 responses.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: It is allowed to not close this stream. Added this logic any way.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
artsiomkorzun: Request body is a small json object that contains path which to download a file from.

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
artsiomkorzun: reworked

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
artsiomkorzun: Not right, it is the size which currently buffered in memory.

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
artsiomkorzun: please collaborate

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
artsiomkorzun: I need to perform two writes under same lock to make it consistent.

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
artsiomkorzun: Not really.

server/src/main/java/com/epam/aidial/core/server/vertx/stream/InputStreamAdapter.java
artsiomkorzun: added

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
astsiapanay: So the client always receives null values for that fields. Perhaps we want to return just those fields which the clients needs to.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
astsiapanay: The client may send a big payload for a harmful purpose. The second thing Core RAM is limited.
We should check content length before reading the body.

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
astsiapanay: It's not clear what's a reason the error we throw here. how to investigate?

server/src/main/java/com/epam/aidial/core/server/controller/CodeInterpreterController.java
astsiapanay: Sorry I didn't notice the request is a POJO.

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
astsiapanay:
```
if (session != null && predicate.test(session)) {
     operatorService.deleteCodeInterpreterDeployment(session.getDeploymentId());
     resourceService.deleteResource(resource, EtagHeader.ANY, false);
     activeSessions.remove(resource.getUrl());
}

```

astsiapanay:
Please update README with the new settings

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
artsiomkorzun: Moved to finally block.

server/src/main/java/com/epam/aidial/core/server/data/codeinterpreter/CodeInterpreterExecute.java
artsiomkorzun: changed

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterClient.java
artsiomkorzun: renamed to responseTimeout

server/src/main/java/com/epam/aidial/core/server/service/codeinterpreter/CodeInterpreterService.java
astsiapanay: no, it's not. Please check your code

astsiapanay (approved):

PR #644, opened 2025-01-16
chore: properties grouped to container field for applications with type schema

sergey-zinchenko:
* All custom properties from application type schema are now stored in applicationProeprties field of the application
* The user can create an application with schema, but without applicationProeprties, no validation will be performed
* The user can update applicationProeprties from null to some value but not vice versa
* If applicationProeprties is null, the completion request will fail
* CustomApplicationSchemaId became applicationTypeSchemaId

### DIFF
--- a/config/src/test/java/com/epam/aidial/core/config/validation/CustomApplicationsConformToTypeSchemasValidatorTest.java
+++ b/config/src/test/java/com/epam/aidial/core/config/validation/CustomApplicationsConformToTypeSchemasValidatorTest.java
@@ -58,10 +58,43 @@ class CustomApplicationsConformToTypeSchemasValidatorTest {
     }
 
     @Test
+    void isValidReturnsTrueWhenApplicationHasNullProps() {
+        Map<String, Application> applications = new HashMap<>();
+        Application application = new Application();
+        application.setApplicationTypeSchemaId(URI.create("https://mydial.epam.com/custom_application_schemas/specific_application_type"));
+        applications.put("app1", application);
+        config.setApplications(applications);
+        Map<String, String> schemas = new HashMap<>();
+        String customSchemaStr = "{"
+                                 + "\"$schema\": \"https://dial.epam.com/application_type_schemas/schema#\","
+                                 + "\"$id\": \"https://mydial.epam.com/custom_application_schemas/specific_application_type\","
+                                 + "\"dial:applicationTypeEditorUrl\": \"https://mydial.epam.com/specific_application_type_editor\","
+                                 + "\"dial:applicationTypeDisplayName\": \"Specific Application Type\","
+                                 + "\"dial:applicationTypeCompletionEndpoint\": \"http://specific_application_service/opeani/v1/completion\","
+                                 + "\"properties\": {"
+                                 + "  \"file\": {"
+                                 + "    \"type\": \"string\","
+                                 + "    \"format\": \"dial-file-encoded\","
+                                 + "    \"dial:meta\": {"
+                                 + "      \"dial:propertyKind\": \"client\","
+                                 + "      \"dial:propertyOrder\": 1"
+                                 + "    }"
+                                 + "  }"
+                                 + "},"
+                                 + "\"required\": [\"file\"]"
+                                 + "}";
+        schemas.put("https://mydial.epam.com/custom_application_schemas/specific_application_type", customSchemaStr);
+        config.setApplicationTypeSchemas(schemas);
+
+        assertTrue(validator.isValid(config, context));
+    }
+
+    @Test
     void isValidReturnsFalseWhenSchemaValidationFails() {
         Map<String, Application> applications = new HashMap<>();
         Application application = new Application();
         application.setApplicationTypeSchemaId(URI.create("https://mydial.epam.com/custom_application_schemas/specific_application_type"));
+        application.setApplicationProperties(Map.of());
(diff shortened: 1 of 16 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
<ul>

config/src/test/java/com/epam/aidial/core/config/validation/CustomApplicationsConformToTypeSchemasValidatorTest.java
```diff
@@ -57,11 +57,44 @@ void isValidReturnsTrueWhenApplicationsAreEmpty() {
...
 
+    @Test
+    void isValidReturnsTrueWhenApplicationHasNullProps() {
+        Map<String, Application> applications = new HashMap<>();
+        Application application = new Application();
+        application.setApplicationTypeSchemaId(URI.create("https://mydial.epam.com/custom_application_schemas/specific_application_type"));
+        applications.put("app1", application);
+        config.setApplications(applications);
+        Map<String, String> schemas = new HashMap<>();
+        String customSchemaStr = "{"
```
astsiapanay, line 68: Nit. It had better use multiline string syntax:
```
String s = """
You can put 
multiline text
here
""";
```

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -221,6 +220,11 @@ public Pair<ResourceItemMetadata, Application> putApplication(ResourceDescriptor
             Application existing = ProxyUtil.convertToObject(json, Application.class);
             Application.Function function = application.getFunction();
 
+            if (application.getApplicationTypeSchemaId() != null && existing != null
+                    && existing.getApplicationProperties() != null && application.getApplicationProperties() == null) {
+                throw new HttpException(HttpStatus.BAD_REQUEST, "The application with schema can not be updated to the one without properties");
```
astsiapanay, line 225: Error message doesn't sound to me clear. Would you consider `The application properties can't be reset`?

server/src/main/java/com/epam/aidial/core/server/util/ApplicationTypeSchemaUtils.java
```diff
@@ -91,7 +91,10 @@ public static Map<String, Object> getCustomServerProperties(Config config, Appli
         if (customApplicationSchema == null) {
```
astsiapanay, line 91: The condition will never happen. It should be removed

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -421,7 +425,7 @@ public Application.Logs getApplicationLogs(ResourceDescriptor resource) {
     private void prepareApplication(ResourceDescriptor resource, Application application) {
         verifyApplication(resource);
 
-        if (application.getCustomAppSchemaId() != null) {
+        if (application.getApplicationTypeSchemaId() != null) {
             if (application.getEndpoint() != null || application.getFunction() != null) {
                 throw new IllegalArgumentException("Endpoint must not be set for custom application");
```
astsiapanay, line 430: Error message doesn't seem to be correct. It should be `Endpoint or function must not be set for custom application if schema ID is provided`

server/src/main/java/com/epam/aidial/core/server/util/ApplicationTypeSchemaUtils.java
```diff
@@ -91,7 +91,10 @@ public static Map<String, Object> getCustomServerProperties(Config config, Appli
...
             return Collections.emptyMap();
         }
-        return filterProperties(application.getCustomProperties(), customApplicationSchema, "server");
+        if (application.getApplicationProperties() == null) {
+            throw new ApplicationTypeSchemaValidationException("Typed application's properties not set");
```
astsiapanay, line 95: The condition is checked here but it's never verified when the application is created or updated.

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -421,7 +425,7 @@ public Application.Logs getApplicationLogs(ResourceDescriptor resource) {
     private void prepareApplication(ResourceDescriptor resource, Application application) {
         verifyApplication(resource);
 
-        if (application.getCustomAppSchemaId() != null) {
+        if (application.getApplicationTypeSchemaId() != null) {
```
astsiapanay, line 428: Should we add a verification the app custom properties mustn't be null?

server/src/main/java/com/epam/aidial/core/server/util/ApplicationTypeSchemaUtils.java
astsiapanay: Or it should be removed because `AppendApplicationPropertiesFn` will be failed. Since you declare: 

The user can create an application with schema, but without applicationProeprties, no validation will be performed

and at the same time there is a contradiction to the statement: 

If applicationProeprties is null, the completion request will fail

So I can create an app without props but I'm not able to send a chat completion request to my app.

astsiapanay:
@sergey-zinchenko let's merge the MR I'll create an issue to address the comments
that's what we agreed with @sdryapko

astsiapanay (approved):

PR #649, opened 2025-01-20
feat: Add server-specific files to the request for application with schema while proxying the completion request

sergey-zinchenko:
* CollectRequestApplicationFilesFn adds files from server application properties as auto-shared to the request key
* if application properties are null, then throw an internal server error
* if some of the files from properties are not accessible, then throw a forbidden error
* forbidden error on inaccessible or non-existing files added to all the handlers of custom applications (via validation modification)

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
@@ -17,6 +17,7 @@ import com.epam.aidial.core.server.util.ApplicationTypeSchemaProcessingException
 import com.epam.aidial.core.server.util.ApplicationTypeSchemaUtils;
 import com.epam.aidial.core.server.util.ProxyUtil;
 import com.epam.aidial.core.server.util.ResourceDescriptorFactory;
+import com.epam.aidial.core.server.validation.ApplicationTypeResourceException;
 import com.epam.aidial.core.server.validation.ApplicationTypeSchemaValidationException;
 import com.epam.aidial.core.storage.data.MetadataBase;
 import com.epam.aidial.core.storage.data.ResourceItemMetadata;
@@ -37,6 +38,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 import static com.epam.aidial.core.storage.http.HttpStatus.BAD_REQUEST;
+import static com.epam.aidial.core.storage.http.HttpStatus.FORBIDDEN;
 import static com.epam.aidial.core.storage.http.HttpStatus.INTERNAL_SERVER_ERROR;
 
 @Slf4j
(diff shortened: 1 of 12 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
<ul>

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
```diff
@@ -130,11 +131,9 @@ private Future<?> getResource(ResourceDescriptor descriptor, boolean hasWriteAcc
         Future<Pair<ResourceItemMetadata, String>> responseFuture = (descriptor.getType() == ResourceTypes.APPLICATION)
                 ? getApplicationData(descriptor, hasWriteAccess, etagHeader) : getResourceData(descriptor, etagHeader);
 
-        responseFuture.onSuccess(pair -> {
-                    context.putHeader(HttpHeaders.ETAG, pair.getKey().getEtag())
-                            .exposeHeaders()
-                            .respond(HttpStatus.OK, pair.getValue());
-                })
+        responseFuture.onSuccess(pair -> context.putHeader(HttpHeaders.ETAG, pair.getKey().getEtag())
```
astsiapanay, line 134: Why did you do re-formating the code?

server/src/main/java/com/epam/aidial/core/server/validation/ApplicationTypeResourceException.java
```diff
@@ -0,0 +1,20 @@
+package com.epam.aidial.core.server.validation;
+
+import lombok.Getter;
+
+@Getter
+public class ApplicationTypeResourceException extends RuntimeException {
```
astsiapanay, line 12: It's worth adding some doc explaining when the exception may throw.

server/src/main/java/com/epam/aidial/core/server/function/CollectRequestApplicationFilesFn.java
```diff
@@ -0,0 +1,74 @@
...
+                return false;
+            }
+            if (application.getApplicationProperties() == null) {
+                throw new HttpException(HttpStatus.INTERNAL_SERVER_ERROR, "Typed application's properties not set");
+            }
+            List<ResourceDescriptor> resources = ApplicationTypeSchemaUtils.getServerFiles(context.getConfig(), application, proxy.getEncryptionService(),
+                    proxy.getResourceService());
+            ApiKeyData keyData = context.getProxyApiKeyData();
+            appendFilesToProxyApiKeyData(keyData, resources);
+            String perRequestKey = keyData.getPerRequestKey();
```
astsiapanay: We want to create per request key in a single place. Let that will be com.epam.aidial.core.server.controller.DeploymentPostController#handleRequestBody

Once just after the processing functions is done let's assign per request key there.

Note. Please update the method com.epam.aidial.core.server.function.CollectRequestAttachmentsFn#processAttachedFile accordingly.

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
sergey-zinchenko: its not about reformatting but about removing unneeded block brackets

server/src/main/java/com/epam/aidial/core/server/function/CollectRequestApplicationFilesFn.java
sergey-zinchenko: done

server/src/main/java/com/epam/aidial/core/server/validation/ApplicationTypeResourceException.java
sergey-zinchenko: done

astsiapanay (approved):

Issue #729, opened 2025-03-12
Implement LLM Caching


astsiapanay:
### Name and Version

dial 0.26.0

### What is the problem this feature will solve?

Improve response time and reduce LLM usage cost

### What is the feature you are proposing to solve the problem?

The requests should come to upstreams which have cached responses.

PR #731, opened 2025-03-13
feat: Implement LLM Caching #729

astsiapanay:


ai-dial-actions:
<h1>Dependency Review</h1>
<ul>

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,194 @@
...
+
+            map.putAll(fields);
+
+            Instant expireAt = expireAt(expireAtStr);
+            if (expireAt == null && exists) {
+                // adapter didn't return expireAt for a new cache entry
+                expireAt = Instant.now().plus(DEFAULT_TTL);
+            }
+            if (expireAt != null) {
+                map.expire(expireAt);
```
artsiomkorzun, line 184: Why don't you always reset expiration?

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,194 @@
...
+    private static String toString(byte[] digest) {
+        StringBuilder hexString = new StringBuilder();
+        for (byte b : digest) {
+            hexString.append(String.format("%02x", b));
+        }
+        return hexString.toString();
+    }
+
+    private String getEntryKey(String modelName, String hash) {
+        return prefix + ":" + modelName + ":" + hash;
```
artsiomkorzun: Please use the same way of how we build prefix. Something like:
```
"upstream:" + BlobStorageUtil.toStoragePath(prefix, BlobStorageUtil.toStoragePath(modelName, hash))
```

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRoute.java
```diff
@@ -77,9 +93,15 @@ public Upstream next() {
...
-            throw balancer.createUpstreamUnavailableException();
+        if (upstreamCacheContext == null
+                || (upstreamCacheContext.getPolicy() != CachePolicy.CACHE_PRIORITY && attemptCount > 1)
+                || upstreamCacheContext.getOriginalUpstream() == null) {
+            upstream = balancer.next(usedUpstreams);
+            if (upstream == null) {
+                throw balancer.createUpstreamUnavailableException();
+            }
+        } else {
+            upstream = upstreamCacheContext.getOriginalUpstream();
```
artsiomkorzun, line 104: Where do you check that the upstream is still present in the config?

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,194 @@
...
+            for (; i >= 0 && count < BATCH_SIZE; count++, i--) {
+                String prefixPath = breakpoints.get(i);
+                String hash = prefixToHash.get(prefixPath);
+                String key = getEntryKey(model.getName(), hash);
+                batch.getMap(key, REDIS_MAP_CODEC).getAllAsync(ALL_FIELDS);
+            }
+            BatchResult<?> result = batch.execute();
+            List<?> responses = result.getResponses();
+            for (int j = 0; j < count; j++) {
+                Map<String, String> map = (Map<String, String>) responses.get(j);
```
artsiomkorzun: We support volatile cache policy, that means that Redis could return stale entries with expired ttl.

astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,219 @@
...
+            List<?> responses = result.getResponses();
+            for (int j = 0; j < responses.size(); j++) {
+                Boolean exists = (Boolean) responses.get(j);
+                if (exists) {
+                    int index = i + count - j;
+                    String prefixPath = breakpoints.get(index);
+                    String hash = prefixToHash.get(prefixPath);
+                    String key = getEntryKey(model.getName(), hash);
+                    RMap<String, String> map = redisClient.getMap(key, REDIS_MAP_CODEC);
+                    Map<String, String> fields = map.getAll(ALL_FIELDS);
```
artsiomkorzun, line 150: It could be already missing, it might be evicted between two calls

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,219 @@
...
+                    long expiredAt = Long.parseLong(fields.get(EXPIRED_AT_FIELD));
```
artsiomkorzun: Will fail with null pointer

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+            if (expireAt != null) {
+                fields.put(EXPIRED_AT_FIELD, Long.toString(expireAt.toEpochMilli()));
+            }
+
+            map.putAll(fields);
+
+            if (expireAt != null) {
+                map.expire(expireAt);
```
artsiomkorzun: It could be already evicted, and you do not set expiration

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,219 @@
...
+
+    public void updateEntry(String hash, CachedUpstreamEntry entry, Model model, String expireAtStr) {
+        String key = getEntryKey(model.getName(), hash);
+        Map<String, String> fields = new HashMap<>();
+        fields.put(UPSTREAM_ENDPOINT_FIELD, entry.endpoint());
+        fields.put(PREFIX_PATH_FIELD, entry.prefixPath());
+        try (var ignore = lockService.lock(key)) {
+
+            RMap<String, String> map = redisClient.getMap(key, REDIS_MAP_CODEC);
+            boolean exists = map.isExists();
```
artsiomkorzun: Check for expiration field?

artsiomkorzun (dismissed):

config/src/main/java/com/epam/aidial/core/config/Model.java
```diff
@@ -17,6 +18,8 @@ public class Model extends Deployment {
     private List<Upstream> upstreams = List.of();
     // if it's set then the model name is overridden with that name in the request body to the model adapter
     private String overrideName;
+    @JsonAlias({"fieldsHashingOrder", "fields_hashing_order"})
+    private List<String> fieldsHashingOrder = List.of("prefix.body.tools", "prefix.body.messages");
```
adubovik, line 22: Dot as a separator in a list is a very unusual choice.
`prefix.body.tools` looks like a path in a nested structure.

1. 

How about a comma? `prefix,body,tools`

server/src/main/java/com/epam/aidial/core/server/data/cache/CachePolicy.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public static CachePolicy fromString(String val) {
+        for (CachePolicy policy : CachePolicy.values()) {
+            if (policy.val.equals(val)) {
+                return policy;
+            }
+        }
+        // fallback
+        return AVAILABILITY_PRIORITY;
```
adubovik: I don't think that we should tolerate invalid `X-DIAL-CACHE-POLICY` headers by falling back to `AVAILABILITY_PRIORITY`.

Imagine, someone made a typo in the header: `X-DIAL-CACHE-POLICY: cache-priorty`
It will be treated as `AVAILABILITY_PRIORITY`, which is very surprising for the client.

Why not reject invalid requests with 400?

config/src/main/java/com/epam/aidial/core/config/Model.java
adubovik, line 22: I don't see any meaning in the `prefix.body.` prefix.
Simply `List.of("tools", "messages");` is enough.

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,211 @@
...
+            String nodeName = matcher.group("nodeName");
+            JsonNode node = body.get(nodeName);
+            if (node == null || !node.isArray()) {
+                // embedding request is not supported yet
+                continue;
+            }
+            ArrayNode arrayNode = (ArrayNode) node;
+            for (int index = 0; index < arrayNode.size(); index++) {
+                ObjectNode objectNode = (ObjectNode) arrayNode.get(index);
+                for (Iterator<Map.Entry<String, JsonNode>> it = objectNode.fields(); it.hasNext(); ) {
```
adubovik: How do you ensure that the order in which JSON dictionary fields are enumerated is fixed?

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,211 @@
...
+                    if (entry.getKey().equals(CUSTOM_FIELDS_NODE)) {
+                        continue;
+                    }
+                    if (entry.getKey().equals("custom_content")) {
+                        // include attachments only
+                        if (entry.getValue().has("attachments")) {
+                            messageDigest.update(entry.getValue().get("attachments").toString().getBytes(StandardCharsets.UTF_8));
+                        }
+                    } else {
+                        messageDigest.update(entry.getValue().toString().getBytes(StandardCharsets.UTF_8));
```
adubovik, line 108: How do you ensure that `.toString()` serializes the fields of the same JSON dictionary in the same order?

https://stackoverflow.com/questions/27577701/jackson-objectmapper-specify-serialization-order-of-object-properties

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+        try {
+            return Instant.ofEpochSecond(Long.parseLong(val));
+        } catch (NumberFormatException e) {
+            log.error("Invalid expireAt datetime format: " + val);
+            return null;
+        }
+    }
+
+    private static String toString(byte[] digest) {
```
adubovik, line 194: Looks like dead code.

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRoute.java
```diff
@@ -114,10 +136,46 @@ void fail(HttpStatus status, long retryAfterSeconds) {
...
 
+    private void updateUpstreamCache(HttpClientResponse proxyResponse, Model model) {
+        String breakpointPath = proxyResponse.getHeader(Proxy.HEADER_CACHE_BREAKPOINT_PATH);
+        if (breakpointPath == null) {
+            // no cache
+            return;
+        }
+        String hash = upstreamCacheContext.getPrefixToHash().get(breakpointPath);
+        if (hash == null) {
+            if (model.getType() != ModelType.EMBEDDING) {
```
adubovik: `if (model.getType() == ModelType.CHAT) {`

because we only support caching for chat deployments.

server/src/main/java/com/epam/aidial/core/server/data/cache/CachePolicy.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRoute.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
astsiapanay: According to javadoc:

`Feature that defines default property serialization order used for POJO properties. If enabled, default ordering is alphabetic (similar to how com.fasterxml.jackson.annotation.JsonPropertyOrder.alphabetic() works); if disabled, order is unspecified (based on what JDK gives us, which may be declaration order, but is not guaranteed).`

The service take JsonNode is an arbitrary JSON object. So I had to write own serializer.

astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/UpstreamCacheService.java
```diff
@@ -182,6 +174,54 @@ public void updateEntry(String hash, CachedUpstreamEntry entry, Model model, Str
         }
     }
 
+    @VisibleForTesting
+    static void traverse(JsonNode node, StringBuilder content) {
```
adubovik: Why not sort the dictionary by default?

https://github.com/epam/ai-dial-core/blob/ecc3c355a08d3c5f98bb53de2489470d80cd010e/server/src/main/java/com/epam/aidial/core/server/util/ProxyUtil.java#L35-L37

```diff
+ import com.fasterxml.jackson.databind.SerializationFeature;

    public static final JsonMapper MAPPER = JsonMapper.builder()
            .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+           .enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)
            .build();
```

astsiapanay: it doesn't work for JsonNode. See https://stackoverflow.com/a/36465506

astsiapanay: This approach https://cowtowncoder.medium.com/jackson-tips-sorting-json-using-jsonnode-ce4476e37aee should work

astsiapanay: There are 3 options to make the fixed order of object properties:

1. Make a single ObjectMapper with the ordered object factory for all functions. That implies that the service depends on a caller. The caller must provide JSON object with ordered properties. We don't need this for all cases.
2. Change the method's signature by passing byte array instead of Json object. That makes the method to parse the entire request body one more time.
3. Sort object properties recursively in place. The approach is more light weight than the 2nd option and more reliable than the 1st one.

astsiapanay: fixed

adubovik (approved):

astsiapanay:
/deploy-review
>Environment: [review-environment](https://chat-ai-dial-core-pr-731.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/583236)
>Overlay E2E tests status: [success](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/jobs/1716723)
>Chat E2E tests status: [success](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/jobs/1716722)

PR #1009, opened 2025-08-29
fix: Auth statuses incorrect after signin into ToolSet

### LINKED ISSUE #1008: [ToolSets] Auth statuses incorrect after signin into ToolSet
### Name and Version

dial-core

1a. Signin into ToolSet with "Global" credentials.
2a. Get ToolSet and check "global_auth_status" property.

1b. Signin into ToolSet with "USER" credentials.
2b. Get ToolSet and check "user_level_auth_status" property.

a. "global_auth_status" should be "SIGNED_IN"

b. "user_level_auth_status" should be "SIGNED_IN"

### What do you see instead?

a. "global_auth_status" is "SIGNED_OUT"

b. "user_level_auth_status" is "SIGNED_OUT"

### Additional information

Intermittent issue. Might be reproduced after few signin operations and when some of credentails from previous signins are expired.

DmytroZaichenkoDev:
### Applicable issues

- fixes #1008

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
@@ -84,8 +84,7 @@ public class ResourceAuthSettingsService {
         Optional<ResourceCredentials> userResourceCredentials = resourceCredentialsList.stream()
                 .filter(resourceCredentials -> resourceCredentials.getCredentialsLevel().equals(CredentialsLevel.USER)
                     && resourceCredentials.getAuthenticationType().equals(AuthenticationType.OAUTH)
-                    && resourceCredentials.getUserSub() != null
-                    && resourceCredentials.getUserSub().equals(userSub)
+                    && userSub.equals(resourceCredentials.getUserSub())
                     && !resourceCredentials.isTokenExpired())
                 .findFirst();
         // TODO: implement logic for API_KEY auth type
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
@@ -71,18 +71,25 @@ public class ResourceAuthSettingsService {
     }
 
     public void setResourceAuthStatuses(String resourceId,
-                                        ResourceAuthSettings resourceAuthSettings) {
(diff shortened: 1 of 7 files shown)


credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
```diff
@@ -71,18 +71,25 @@ private boolean shouldRegisterResourceDynamically(ResourceAuthSettings resourceA
...
     private void setUserAuthStatus(ResourceAuthSettings resourceAuthSettings,
-                                   List<ResourceCredentials> resourceCredentialsList) {
+                                   List<ResourceCredentials> resourceCredentialsList,
+                                   String userSub) {
         Optional<ResourceCredentials> userResourceCredentials = resourceCredentialsList.stream()
-                .filter(resourceCredentials -> resourceCredentials.getCredentialsLevel().equals(CredentialsLevel.USER))
+                .filter(resourceCredentials -> resourceCredentials.getCredentialsLevel().equals(CredentialsLevel.USER)
+                    && resourceCredentials.getAuthenticationType().equals(AuthenticationType.OAUTH)
+                    && resourceCredentials.getUserSub() != null
+                    && resourceCredentials.getUserSub().equals(userSub)
```
astsiapanay: Nit. That might be more simple to check 
userSub.equals(credentials.getUserSub()) since userSub can’t be null.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
```diff
@@ -71,18 +71,25 @@ private boolean shouldRegisterResourceDynamically(ResourceAuthSettings resourceA
...
+                    && !resourceCredentials.isTokenExpired())
```
astsiapanay, line 88: What if user has multiple credentials and we always take the first one. The rest of ones won’t be updated. Why?

DmytroZaichenkoDev: Eventually, user won't have more then one personal credentials of some type (OAUTH for example) for single ToolSet. This would be implemented in separate PR's. 

So the idea here is to find if user has at least one non-expired credentials of some type. in our case it's OAUTH type, USER level. It's enough for Auth Status calculation.

astsiapanay (approved):

PR #1021, opened 2025-09-03
fix: [ToolSets] Incorrect response on ToolSet creation when OAuth is unsupported by MCP server

### LINKED ISSUE #1020: [ToolSets] Incorrect response on ToolSet creation when OAuth is unsupported by MCP server
### Name and Version

dial-core

Create ToolSet with toolset.auth_settings.auth_type = OAUTH and MCP server, which doesn't support OAuth (toolset.endpoint=https://mcp.deepwiki.com/mcp).

Server returns 400 Bad Request with message: "The MCP server does not support OAuth authentication."

### What do you see instead?

Server returns 500 Internal Server Error.

### Additional information

_No response_

DmytroZaichenkoDev:
### Applicable issues

- fixes #1020 

### Checklist


credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
```diff
@@ -58,6 +58,8 @@ public void enrichResourceAuthSettings(String resourceId,
             resourceAuthSettings.setCodeChallenge(codeChallenge.getValue());
             resourceAuthSettings.setCodeVerifier(codeVerifier.getValue());
             resourceAuthSettings.setCodeChallengeMethod(codeChallengeMethod.getValue());
+        } catch (IllegalArgumentException e) {
```
astsiapanay: why do we need this if we just re-throw the exception further?

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -35,6 +35,8 @@ public <R> R executeGet(String url, Class<R> responseType) {
             Thread.currentThread().interrupt();
             log.error("Request was interrupted: {}", e.getMessage(), e);
             throw new CredentialsInternalException("Request was interrupted", e);
+        } catch (HttpException e) {
+            throw e;
         } catch (Exception e) {
             log.error("Unexpected error occurred: {}", e.getMessage(), e);
             throw new CredentialsInternalException("Unexpected error occurred while making GET request", e);
```
astsiapanay: What's the point to wrap of the original exception where we handle it?

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -66,6 +68,8 @@ public <R> R executePost(String url, Object requestPayload, String contentType,
             Thread.currentThread().interrupt();
```
astsiapanay: Why do we interrupt the interrupted thread?

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
DmytroZaichenkoDev: we don't need next (more generic) catch block to handle it and wrap with CredentialsInternalException, so controller can handle it later and throw 400 instead of 500.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
DmytroZaichenkoDev: It was a warning from Sonar and I've followed suggested 'Compliant solution': 
https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2142/

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
DmytroZaichenkoDev: Fixed in com.epam.aidial.core.server.controller.ResourceController#handleError
Client will get 500 Internal server error, but with client-friendly message.

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
```diff
@@ -329,6 +330,8 @@ private void handleError(ResourceDescriptor descriptor, Throwable error) {
             context.respond(HttpStatus.NOT_FOUND, "Not found: " + descriptor.getUrl());
         } else if (error instanceof PermissionDeniedException) {
             context.respond(HttpStatus.FORBIDDEN, error.getMessage());
+        } else if (error instanceof CredentialsInternalException) {
```
astsiapanay: We could create an http exception as we want and send back to the client. I'm not sure if we need a specialized exception for the matter.

server/src/main/java/com/epam/aidial/core/server/AiDial.java
```diff
@@ -177,7 +177,8 @@ void start() throws Exception {
             UpstreamCacheService upstreamCacheService = new UpstreamCacheService(redis, lockService, clock, storage.getPrefix());
             UpstreamRouteProvider upstreamRouteProvider = new UpstreamRouteProvider(vertx, taskExecutor, Random::new, upstreamCacheService);
 
-            ResourceAuthorizationClient resourceAuthorizationClient = new ResourceAuthorizationClient();
+            java.net.http.HttpClient httpClient = java.net.http.HttpClient.newHttpClient();
```
astsiapanay: Nit. It seems like a trade-off between keeping internals and making better testability.
One more option: create two constructors one for the runtime execution and one for testing and mark the second one with the annotation @VisibleForTesting and package level access only.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
astsiapanay: all right but you do both re-throwing and interrupting. Is that right?

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -66,6 +68,8 @@ public <R> R executePost(String url, Object requestPayload, String contentType,
...
             log.error("Request was interrupted: {}", e.getMessage(), e);
             throw new CredentialsInternalException("Request was interrupted", e);
+        } catch (HttpException e) {
+            throw e;
```
astsiapanay: Please don't logging heavy objects like http response. See the line 86.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -66,6 +68,8 @@ public <R> R executePost(String url, Object requestPayload, String contentType,
             Thread.currentThread().interrupt();
             log.error("Request was interrupted: {}", e.getMessage(), e);
             throw new CredentialsInternalException("Request was interrupted", e);
+        } catch (HttpException e) {
```
astsiapanay: Please use logging level wisely. Since devops see any log messages with error level and rise an alert.

Perhaps you want to use info or even debug. Since this is not Core's issue.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceRegistrationService.java
```diff
@@ -72,22 +73,28 @@ public ClientRegistration createStaticResourceRegistration(String resourceId,
...
-        return ClientRegistration.builder()
-            .resourceId(resourceId)
-            .clientId(resourceAuthSettings.getClientId())
-            .clientSecret(resourceAuthSettings.getClientSecret())
-            .redirectUri(resourceAuthSettings.getRedirectUri())
-            .authorizationEndpoint(resourceAuthorizationEndpoint)
-            .tokenEndpoint(tokenEndpoint)
-            .codeChallengeMethod(codeChallengeMethodSupported)
-            .build();
+    private RuntimeException processHttpException(HttpException e, String resourceId) {
```
astsiapanay: The method definition is a bit weird. We do error handling here and return a runtime exception. How is that so :) ?

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceRegistrationService.java
```diff
@@ -72,22 +73,28 @@ public ClientRegistration createStaticResourceRegistration(String resourceId,
...
+        log.error(e.getMessage(), e);
+        HttpStatus status = e.getStatus();
+        if (status.equals(HttpStatus.NOT_FOUND) || status.equals(HttpStatus.UNAUTHORIZED)) {
```
astsiapanay: Why do we want to handle this in auth client and return:
and log message `MCP server return http code {}, response {}`
 400 with clear message `"The MCP server for Resource: %s does not support OAuth authentication.".formatted(resourceId)`

that's simple and understandable

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
DmytroZaichenkoDev: fixed. removed this code

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
DmytroZaichenkoDev: fixed. CredentialsInternalException has been removed.

server/src/main/java/com/epam/aidial/core/server/AiDial.java
DmytroZaichenkoDev: fixed

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceRegistrationService.java
DmytroZaichenkoDev: fixed

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceRegistrationService.java
DmytroZaichenkoDev: Moved this to com.epam.aidial.core.credentials.service.ResourceRegistrationService#getResourceAuthorizationServerEndpoint

since it's very specific case when we've checked both well-known endpoints and got 401/404 from both, so we assume that MCP doesn't support OAuth. I think ResourceRegistrationService is better place for this handling, then http client.

astsiapanay (approved):

Issue #1019, opened 2025-09-03
Schema-rich application doesn't have access to toolset from own config


astsiapanay:
### Name and Version

core 0.34

### What steps will reproduce the bug?

1. Create schema-rich app with toolsets
2. Call the app

### What is the expected behavior?

The app should call toolset and return response

### What do you see instead?

The core returns 403 because the toolset is not auto-shared to the app

PR #1022, opened 2025-09-03
fix: Schema-rich application doesn't have access to toolset from own config #1019

astsiapanay:


server/src/main/java/com/epam/aidial/core/server/service/ApplicationSchemaService.java
```diff
@@ -267,6 +270,45 @@ private List<ResourceDescriptor> getFiles(Application application, ListCollector
...
+                    result.add(descriptor);
+                } catch (IllegalArgumentException e) {
+                    // ignored
+                }
+            }
+            return result;
+        } catch (ApplicationTypeSchemaValidationException | ApplicationTypeResourceException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new ApplicationTypeSchemaProcessingException("Failed to obtain list of files attached to the custom app", e);
```
DmytroZaichenkoDev: typo: "files" instead of "toolsets".

maybe add TODO to re-factor this method and avoid code duplication in getFiles() and getToolSets()?

astsiapanay: fixed

DmytroZaichenkoDev (approved):

PR #1076, opened 2025-09-22
fix(toolset): toolset list fails if the encryption method changes

oleksii-donets:
### Applicable issues

- fixes #

### Description of changes

Toolsets encrypted with an outdated method can no longer be decrypted.
Instead of returning a corrupted result, we now treat them as not found and do not return them.

### Checklist

- [ ] Title of the pull request follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)


ai-dial-actions:
<h1>Dependency Review</h1>
None

DmytroZaichenkoDev (approved):

PR #1077, opened 2025-09-24
fix: [ToolSets] Auth statuses incorrect after signin into ToolSet with special characters in name

### LINKED ISSUE #1011: [ToolSets] Auth statuses incorrect after signin into ToolSet with special characters in name
### Name and Version

dial-core

1. Create ToolSet with name containing " " character ("ToolSet 1").
2a. Signin into ToolSet with 'Global' credentials.
2b. Signin into ToolSet with 'User' credentials.
3. Get ToolSet.

a. "global_auth_status" should be "SIGNED_IN"

b. "user_level_auth_status" should be "SIGNED_IN"

### What do you see instead?

a. "global_auth_status" is "SIGNED_OUT"

b. "user_level_auth_status" is "SIGNED_OUT"

### Additional information

_No response_

DmytroZaichenkoDev:
### Applicable issues

- fixes #1011

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/controller/ToolSetCredentialsController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/ToolSetCredentialsController.java
@@ -52,6 +52,7 @@ public class ToolSetCredentialsController {
         this.resourceCredentialsService = proxy.getResourceCredentialsService();
     }
 
+    // TODO: fix case with toolset from config
     public Future<?> signIn() {
         context.getRequest()
                 .body()
@@ -62,9 +63,10 @@ public class ToolSetCredentialsController {
                     return taskExecutor.submit(() -> {
                         Deployment deployment = deploymentService.findDeployment(context, toolsetId);
                         if (deployment instanceof ToolSet toolSet) {
-                            verifyAccess(toolsetId, resourceSignInRequest.getCredentialsLevel());
+                            String encodedResourceUrl = UrlUtil.encodePath(toolsetId);
+                            verifyAccess(encodedResourceUrl, resourceSignInRequest.getCredentialsLevel());
                             CredentialsDescriptor credentialsDescriptor = CredentialsDescriptorFactory.fromAnyUrl(
-                                    toolsetId, resourceSignInRequest.getCredentialsLevel(), context);
+                                    encodedResourceUrl, resourceSignInRequest.getCredentialsLevel(), context);
                             resourceCredentialsService.addResourceCredentials(
                                     credentialsDescriptor,
                                     toolSet.getAuthSettings(),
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/epam/aidial/core/server/controller/ToolSetCredentialsController.java
```diff
@@ -62,9 +63,10 @@ public Future<?> signIn() {
                     return taskExecutor.submit(() -> {
                         Deployment deployment = deploymentService.findDeployment(context, toolsetId);
                         if (deployment instanceof ToolSet toolSet) {
-                            verifyAccess(toolsetId, resourceSignInRequest.getCredentialsLevel());
+                            String encodedResourceUrl = UrlUtil.encodePath(toolsetId);
+                            verifyAccess(encodedResourceUrl, resourceSignInRequest.getCredentialsLevel());
                             CredentialsDescriptor credentialsDescriptor = CredentialsDescriptorFactory.fromAnyUrl(
-                                    toolsetId, resourceSignInRequest.getCredentialsLevel(), context);
+                                    encodedResourceUrl, resourceSignInRequest.getCredentialsLevel(), context);
```
oleksii-donets, line 69: Why don't we update CredentialsDescriptorFactory and CredentialsLocatorFactory factories instead?

server/src/test/java/com/epam/aidial/core/server/controller/ToolSetCredentialsControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+    @Mock private EncryptionService encryptionService;
+    @Mock private DeploymentService deploymentService;
+    @Mock private Proxy proxy;
+    @Mock private HttpServerRequest request;
+
+    private ToolSetCredentialsController controller;
+
+    @BeforeEach
+    void setup() {
+        MockitoAnnotations.openMocks(this);
```
oleksii-donets: I am curious why not use @ExtendWith(MockitoExtension.class) instead? Seems clearer to me

server/src/main/java/com/epam/aidial/core/server/controller/ToolSetCredentialsController.java
DmytroZaichenkoDev: in some places (com.epam.aidial.core.server.controller.ToolSetController#getToolSet for example) this url is already encoded, so it will break such cases.
we can think about refactoring later if we wanna streamline it across other services, but for this fix I prefer to keep it as it is.

server/src/test/java/com/epam/aidial/core/server/controller/ToolSetCredentialsControllerTest.java
DmytroZaichenkoDev: fixed

oleksii-donets (approved):

PR #1457, opened 2026-04-13
chore(deps): update dependency review configuration

melvinsh:
Routine dependency review update.


melvinsh:
Closing — workflow is currently disabled with `if: false`. Will test against other affected repos.

PR #1480, opened 2026-04-21
fix: [Toolsets] Return null instead of throwing ResourceNotFoundException when no credentials stored

### LINKED ISSUE #1478: [Toolsets] False ResourceNotFoundException logged as ERROR when no credentials stored
## Problem

When a toolset with `authentication_type=OAUTH` is invoked but the user hasn't signed in yet, `ResourceCredentialsService.getRefreshedResourceCredentials()` throws `ResourceNotFoundException`. This exception is caught by `ToolSetProxyController.setToolsetCredentials()` and logged as **ERROR**, even though it's a valid flow — the request proceeds to the MCP server without auth headers.

This creates noisy false-positive errors in production logs.

## Root Cause

`getAndRefreshCredentials()` throws `ResourceNotFoundException` when no credentials exist in storage (line 220). The USER-level lookup is wrapped in a try-catch, but the GLOBAL-level lookup is not — so the exception propagates up to `ToolSetProxyController` where it's logged as ERROR.

## Fix

- `getAndRefreshCredentials()` returns `null` instead of throwing when no credentials exist
- `getRefreshedResourceCredentials()` returns `null` instead of throwing when neither USER nor GLOBAL credentials are found
- All callers already handle `null` correctly (`ToolSetProxyController`, `ToolSetToolsController`, `AuthorizationHeaderProvider`)

No behavior change for clients — request proceeds without auth header in both cases.

DmytroZaichenkoDev:
getRefreshedResourceCredentials() was throwing ResourceNotFoundException when no credentials existed, causing false ERROR logs in ToolSetProxyController. Now returns null which all callers already handle correctly.

### Applicable issues

- fixes #1478 

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
@@ -178,18 +178,14 @@ public class ResourceCredentialsService {
             return null;
         }
 
-        try {
-            ResourceCredentials userCredentials = getAndRefreshCredentials(
-                    credentialsLocator.getCredentialsDescriptors().get(CredentialsLevel.USER),
-                    authSettings
-            );
-            if (userCredentials != null
-                    && userCredentials.getCredentialsLevel().equals(CredentialsLevel.USER)
-                    && userCredentials.getUserId().equals(userSub)) {
-                return userCredentials;
-            }
-        } catch (ResourceNotFoundException e) {
-            log.debug(e.getMessage(), e); // if User credentials are not found - let's look for Global one
+        ResourceCredentials userCredentials = getAndRefreshCredentials(
+                credentialsLocator.getCredentialsDescriptors().get(CredentialsLevel.USER),
+                authSettings
+        );
+        if (userCredentials != null
+                && userCredentials.getCredentialsLevel().equals(CredentialsLevel.USER)
+                && userCredentials.getUserId().equals(userSub)) {
+            return userCredentials;
         }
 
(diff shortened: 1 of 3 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
```diff
@@ -178,18 +178,14 @@ public ResourceCredentials getRefreshedResourceCredentials(CredentialsLocator cr
...
-        } catch (ResourceNotFoundException e) {
-            log.debug(e.getMessage(), e); // if User credentials are not found - let's look for Global one
+        ResourceCredentials userCredentials = getAndRefreshCredentials(
+                credentialsLocator.getCredentialsDescriptors().get(CredentialsLevel.USER),
+                authSettings
+        );
+        if (userCredentials != null
+                && userCredentials.getCredentialsLevel().equals(CredentialsLevel.USER)
+                && userCredentials.getUserId().equals(userSub)) {
+            return userCredentials;
```
astsiapanay, line 191: It’s worth annotating the method with @Nullable

DmytroZaichenkoDev: fixed

astsiapanay (dismissed):

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/24785633113)

astsiapanay (approved):

PR #1486, opened 2026-04-22
fix: Remove OAuth toolset credentials on 400/401 from token refresh flow

### LINKED ISSUE #1385: [Toolsets] Set OAuth toolset 'LOGGED IN' status based on refresh token validity instead of access token
### Name and Version

dial-core

1. Create OAuth Toolset and log in into it.
2. Wait for access token to expire (~1 hour).

Toolset status still 'LOGGED IN'.

### What do you see instead?

Toolset status is 'LOGGED OUT'.

### Additional information

_No response_

DmytroZaichenkoDev:
The refresh flow previously only removed stored credentials when the authorization server returned a spec-compliant `{"error":"invalid_grant"}` body. Providers that respond with a different OAuth error code (e.g., Notion MCP returning 401 + `invalid_token`) left credentials on disk, so the UI kept showing SIGNED_IN while every request failed 401 in a loop.

Now any HTTP 400 or 401 from the token endpoint removes the stored credentials so the user is prompted to sign in again. Transient failures (5xx, network errors) still rethrow and keep credentials intact.

Also elevates lifecycle logs (sign-in, sign-out, delete, refresh start/end) from DEBUG to INFO so production operators can see toolset-auth activity without enabling DEBUG.

### Applicable issues

- fixes #1385 

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
@@ -83,7 +83,7 @@ public class ResourceAuthorizationClient {
             String body = response.body();
 
             if (status != 200 && status != 201) {
-                log.debug("Error executing request {}: status {}, response {}, headers: {}",
+                log.info("Error executing request {}: status {}, response {}, headers: {}",
                         request.uri(), response.statusCode(), response.body(), response.headers());
                 if (status == 401) {
                     throw new HttpException(HttpStatus.UNAUTHORIZED, "Authorization server returns 401 error code",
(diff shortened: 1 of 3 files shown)


ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/24778590946)

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -83,8 +83,8 @@ private <R> R execute(HttpRequest request, Class<R> responseType) {
             String body = response.body();
 
             if (status != 200 && status != 201) {
-                log.debug("Error executing request {}: status {}, response {}, headers: {}",
-                        request.uri(), response.statusCode(), response.body(), response.headers());
+                log.info("Error executing request {}: status {}, response {}",
```
astsiapanay: Perhaps log level `warn` is more appropriate

astsiapanay (approved):

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/24781237695)

PR #1496, opened 2026-04-24
fix: [Toolsets] Decode gzip-encoded OAuth discovery responses #1493

### LINKED ISSUE #1493: 400 bad request (schema validation error) when saving external mcp toolset in quickapp
### Name and Version

0.42.2

Backend returns  400 bad request error during the save process of an external mcp toolset within quickapp, citing a payload schema mismatch.

1. Create new quickApp
2. Add external endpoint for mcp toolset
3. Select Ouath authentication with login
4. Save changes

Image

quickapp should be saved successfully, and a 200 OK or 201 Created response should be returned

### What do you see instead?

The request fails with a 400 Bad Request with error message: "Payload does not match the required schema" 

Image

Image

Image

### Additional information

_No response_

DmytroZaichenkoDev:
CDN-fronted OAuth servers (e.g., Akamai) may return Content-Encoding: gzip regardless of the client's Accept-Encoding. Java's built-in HttpClient does not auto-decompress, so ResourceAuthorizationClient now reads the response as bytes and decodes gzip bodies when the header is present.

### Applicable issues

- fixes #1493

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
@@ -9,6 +9,7 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hc.core5.http.ContentType;
 
+import java.io.ByteArrayInputStream;
 import java.net.ConnectException;
 import java.net.ProxySelector;
 import java.net.URI;
@@ -20,6 +21,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.Map;
+import java.util.zip.GZIPInputStream;
 import javax.annotation.Nullable;
 
 @Slf4j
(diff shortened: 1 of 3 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -116,6 +118,23 @@ private static boolean hasUnresolvedAddressException(Throwable ex) {
...
+    // Some OAuth servers (e.g., CDN-fronted) return Content-Encoding: gzip even when the client
+    // did not request it. Java's built-in HttpClient does not auto-decompress, so we do it here.
+    @SneakyThrows
+    private static String decodeBody(HttpResponse<byte[]> response) {
+        byte[] body = response.body();
+        if (body == null || body.length == 0) {
+            return "";
+        }
+        String encoding = response.headers().firstValue("Content-Encoding").orElse("");
+        if ("gzip".equalsIgnoreCase(encoding)) {
```
astsiapanay: Content encoding may have multiple values separated by comma. Please see examples how we check content type in other controllers using Strings.CI.contains

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
```diff
@@ -116,6 +118,23 @@ private static boolean hasUnresolvedAddressException(Throwable ex) {
...
+            try (GZIPInputStream gzip = new GZIPInputStream(new ByteArrayInputStream(body))) {
```
astsiapanay, line 144: I think we should check in advance the encoding is gzip and add warning if encoding is not supported

ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthorizationClient.java
DmytroZaichenkoDev: fixed

astsiapanay (approved):

PR #1543, opened 2026-05-14
fix: "Invalid client" error for Snowflake-managed MCP #1527

### LINKED ISSUE #1527: "Invalid client" error when authenticating to Snowflake-managed mcp
### Name and Version

0.43.0

Core service fails to authenticate with Snowflake-managed mcp via auth toolset, returning a 400 Bad Request with an "invalid_client" error.

1. Navigate to the Toolset Configuration
2. Configure a connection to a Snowflake-managed mcp
3. Attempt to sign in/authenticate to the toolset
4. Observe the error in ui and core service logs

Core service should successfully exchange credentials for a token, similar to the successful Postman execution.

### What do you see instead?

The UI displays: "Failed to sign in to toolset."Core logs show a 400 Bad Request from Snowflake:
`{
  "data": null,
  "error": "invalid_client",
  "code": null,
  "message": "This is an invalid client.",
  "success": false
}`

### Additional information

_No response_

DmytroZaichenkoDev:
Adds OAuth 2.0 `token_endpoint_auth_method` support (`client_secret_post`, `client_secret_basic`, `none`) so DIAL can authenticate against authorization servers — like Snowflake-managed MCP — that don't accept credentials in the request body. The method is read from static config or the DCR response, and `TokenService` dispatches credential placement accordingly. Defaults to `client_secret_post` for backward compatibility.

### Applicable issues

- fixes #1527

### Description of changes

- New `TokenEndpointAuthMethod` enum with `resolveOrDefault` (null/blank → `client_secret_post`, unknown → `IllegalArgumentException`).
- `ResourceAuthSettings` and `ClientRegistration` gain a `tokenEndpointAuthMethod` field (String for JSON wire compatibility).
- `TokenService` dispatches credential placement per method:
  - `client_secret_post` — credentials in the form body (existing behavior).
  - `client_secret_basic` — `Authorization: Basic` header; client_id/client_secret URL-encoded before Base64 per RFC 6749 §2.3.1.
  - `none` — public client; only `client_id` is sent, no secret.
- `ResourceAuthorizationClient.executePost` accepts an optional extra-headers map so `TokenService` can pass the Authorization header for the Basic method.
- `DynamicResourceRegistrationStrategy` propagates `token_endpoint_auth_method` from the DCR response; `StaticResourceRegistrationStrategy` propagates it from operator config.
- `ResourceAuthSettingsService.preserveSensitiveFields` preserves the existing method when omitted on update, mirroring `clientSecret`.
- Tests: added `client_secret_basic`, `none`, and unsupported-method cases to `TokenServiceTest`; extended `ResourceAuthorizationClientTest` for the extra-headers contract; updated `ToolSetApiTest` mock.

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
@@ -74,6 +74,11 @@ public class ResourceAuthSettingsService {
                 resourceAuthSettings.setClientSecret(existingResourceAuthSettings.getClientSecret());
             }
 
+            // preserve tokenEndpointAuthMethod on updates that omit it
+            if (resourceAuthSettings.getTokenEndpointAuthMethod() == null) {
+                resourceAuthSettings.setTokenEndpointAuthMethod(existingResourceAuthSettings.getTokenEndpointAuthMethod());
+            }
+
             resolveCodeChallengeSettings(resourceAuthSettings, existingResourceAuthSettings);
         }
     }
@@ -128,6 +133,7 @@ public class ResourceAuthSettingsService {
         resourceAuthSettings.setTokenEndpoint(clientRegistration.getTokenEndpoint());
         resourceAuthSettings.setRedirectUri(clientRegistration.getRedirectUri());
         resourceAuthSettings.setScopesSupported(clientRegistration.getScopesSupported());
+        resourceAuthSettings.setTokenEndpointAuthMethod(clientRegistration.getTokenEndpointAuthMethod());
         setCodeChallengeProperties(resourceAuthSettings, clientRegistration.getCodeChallengeMethod());
     }
 
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
@@ -1,6 +1,7 @@
 package com.epam.aidial.core.credentials.service.registration;
 
 import com.epam.aidial.core.config.ResourceAuthSettings;
+import com.epam.aidial.core.credentials.data.credentials.TokenEndpointAuthMethod;
 import com.epam.aidial.core.credentials.data.registration.AuthorizationServerMetadata;
 import com.epam.aidial.core.credentials.data.registration.AuthorizationServerProtectedResourceMetadata;
 import com.epam.aidial.core.credentials.data.registration.ClientRegistration;
@@ -83,6 +84,12 @@ public class DynamicResourceRegistrationStrategy implements ResourceRegistration
                 ContentType.APPLICATION_JSON.toString(),
                 ClientRegistrationResponse.class);
 
+        // Reject methods DIAL can't honor (e.g. private_key_jwt) at registration time so the
+        // operator gets a clear error before any end-user sign-in attempt; null response falls
+        // back to the DIAL default so the persisted value always reflects what DIAL will use.
+        String tokenEndpointAuthMethod = TokenEndpointAuthMethod.resolveOrDefault(
+                clientRegistrationResponse.getTokenEndpointAuthMethod()).value();
+
         ClientRegistration clientRegistration = ClientRegistration.builder()
                 .resourceId(clientRegistrationResponse.getClientName())
                 .clientId(clientRegistrationResponse.getClientId())
@@ -91,7 +98,7 @@ public class DynamicResourceRegistrationStrategy implements ResourceRegistration
                 .authorizationEndpoint(authServerMetadata.getAuthorizationEndpoint())
                 .tokenEndpoint(authServerMetadata.getTokenEndpoint())
                 .scopesSupported(supportedScopes)
-                .tokenEndpointAuthMethod(clientRegistrationResponse.getTokenEndpointAuthMethod())
+                .tokenEndpointAuthMethod(tokenEndpointAuthMethod)
                 .build();
 
         Optional<String> codeChallengeMethod = getCodeChallengeMethod(authServerMetadata);
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
@@ -85,8 +85,9 @@ public class DynamicResourceRegistrationStrategy implements ResourceRegistration
                 ClientRegistrationResponse.class);
 
         // Reject methods DIAL can't honor (e.g. private_key_jwt) at registration time so the
-        // operator gets a clear error before any end-user sign-in attempt; null response falls
-        // back to the DIAL default so the persisted value always reflects what DIAL will use.
+        // operator gets a clear error before any end-user sign-in attempt. Null falls back to
+        // client_secret_basic per RFC 6749 §2.3.1; the persisted value always reflects what
+        // DIAL will use.
         String tokenEndpointAuthMethod = TokenEndpointAuthMethod.resolveOrDefault(
                 clientRegistrationResponse.getTokenEndpointAuthMethod()).value();
 
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
@@ -91,6 +91,7 @@ public class DynamicResourceRegistrationStrategy implements ResourceRegistration
                 .authorizationEndpoint(authServerMetadata.getAuthorizationEndpoint())
                 .tokenEndpoint(authServerMetadata.getTokenEndpoint())
                 .scopesSupported(supportedScopes)
+                .tokenEndpointAuthMethod(clientRegistrationResponse.getTokenEndpointAuthMethod())
                 .build();
 
         Optional<String> codeChallengeMethod = getCodeChallengeMethod(authServerMetadata);
--- a/config/src/main/java/com/epam/aidial/core/config/ResourceAuthSettings.java
+++ b/config/src/main/java/com/epam/aidial/core/config/ResourceAuthSettings.java
@@ -68,4 +68,7 @@ public class ResourceAuthSettings {
 
     @JsonAlias({"scopesSupported", "scopes_supported"})
     private List<String> scopesSupported;
+
+    @JsonAlias({"tokenEndpointAuthMethod", "token_endpoint_auth_method"})
+    private String tokenEndpointAuthMethod;
 }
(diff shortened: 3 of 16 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
```diff
@@ -91,6 +91,7 @@ public ClientRegistration register(String resourceId, String resourceEndpoint, R
                 .authorizationEndpoint(authServerMetadata.getAuthorizationEndpoint())
                 .tokenEndpoint(authServerMetadata.getTokenEndpoint())
                 .scopesSupported(supportedScopes)
+                .tokenEndpointAuthMethod(clientRegistrationResponse.getTokenEndpointAuthMethod())
```
astsiapanay: ### C2 + C4 — DCR response with unknown method silently poisons the toolset, and the propagation has no test

**C2 (production behaviour):** This line writes whatever `tokenEndpointAuthMethod` the AS returns verbatim into the persisted `ClientRegistration` → `ResourceAuthSettings`. RFC 7591 §2 permits the server to return `private_key_jwt`, `tls_client_auth`, `client_secret_jwt`, `self_signed_tls_client_auth`, or anything in its `token_endpoint_auth_methods_supported`. Real auth servers (Microsoft Entra DCR, Google, etc.) often pick the strongest method they support, not `client_secret_post`.

With the current code: registration returns 200, settings persist, and the operator gets no signal until the first end-user clicks sign-in. Then `TokenEndpointAuthMethod.resolveOrDefault(...)` throws `IllegalArgumentException`, which on the refresh path collapses to a generic 500 (see C1 in the review summary). The toolset is effectively bricked between registration and first failure — possibly weeks later.

Fix options (any one is enough):
1. **Validate immediately:** `TokenEndpointAuthMethod.resolveOrDefault(clientRegistrationResponse.getTokenEndpointAuthMethod()).value()` so an unsupported method fails registration with a clear 400. Operator learns at config time.
2. **Request DIAL's supported methods** in `ClientRegistrationRequest` so the AS picks from a known-good set.
3. **Fail-safe to null** on unsupported values so `resolveOrDefault` later applies the legacy `client_secret_post` default — gentler, trades silent-config-loss for silent-fallback.

**C4 (test gap):** This line has no unit test. `DynamicResourceRegistrationStrategyTest.java` has zero hits for `tokenEndpointAuthMethod`. Delete this builder line and no test fails — even though this is the actual Snowflake user flow the PR exists to fix. Please add:

- `testRegister_propagatesTokenEndpointAuthMethodFromResponse` — mock DCR response with `token_endpoint_auth_method=client_secret_basic`, assert the built `ClientRegistration` carries it.
- `testRegister_nullTokenEndpointAuthMethodInResponse` — server omits the field → `ClientRegistration` has `null` (so `resolveOrDefault` later applies legacy default).
- Ideally a `ToolSetApiTest` end-to-end variant that mocks `/.well-known/oauth-authorization-server` + `/register` + `/token` to verify the full Snowflake-shaped DCR flow.

The parallel propagation in `StaticResourceRegistrationStrategy.java:103` is also untested — please add a matching `testRegister_propagatesTokenEndpointAuthMethod_*` series there too.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
```diff
@@ -74,6 +74,11 @@ public void processResourceAuthSettings(SecuredResource updatedResource,
                 resourceAuthSettings.setClientSecret(existingResourceAuthSettings.getClientSecret());
             }
 
+            // preserve tokenEndpointAuthMethod on updates that omit it
+            if (resourceAuthSettings.getTokenEndpointAuthMethod() == null) {
```
astsiapanay, line 78: ### C3 — preserve-on-update has no test, and would silently downgrade `client_secret_basic` toolsets on PATCH-style updates

This preserve block mirrors the existing `clientSecret` preservation above — but unlike that path, **there is no test asserting this behaviour**. `ResourceAuthSettingsServiceTest.java` was not modified in this PR; zero hits for `tokenEndpointAuthMethod`. The existing `Update_*` tests assert nothing about this field.

**Regression this gap permits:** if a future refactor removes or reorders this block (e.g., copies fields before the null-check), an admin doing a PATCH-style PUT that omits `token_endpoint_auth_method` will silently downgrade a previously-`client_secret_basic`-configured toolset to `client_secret_post` — credentials get sent in the form body to a strict-Basic auth server, which rejects with `invalid_client`. **This is exactly the Snowflake regression the PR exists to fix**, and the new integration test does *not* catch it because both `ToolSetApiTest` flows PUT a *fresh* resource (no prior version with a method to preserve).

Please add at minimum:

- `testProcessResourceAuthSettings_OauthUpdate_TokenEndpointAuthMethodNull_PreservesExisting` — existing toolset has `tokenEndpointAuthMethod="client_secret_basic"`, PUT body sets it to `null`, after `processResourceAuthSettings` the value is still `"client_secret_basic"`.
- `testProcessResourceAuthSettings_OauthUpdate_TokenEndpointAuthMethodChanged_Overrides` — explicit transition basic → none should replace, not preserve.
- An integration variant in `ToolSetApiTest`: PUT with `client_secret_basic`, then PUT again with the field omitted, GET, assert `token_endpoint_auth_method` is still `"client_secret_basic"`.

Separately (not part of C3, just adjacent — flagging because it's load-bearing on the same line): the `== null` check treats `""`/`"   "` differently from `null`. A UI doing a partial update can easily emit `"token_endpoint_auth_method": ""`; the null-check fails to preserve, blank persists, and `resolveOrDefault("")` silently returns `CLIENT_SECRET_POST` per its `isBlank()` early-return. Consider `StringUtils.isBlank(...)` to match the resolver's own contract.

astsiapanay (commented):
## Critical issues from a multi-agent review

Reviewing this PR with five specialist agents (code-review, tests, errors, types, comments) surfaced 4 critical issues that I'd like to flag before merge. Three of them have inline anchors below; one (C1) anchors to a file outside this PR's diff so I'm putting it here.

Happy to discuss any of these — the goal isn't to gate the PR, just to make sure these don't slip through.

---

### C1 — `IllegalArgumentException` from `resolveOrDefault` is swallowed on the tools-listing path

**File:** `server/src/main/java/com/epam/aidial/core/server/controller/ToolSetToolsController.java:169-175` *(not in this PR's diff, so flagged here rather than inline)*

`fetchTools()`'s `onFailure` collapses any throwable from the refresh chain (`injectToolsetCredentials` → `getRefreshedResourceCredentials` → `tokenService.getToken` → `applyClientAuthentication` → `TokenEndpointAuthMethod.resolveOrDefault`) into a generic 500 with body `"Failed to handle request to MCP server"`. The actionable `"Unsupported token_endpoint_auth_method: <value>"` message is logged at WARN but never reaches the operator/UI.

The sibling `McpProxyController.handleError` already maps `IllegalArgumentException` → 400 with `error.getMessage()` — the two MCP controllers treat the same exception class differently. Fix by unifying the error mapping; this also benefits `resolveRedirectUri`'s `IllegalArgumentException` which has the same problem today.

**Why this matters first:** without C1, every other unknown-method failure mode in this PR (C2 below, plus several Important-level issues) is undebuggable in production — operators see a generic 500 and have to read DIAL's logs to find the actual cause.

ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/registration/DynamicResourceRegistrationStrategy.java
DmytroZaichenkoDev: **C2 — addressed** in 8d8a49ac (rebased): `DynamicResourceRegistrationStrategy` now calls `TokenEndpointAuthMethod.resolveOrDefault(...)` on the DCR response. Unsupported methods (`private_key_jwt`, etc.) throw `IllegalArgumentException` at registration time so the operator sees a clear error at toolset save.

Followup in 01c1cadb: same validation on the static-creation/update path via `OauthAuthSettingsValidator.parse(...)` so user-supplied values are also rejected at save, not at first sign-in. The null default was also flipped to `client_secret_basic` per RFC 6749 §2.3.1 (BASIC is MUST-support; POST is NOT RECOMMENDED) and OAuth 2.1's deprecation of POST.

**C4 — addressed** in 8d8a49ac:
- `DynamicResourceRegistrationStrategyTest.testRegister_propagatesTokenEndpointAuthMethodFromResponse`
- `DynamicResourceRegistrationStrategyTest.testRegister_nullTokenEndpointAuthMethodInResponse_resolvesToSpecDefault`
- `DynamicResourceRegistrationStrategyTest.testRegister_unsupportedTokenEndpointAuthMethodInResponse_fails`
- Matching trio in `StaticResourceRegistrationStrategyTest`
- End-to-end coverage via `ToolSetApiTest.testOauthUpdate_PreservesTokenEndpointAuthMethodWhenOmitted`

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceAuthSettingsService.java
DmytroZaichenkoDev: **C3 — addressed** in 8d8a49ac:
- `ResourceAuthSettingsServiceTest.testProcessResourceAuthSettings_OauthUpdate_TokenEndpointAuthMethodNull_PreservesExisting`
- `ResourceAuthSettingsServiceTest.testProcessResourceAuthSettings_OauthUpdate_TokenEndpointAuthMethodChanged_Overrides`
- Integration variant in `ToolSetApiTest.testOauthUpdate_PreservesTokenEndpointAuthMethodWhenOmitted` (PUT with basic → PUT without → GET asserts basic still set)

**`""` / `"   "` edge case — handled at a different layer**, not by `StringUtils.isBlank` in the preserve block. `OauthAuthSettingsValidator.validate()` (added in 01c1cadb) runs before the preserve block and calls `TokenEndpointAuthMethod.parse(...)` on any non-null value — `""` and `"   "` throw `Unsupported token_endpoint_auth_method: ...` with a 400, so the malformed value never reaches the preserve block. Strict over lenient: a UI sending `""` has a serialization bug, and a silent preserve would mask it. Test cases in baba7c5e pin this contract.

If you'd still prefer the lenient route (treat blank → null → preserve), happy to add it, but the strict path felt closer to "fail fast at the API boundary."

DmytroZaichenkoDev:
Review summary — status:

- **C1** (`IllegalArgumentException` swallowed by `ToolSetToolsController.fetchTools()` → generic 502): **deferred to a separate fix.** The exception-mapping unification belongs with the sibling `McpProxyController.handleError` path and touches code outside this PR's scope.
- **C2** (DCR response with unsupported method silently poisons toolset): **addressed** in 8d8a49ac. Replied inline.
- **C3** (preserve-on-update test gap, plus `""`/`"   "` edge case): **tests addressed** in 8d8a49ac; **blank-string edge case handled at the validator layer** (01c1cadb + baba7c5e), not the preserve block. Replied inline.
- **C4** (DCR/static propagation untested): **addressed** in 8d8a49ac. Replied inline.

Bonus work in this round (not flagged by the review but tangentially relevant): null default flipped from `client_secret_post` to `client_secret_basic` per RFC 6749 §2.3.1, with `OauthAuthSettingsValidator` rejecting unsupported values at save time so failures surface at the API boundary instead of at first sign-in.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1566, opened 2026-05-22
fix: expose application route upstreams and toolset endpoint securely #1565

### LINKED ISSUE #1565: Some fields should be exposed securely for applications and toolsets
### Name and Version

core:development

## Open Application API
Endpoints `^/+openai/applications/(?<id>.+?)$` and `^/+openai/applications$`
Application route upstreams should be exposed if a requester has write permission.

## Resource API
Endpoint `^/v1/toolsets/(?<bucket>[a-zA-Z0-9]+)/(?<path>.*)$`

Toolset endpoint should be exposed if a requester has write permission.

_No response_

### What do you see instead?

The fields are exposed for any user

### Additional information

_No response_

astsiapanay:
Sensitive fields were exposed to all users regardless of permission level. Application route `upstreams` and toolset `endpoint` are now only visible to users with write access.

### Applicable issues

- fixes #1565

### Description of changes

- `ApplicationController`: added `hasWriteAccess(Application)` to determine write access for both config-based apps (admin check) and custom apps (resource descriptor write-access check); `mapApplication()` now strips `upstreams` from each route when the caller lacks write access
- `ResourceController`: `getToolsetData()` now nulls out `toolSet.endpoint` when `hasWriteAccess` is false
- Updated existing test assertions to reflect correct secure behavior; added `testToolsetEndpointHiddenForReadOnlyUser` and `testRouteUpstreamsHiddenForReadOnlyUser` tests

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
@@ -197,16 +197,40 @@ public class ApplicationController {
     }
 
     private void respondError(Throwable error) {
-        if (error instanceof IllegalArgumentException) {
-            context.respond(HttpStatus.BAD_REQUEST, error.getMessage());
-        } else if (error instanceof PermissionDeniedException) {
-            context.respond(HttpStatus.FORBIDDEN, error.getMessage());
-        } else if (error instanceof ResourceNotFoundException) {
-            context.respond(HttpStatus.NOT_FOUND, error.getMessage());
-        } else {
-            context.respond(error, "Internal error");
-            log.error("Failed to handle application request", error);
+        switch (error) {
+            case IllegalArgumentException ignored ->
+                    context.respond(HttpStatus.BAD_REQUEST, error.getMessage());
+            case PermissionDeniedException ignored ->
+                    context.respond(HttpStatus.FORBIDDEN, error.getMessage());
+            case ResourceNotFoundException ignored ->
+                    context.respond(HttpStatus.NOT_FOUND, error.getMessage());
+            case null, default -> {
+                context.respond(error, "Internal error");
+                log.error("Failed to handle application request", error);
+            }
         }
+    }
+
+    private boolean hasWriteAccess(Application application) {
+        String appName = application.getName();
+        if (appName != null && appName.equals(context.getDecodedSourceDeployment())) {
+            return true;
+        }
+        if (appName != null) {
+            try {
+                ResourceDescriptor resource = ResourceDescriptorFactory.fromAnyUrl(appName, encryptionService);
+                if (resource.getType() == ResourceTypes.APPLICATION) {
+                    return accessService.hasWriteAccess(resource, context);
+                }
+            } catch (Exception e) {
+                // appName is a config-based deployment name, fall through to admin check
+            }
+        }
+        return accessService.hasAdminAccess(context);
+    }
+
+    private static void clearUpstreams(Map<String, Route> routes) {
+        routes.forEach((name, route) -> route.setUpstreams(null));
     }
 
     private ApplicationData mapApplication(Application application) {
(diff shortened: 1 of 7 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
```diff
@@ -197,18 +197,42 @@ private void checkAccess(ResourceDescriptor resource) {
...
+        if (appName != null && appName.equals(context.getDecodedSourceDeployment())) {
+            return true;
+        }
+        if (appName != null) {
+            try {
+                ResourceDescriptor resource = ResourceDescriptorFactory.fromAnyUrl(appName, encryptionService);
+                if (resource.getType() == ResourceTypes.APPLICATION) {
+                    return accessService.hasWriteAccess(resource, context);
+                }
+            } catch (Exception e) {
```
Oleksii-Klimov, line 225: How is it expected to fail?

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
```diff
@@ -197,18 +197,42 @@ private void checkAccess(ResourceDescriptor resource) {
...
+                // appName is a config-based deployment name, fall through to admin check
+            }
+        }
+        return accessService.hasAdminAccess(context);
+    }
+
+    private static void clearUpstreams(Map<String, Route> routes) {
+        routes.forEach((name, route) -> route.setUpstreams(null));
```
Oleksii-Klimov, line 233: Can the upstream list be empty instead?

astsiapanay: The empty list is a wrong configuration. The core doesn't know where to route a request to

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
astsiapanay: The comment explains the case. See the code changes
```
// appName is a config-based deployment name, fall through to admin check
```

Oleksii-Klimov: I didn’t understand the comment initially. The message itself needs an explanation. I think it’s worth pointing out that the app name is not a URL, so this check is not applicable. However, I think it would be better to validate the URL explicitly instead of relying on a generic catch block to catch an obscure error.

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
Oleksii-Klimov: I wonder why `null` is considered an acceptable configuration, then. It’s usually not good practice to distinguish between `null` and an empty collection. I’m just curious why that was needed here.

astsiapanay: We don't want to serialize nullable values in a response. The field has to be non empty and non-nullable list anyway. The intention is not expose the field when user doesn't have access to.

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
astsiapanay: updated the comment

server/src/main/java/com/epam/aidial/core/server/controller/ApplicationController.java
astsiapanay: fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [26440635372](https://github.com/epam/ai-dial-ci/actions/runs/26440635372)

| [deploy-review](https://chat-ai-dial-core-pr-1566.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26440635372/job/77833914332) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26440635372/job/77835971555) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1568, opened 2026-05-22
feat: support user defined upstream endpoints #1567

### LINKED ISSUE #1567: Support user defined upstream endpoints
### Name and Version

core:development

Add ability to support user defined upstream endpoints for chat completion and responses API requests. At the moment Core does load balancing based on upstreams provided in config. So user is not be able to override the behavior by sticking upstream where a request should be routed to.

DIAL Core should accept HTTP header `X-UPSTREAM-ID` from request:
1. Find upstream from a list of upstreams for the given deployment.
2. If the upstream is found then pass it to UpstreamProvider as a list of a single upstream.
3. Otherwise return HTTP error code 400 with the error message in the body: Unknown upstream id {upstream.id}.

The scope is chat completion  and responses API requests.

_No response_

astsiapanay:
Adds support for an `X-UPSTREAM-ID` request header that lets clients pin a single upstream for chat completion and Responses API requests. If the id matches a configured upstream for the deployment it is the only upstream used; otherwise Core responds 400 `Unknown upstream id {id}`.

### Applicable issues

- fixes #1567

### Description of changes

- Added optional `id` field on `Upstream` (config), JSON-aliased as `id` / `dial:id`.
- Introduced `X-UPSTREAM-ID` header constant in `Proxy`.
- New `UpstreamRouteProvider.get(deployment, breakpointContext, upstreamId)` overload that narrows the upstream list to the single match and throws `HttpException(400, "Unknown upstream id <id>")` on miss. Blank/null id is ignored (full load balancing).
- Wired the header through `DeploymentPostController` (chat completions) and `ResponsesController` (responses API); each translates the 400 to a response via its existing error path.
- Tests for the new provider logic; existing test call sites and the `ResponsesControllerTest` deep-stub updated for the new constructor / overload.

### Checklist

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/Upstream.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Upstream.java
@@ -31,4 +31,7 @@ public class Upstream {
     private int weight = 1;
     @JsonAlias({"tier", "dial:tier"})
     private int tier = 0;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonAlias({"id", "dial:id"})
+    private String id;
 }
--- a/server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
+++ b/server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
@@ -75,7 +75,8 @@ public class UpstreamRouteProvider {
         if (upstreamId != null && !upstreamId.isBlank()) {
             Upstream selected = null;
             for (Upstream upstream : upstreams) {
-                if (upstreamId.equals(upstream.getId())) {
+                String id = Objects.requireNonNullElse(upstream.getId(), upstream.getEndpoint());
+                if (upstreamId.equals(id)) {
                     selected = upstream;
                     break;
                 }
--- a/server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
+++ b/server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
@@ -10,6 +10,8 @@ import com.epam.aidial.core.server.data.cache.CacheBreakpointContext;
 import com.epam.aidial.core.server.data.cache.CachedUpstreamEntry;
 import com.epam.aidial.core.server.service.UpstreamCacheService;
 import com.epam.aidial.core.server.vertx.AsyncTaskExecutor;
+import com.epam.aidial.core.storage.http.HttpException;
+import com.epam.aidial.core.storage.http.HttpStatus;
 import io.vertx.core.Vertx;
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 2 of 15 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
```diff
@@ -54,13 +56,36 @@ public UpstreamRouteProvider(Vertx vertx, AsyncTaskExecutor taskExecutor, Suppli
...
+                if (upstreamId.equals(upstream.getId())) {
+                    selected = upstream;
+                    break;
+                }
+            }
+            if (selected == null) {
+                throw new HttpException(HttpStatus.BAD_REQUEST, "Unknown upstream id " + upstreamId);
+            }
+            upstreams = List.of(selected);
+            key = key + ":upstream:" + upstreamId;
```
adubovik, line 88: Why do we need to modify the `key`?

astsiapanay: We don't want to pollute the state of deployments for a normal balancing of upstreams. So we treat is a specific use rather than a regular one.

adubovik (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/Upstream.java
```diff
@@ -31,4 +31,7 @@ public class Upstream {
     private int weight = 1;
     @JsonAlias({"tier", "dial:tier"})
     private int tier = 0;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonAlias({"id", "dial:id"})
+    private String id;
```
Oleksii-Klimov, line 36: Should this go to the CachedUpstreamEntry?

astsiapanay: Yes, it should. Moreover we should revisit the logic of selecting upstream from a list of the deployment.
As of now it relies on endpoint only. However some upstream may miss the field. So the Core should switch on id and as a fallback to endpoint for backward compatibility.

Oleksii-Klimov: In this PR?

ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/Upstream.java
astsiapanay: yes, it's

Oleksii-Klimov (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [26440602562](https://github.com/epam/ai-dial-ci/actions/runs/26440602562)

| [deploy-review](https://chat-ai-dial-core-pr-1568.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26440602562/job/77833803197) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26440602562/job/77835854048) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

adubovik (approved):

PR #1609, opened 2026-06-10
fix: send client_id in token request body for OAuth toolsets #1608

### LINKED ISSUE #1608: Login to Toolset fails with "invalid_grant" (OAuth client_id mismatch)
### Name and Version

0.44.2 

Logging in to Toolset fails with an OAuth invalid_grant error on core version 0.44.2. The error indicates a client_id mismatch between the initial authorization request and the token exchange. Reverting to core 0.42.3 resolves the issue.

**Steps to reproduce**
1. Run core version 0.44.2
2. Attempt to log in to Toolset using the MCP FastMCP OAuth proxy flow
   (ref: https://gofastmcp.com/servers/auth/oauth-proxy#oauth-flow)
3. Observe the error response above

Login completes successfully, as it does on core 0.42.3.

### What do you see instead?

Authorization fails at the token exchange step with invalid_grant — the server returns a client_id mismatch error, suggesting the client_id used in the initial /authorize request differs from the one sent in the /token request.

Error message
{"error": "invalid_grant", "error_description": "Client ID does not match the one used in the initial request."}

### Additional information

- Affected version: core 0.44.2
- Working version: core 0.42.3
- Auth method: MCP FastMCP OAuth proxy

DmytroZaichenkoDev:
## Summary

OAuth toolsets whose token endpoint is fronted by a **FastMCP OAuth Proxy** (or any MCP server built on the MCP Python SDK) fail sign-in with:

```
invalid_grant: Client ID does not match the one used in the initial request
```

Root cause: in `client_secret_basic` mode DIAL sent `client_id` **only** in the `Authorization: Basic` header and omitted it from the request body. The MCP SDK's token endpoint reads `client_id` from the form body unconditionally (`form_data.get("client_id")`) and matches the client by that value; with no body `client_id` it cannot bind the code to the client → `invalid_grant`. (Worked on 0.42.3, regressed in 0.44.)

## Fix

`TokenService.applyClientAuthentication` now also sets `client_id` in the body for the `CLIENT_SECRET_BASIC` branch (RFC 6749 §3.2.1 permits `client_id` in the body alongside the Basic header). The Basic header is unchanged, so servers that authenticate via the header (e.g. Snowflake) are unaffected.

Additionally, `TokenEndpointAuthMethod.resolve(value, clientSecret)` infers the method when the server omits it: no secret → `none` (public/PKCE client), secret → `client_secret_basic` (RFC 7591 §2 default). A secretless client defaulting to `none` keeps `client_id` in the body with no header — which public/PKCE DCR servers like FastMCP require.

## Verification

- Updated unit tests (`TokenServiceTest`) and integration tests (`ToolSetApiTest`) assert `client_id` is present in the body for both `client_secret_basic` and `none` modes.
- Reproduced end-to-end against a local FastMCP OAuth Proxy: prior behavior → `invalid_grant`; with the fix → `/token` returns 200 and credentials are stored.
- Empirically confirmed the proxy validates on body `client_id` + PKCE and ignores the secret/header — matching the client's failure signature.

Fixes #1608

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/data/credentials/TokenEndpointAuthMethod.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/data/credentials/TokenEndpointAuthMethod.java
@@ -4,16 +4,9 @@ package com.epam.aidial.core.credentials.data.credentials;
  * Client authentication methods at the OAuth 2.0 token endpoint
  * (RFC 6749 §2.3 and RFC 8414 §2 — {@code token_endpoint_auth_methods_supported}).
  *
- * <p>A null/blank value resolves to {@link #CLIENT_SECRET_BASIC} per RFC 6749 §2.3.1
- * (BASIC is {@code MUST} for compliant authorization servers; POST is {@code MAY} and
- * {@code NOT RECOMMENDED}). OAuth 2.1, which MCP references, deprecates POST entirely.
- *
- * <p>DIAL historically sent {@code client_secret_post} unconditionally. Operators whose AS
- * only accepts POST must set this field explicitly — those servers are non-compliant with
- * RFC 6749 §2.3.1 and would be unable to interop with most OAuth 2.1 clients anyway.
- *
- * <p>TODO: require the field at creation time so the default applies only to pre-PR
- * persisted data and can eventually be removed.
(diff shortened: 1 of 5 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1613, opened 2026-06-10
feat: track usage of background jobs

Oleksii-Klimov:
Adds usage accounting for Responses API background jobs.

Background jobs are persisted in Redis, indexed for restart recovery, and polled with exponential backoff until a terminal Responses API result is available. On completion, Core records model token usage against the original initiator bucket, ends the root trace span, and invalidates the per-request key.

Updates response handling to save and poll non-streaming background jobs, finalize streamed jobs without double-counting usage, and return `409 Conflict` when deleting a response with an active background job.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov:
/deploy-review

>GitHub actions run: [27756911082](https://github.com/epam/ai-dial-ci/actions/runs/27756911082)

| [deploy-review](https://chat-ai-dial-core-pr-1613.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/27756911082/job/82120957817) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/27756911082/job/82122771022) |

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
```diff
@@ -150,16 +151,19 @@ protected Future<TokenUsage> collectTokenUsage(Buffer responseBody) {
...
                 }
                 context.setTokenUsage(tokenUsage);
-                tokenUsageFuture = proxy.getRateLimiter().increase(context, context.getDeployment())
-                        .transform(result -> {
-                            if (result.failed()) {
-                                log.warn("Failed to increase limit", result.cause());
-                            }
-                            return proxy.getTokenStatsTracker().updateModelStats(context);
-                        });
+                tokenUsageFuture = proxy.getTokenStatsTracker().collectUsage(
```
astsiapanay: I don't think that's a good idea to inject rate limiter into token stats tracker. Since they do different jobs. Low coupling and tight cohesion are good reasons to make them separate

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
```diff
@@ -131,8 +132,8 @@ protected void handleRequestBodyError(Throwable error) {
         log.warn("Failed to receive client body. Error: {}", error.getMessage());
     }
 
-    protected Future<TokenUsage> collectTokenUsage(Buffer responseBody) {
-        Future<TokenUsage> tokenUsageFuture = Future.succeededFuture();
+    protected Future<Void> collectTokenUsage(Buffer responseBody) {
```
astsiapanay, line 162: Why did you change result type?

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
```diff
@@ -74,70 +85,80 @@ private Future<Void> forwardToUpstream(ResponseMapping mapping) {
...
+                .compose(body -> {
+                    if (proxyResponse.statusCode() != 200) {
+                        return sendResponse(proxyResponse, body);
                     }
-                    context.getResponse().putHeader(HttpHeaders.CONTENT_LENGTH, Integer.toString(rewritten.length()));
-                    return context.getResponse().end(rewritten);
-                })
-                .mapEmpty();
+                    return proxy.getTaskExecutor()
+                            .submit(() -> rewriteId(body, mapping.getUpstreamResponseId()))
```
astsiapanay, line 194: Nit. rewriteId is not blocking operation and could be executed in event loop thread.

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
```diff
@@ -166,27 +187,27 @@ private Future<Void> collectAndForwardStreaming(HttpClientResponse proxyResponse
     private static HttpException notFoundException() {
         return new HttpException(
                 HttpStatus.NOT_FOUND,
-                "{\"error\":{\"message\":\"Unknown or expired response_id\","
-                        + "\"type\":\"invalid_request_error\",\"param\":\"response_id\",\"code\":null}}");
+                """
+                        {
+                            "error": {
+                                "message": "Response with id 'resp_123' not found.",
```
astsiapanay: Please replace a constant with a real response id

astsiapanay: You may use DIAl error template class instead

server/src/main/java/com/epam/aidial/core/server/controller/ResponsesController.java
```diff
@@ -285,7 +289,7 @@ private Future<Buffer> rewriteResponseId(HttpClientResponse proxyResponse, Buffe
         JsonNode tree = JsonUtil.tryParse(body.getBytes());
         if (tree.isObject() && tree instanceof ObjectNode object) {
             JsonNode idNode = object.path("id");
-            if (!idNode.isNull()) {
+            if (idNode.isTextual()) {
```
astsiapanay: Why do we need this check? Eventually we may present value as text
Perhaps we want to log if there is no value.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobPoller.java
```diff
@@ -0,0 +1,62 @@
...
+import com.epam.aidial.core.server.upstream.UpstreamRouteProvider;
+import com.epam.aidial.core.server.util.ProxyUtil;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.vertx.core.Future;
+import io.vertx.core.http.HttpMethod;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class BackgroundJobPoller {
```
astsiapanay: This class is highly specialized and being used by BackgroundJobService only. Can we move the method there?

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,303 @@
...
+        this.tokenStatsTracker = tokenStatsTracker;
+        this.responseMappingService = responseMappingService;
+        this.poller = poller;
+    }
+
+    public void init() {
+        resumeActiveJobs();
+    }
+
+    public Future<Void> saveJob(ProxyContext context) {
```
astsiapanay: We don't need entire context here. Just perRequestKey and jobId

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,303 @@
...
+                        return Future.succeededFuture();
+                    }
+                    return processResult(dialResponseId, record, mapping, usage);
+                });
+    }
+
+    public void startPolling(String id) {
+        loadRecordAsync(id)
+                .compose(record -> {
+                    if (record == null) {
```
astsiapanay: Is the state ever possible?

server/src/main/java/com/epam/aidial/core/server/controller/ResponsesController.java
```diff
@@ -246,36 +257,29 @@ private void handleProxyResponse(HttpClientResponse proxyResponse) {
...
-                })
-                .onComplete(future -> {
-                    if (future.failed()) {
-                        log.warn("Failed to collect attachments from response", future.cause());
-                    }
-                    completeProxyResponse();
-                })
-                .mapEmpty();
+    private void finishNonStreamingResponse() {
+        if (context.isBackgroundJob() && context.getResponseId() != null) {
```
astsiapanay: That change leads to major side effects:

1. API keys won't be invalidated but they will be dropped by Redis eviction process.
2. Token stats are not updated correctly if application calls background job. In this case DIAL Core remove trace when a request initiator receives a response from the app.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,303 @@
...
+                                    if (record == null) {
+                                        log.info("Background job {} record not found, stopping polling", id);
+                                        return Future.succeededFuture();
+                                    }
+                                    return poller.poll(mapping)
+                                            .compose(
+                                                    result -> {
+                                                        if (result == null) {
+                                                            failureCount.set(0);
+                                                            scheduleNextPoll(id, mapping, failureCount, nextIntervalMs);
```
astsiapanay: Please investigate what there is no memory consumption on every schedule because of parent context is included

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,303 @@
...
+                        log.info("Background job {} record already deleted, skipping completion processing", jobId);
+                    }
+                    return Future.succeededFuture();
+                });
+    }
+
+    private void scheduleNextPoll(String id, ResponseMapping mapping, AtomicInteger failureCount, long currentIntervalMs) {
+        vertx.setTimer(currentIntervalMs, ignored -> {
+            long nextIntervalMs = Math.min(currentIntervalMs * 2, maxPollIntervalMs);
+            vertx.executeBlocking(() -> lockService.tryLock("background_job_poll:" + id))
```
astsiapanay: Replace to AsyncTaskExecutor

server/src/main/java/com/epam/aidial/core/server/service/ResponsesApiClient.java
```diff
@@ -0,0 +1,57 @@
...
+                        .putHeader(Proxy.HEADER_UPSTREAM_ENDPOINT, upstream.getResponsesEndpoint())
+                        .putHeader(Proxy.HEADER_UPSTREAM_EXTRA_DATA, UpstreamExtraDataMerger.merge(upstream))
+                        .send());
+    }
+
+    private static boolean isTerminal(String status) {
+        return !("queued".equals(status) || "in_progress".equals(status));
+    }
+
+    public static TerminalResult parseTerminalBody(ObjectNode tree) {
```
astsiapanay: annotate with @Nullable

server/src/main/java/com/epam/aidial/core/server/util/BucketBuilder.java
```diff
@@ -33,15 +34,22 @@ public String buildAppDataBucket(ProxyContext context) {
     }
 
     public static String buildInitiatorBucket(ProxyContext context) {
-        String userId = context.getUserId();
-        String apiKeyId = context.getProject();
+        return buildInitiatorBucket(context.getUserId(), context.getProject());
+    }
+
+    public static String buildInitiatorBucket(ApiKeyData apiKeyData) {
+        String userId = apiKeyData.getExtractedClaims() != null ? apiKeyData.getExtractedClaims().userId() : null;
```
astsiapanay: Why? 

`context.getUserId()` returns userId or null

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
Oleksii-Klimov: It's just not used anywhere.

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
Oleksii-Klimov: Ah, right. I copied the response as is without updating it. I'll fix it.

server/src/main/java/com/epam/aidial/core/server/util/BucketBuilder.java
Oleksii-Klimov: Removed the method.

server/src/main/java/com/epam/aidial/core/server/service/ResponsesApiClient.java
Oleksii-Klimov: Added.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Done.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Refactored.

server/src/main/java/com/epam/aidial/core/server/controller/ResponsesController.java
Oleksii-Klimov: 1. Extended TTL for background jobs. 
2. As discussed, this is a separate issue that should also be fixed for the Chat Completions API.

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
Oleksii-Klimov: It may not be ideal to parse arbitrary JSON in an event loop thread. Additionally, it handles exceptions for me. However, I can update it if you want.

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
Oleksii-Klimov: Split.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java
```diff
@@ -389,14 +397,101 @@ private void handleResponseError(Throwable error, BufferingReadStream responseSt
...
 
+    /**
+     * Assembles streaming response into a single one.
+     * The assembling process merges chunks of the streaming response one by one using separator: <code>\n*data: *</code>
+     *
+     * @param response byte array response to be assembled.
+     * @return assembled streaming response
+     */
+    @Nullable
+    static String assembleStreamingResponse(@Nullable Buffer response) {
```
Oleksii-Klimov: Moved from GfLogStore

server/src/main/java/com/epam/aidial/core/server/controller/ResponsesController.java
Oleksii-Klimov: Added logging.

server/src/main/java/com/epam/aidial/core/server/log/AnalyticsLogContext.java
```diff
@@ -0,0 +1,124 @@
...
+                .requestTimestamp(record.requestTimestamp())
+                .requestBody(Buffer.buffer(record.requestBody()))
+                .upstreamEndpoint(record.upstreamEndpoint())
+                .responseStatusCode(result == null ? 500 : 200) // 500 when expired
+                .responseBody(result == null ? null : result.body())
+                .tokenUsage(result == null ? null : result.usage())
+                .build();
+    }
+
+    public static String getParentDeployment(String sourceDeployment, List<String> interceptors, List<String> executionPath) {
```
Oleksii-Klimov, line 137: Moved from GfLogStore

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobPoller.java
Oleksii-Klimov: This was done to simplify testing. Moved.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -235,7 +236,8 @@ public class ControllerSelector {
         });
         post(RouteTemplate.LLM_RESPONSES_API_CANCEL, (proxy, context, pathMatcher) -> {
             String id = UrlUtil.decodePath(pathMatcher.group("id"));
-            return new ResponseItemController(proxy, context, id, ResponseItemController.Operation.CANCEL);
+            context.setDialResponseId(id);
```
astsiapanay: why do we need to put response ID to context here?

server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java
```diff
@@ -337,7 +342,10 @@ private void completeProxyResponse(BufferingReadStream responseStream) {
         HttpServerResponse response = context.getResponse();
         responseStream.end(response);
 
-        proxy.getLogStore().save(context);
+        if (isEventStreamResponse(context.getProxyResponse())) {
+            context.setAssembledStreamingResponse(assembleStreamingResponse(context.getResponseBody()));
```
astsiapanay: I support assembling streaming response should be a part of building log context.

server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java
astsiapanay: Please move to LogContext

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,523 @@
...
+            resourceService.putResource(descriptor, json, EtagHeader.NEW_ONLY);
+            return null;
+        })
+        .compose(ignored -> toFuture(schedule.addAsync(now, dialId)))
+        .mapEmpty();
+    }
+
+    public Future<Boolean> isJobActive(String dialId) {
+        return taskExecutor.submit(() ->
+                resourceService.getResourceMetadata(ResponseIdUtil.getBackgroundJobDescriptor(dialId)) != null);
```
astsiapanay: replace with hasResource

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
```diff
@@ -90,75 +102,82 @@ private Future<Void> forwardToUpstream(ResponseMapping mapping) {
...
+                            .submit(() -> rewriteId(body, mapping.getUpstreamResponseId()))
+                            .compose(rewritten -> {
+                                if (operation == Operation.DELETE) {
+                                    return proxy.getTaskExecutor().submit(() -> {
+                                        proxy.getResponseMappingService().deleteMapping(context.getDialResponseId());
+                                        return null;
+                                    }).compose(ignored -> sendResponse(proxyResponse, rewritten));
+                                }
+                                if (operation == Operation.GET) {
+                                    ResponsesApiClient.TerminalResult terminalResult = tryParseTerminalResult(rewritten);
```
astsiapanay, line 203: Since terminalResult might be null. We can do a check here instead of calling the service

server/src/main/java/com/epam/aidial/core/server/data/BackgroundJobRecord.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Builder;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record BackgroundJobRecord(
+        String perRequestKey,
```
astsiapanay, line 20: Is the PRK encrypted?

server/src/main/java/com/epam/aidial/core/server/log/LogContext.java
```diff
@@ -0,0 +1,124 @@
...
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.List;
+import java.util.Optional;
+
+@Slf4j
+@Getter
+@Builder
+public class LogContext {
```
astsiapanay: Perhaps `AnalyticsLogContext` is more precise name.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,565 @@
...
+        this.client = client;
+        this.logStore = logStore;
+        this.script = redis.getScript(StringCodec.INSTANCE);
+        this.scheduleKey = "background_job_schedule:" + BlobStorageUtil.toStoragePath(prefix, "queue");
+        this.schedule = redis.getScoredSortedSet(scheduleKey, StringCodec.INSTANCE);
+    }
+
+    public void init() {
+        AtomicBoolean polling = new AtomicBoolean(false);
+        vertx.setPeriodic(0, 1, id -> {
```
astsiapanay: The interval is too short(1ms). Please change to 10 seconds and make it configurable.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,565 @@
...
+                            return {}
+                        end
+                        redis.call('HSET', KEYS[2], 'owner', ARGV[4])
+                        redis.call('ZADD', KEYS[1], tonumber(ARGV[2]), ARGV[3])
+                        local attempts = redis.call('HGET', KEYS[2], 'attempts')
+                        local errors = redis.call('HGET', KEYS[2], 'errors')
+                        return {attempts or '0', errors or '0'}
+                        """,
+                RScript.ReturnType.MULTI,
+                List.of(scheduleKey, stateKey(entry.getValue())),
```
astsiapanay: Redis Cluster supports multiple key operations as long as all of the keys involved in a single command execution (or whole transaction, or Lua script execution) belong to the same hash slot. The user can force multiple keys to be part of the same hash slot by using a feature called hash tags.

https://redis.io/docs/latest/operate/oss_and_stack/management/scaling/#redis-cluster-data-sharding

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,565 @@
...
+                        return Future.failedFuture("Unexpected status " + statusCode + " from upstream for background job " + mapping.getUpstreamResponseId());
+                    }
+                    return response.body().map(ResponsesApiClient::parseTerminalBody);
+                });
+    }
+
+    private Future<Void> poll(String dialId, ClaimResult claimResult) {
+        return taskExecutor.submit(() -> loadJobData(dialId))
+                .compose(pair -> {
+                    if (pair == null) {
```
astsiapanay: Can we run the logic sequentially in a virtual thread?

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    private Void tryClaimAndPoll() {
+        if (inFlightJobIds.size() >= settings.getMaxParallelJobs()) {
+            return null;
+        }
+
+        while (true) {
+            long previous = nextJobTime.get();
+            ScoredEntry<String> scoredEntry = schedule.firstEntry();
```
astsiapanay: Can we query range let's say 100 instead of a top element?

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
```diff
@@ -0,0 +1,565 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.Nullable;
+
+@Slf4j
+public class BackgroundJobService {
```
astsiapanay, line 43: I would decompose the service into logical modules:
- Scheduler - schedule the next job.
- Worker aka JobPooling and post processing.

server/src/main/java/com/epam/aidial/core/server/controller/ResponseItemController.java
Oleksii-Klimov: Done.

server/src/main/java/com/epam/aidial/core/server/data/BackgroundJobRecord.java
Oleksii-Klimov: Fixed.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Split it into a job scheduler and a job service.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: I need the context now to store information for logging too.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Replaced.

server/src/main/java/com/epam/aidial/core/server/log/LogContext.java
Oleksii-Klimov: Renamed.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Done.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Using a single key now.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Moved.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: I’m not sure it’s worth optimizing this now. I only need the items that should be checked now, plus the next one after those, so I know when to check again.

server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java
Oleksii-Klimov: Moved.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
Oleksii-Klimov: Removed from context.

server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java
Oleksii-Klimov: This logic differs between responses from Chat Completions and the Responses API.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov:
/deploy-review

>GitHub actions run: [29249470385](https://github.com/epam/ai-dial-ci/actions/runs/29249470385)

| [deploy-review](https://chat-ai-dial-core-pr-1613.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29249470385/job/86814212842) |
| chat | [Failed ❌](https://github.com/epam/ai-dial-ci/actions/runs/29249470385/job/86815992884) |

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
```diff
@@ -0,0 +1,340 @@
...
+                taskExecutor.submit(this::tryClaimAndPoll)
+                        .onComplete(ignore -> polling.set(false))
+                        .onFailure(e -> log.warn("Failed to poll background jobs", e));
+            }
+        });
+        // Set up periodic scans in case of a Redis restart
+        vertx.setPeriodic(0, settings.getScanIntervalMs(), id -> taskExecutor.submit(this::scan));
+    }
+
+    void schedule(String dialId, long nextPollTime) {
```
astsiapanay: It worth renaming `dialId` to `jobId`

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
```diff
@@ -0,0 +1,340 @@
...
+                        end
+                        redis.call('HSET', KEYS[1], 'owner', ARGV[5])
+                        redis.call('ZADD', ARGV[1], tonumber(ARGV[3]), ARGV[4])
+                        local attempts = redis.call('HGET', KEYS[1], 'attempts')
+                        local errors = redis.call('HGET', KEYS[1], 'errors')
+                        return {attempts or '0', errors or '0'}
+                        """,
+                RScript.ReturnType.MULTI,
+                List.of(stateKey(entry.getValue())),
+                scheduleKey,
```
astsiapanay: Why do you pass the key as an argument? That breaks the contract:

```
Important: to ensure the correct execution of scripts, both in standalone and clustered deployments, all keys that a script accesses must be explicitly provided as input key arguments. 

Scripts should only access keys whose names are given as input arguments.

Scripts should never access keys with programmatically-generated names or based on the contents of data structures stored in the database.
```

See https://redis.io/docs/latest/commands/eval/ for more details

astsiapanay (commented):
Code review of `BackgroundJobScheduler` — one functional issue around `nextJobTime` and a couple of smaller notes. See inline comments.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
```diff
@@ -0,0 +1,340 @@
...
+        if (result.isEmpty()) {
+            return null;
+        }
+        int attempts = Integer.parseInt((String) result.get(0));
+        int errors = Integer.parseInt((String) result.get(1));
+
+        return new ClaimResult(owner, attempts, errors);
+    }
+
+    private Future<Void> reschedule(String dialId, long owner, int attempts, int errors, long nextPollTime) {
```
astsiapanay: **Polling degrades to the lease-timeout cadence: `reschedule()` (and `complete()`) never lower `nextJobTime`.**

`nextJobTime` is the gate that decides when the periodic timer actually runs `tryClaimAndPoll` (lines 76-83). Only `schedule()` lowers it via `accumulateAndGet(..., Math::min)` (line 92) — `reschedule()` does not.

Trace a not-yet-finished job:
1. `claim()` runs `ZADD schedule leaseUntil dialId`, so the job's Redis score becomes `now + leaseTimeoutMs` (5 min).
2. Back in the claim loop, the next `firstEntry()` sees that job at `leaseUntil > now`, sets `nextJobTime = leaseUntil` and breaks (lines 119-121).
3. Asynchronously the poll finishes "not done" and `reschedule()` sets the Redis score to `now + backoff` (starts at 10 s) — but `nextJobTime` stays at `leaseUntil`.

So the job is due in Redis at `now + 10s`, but the timer won't wake until `now + 5min`. The 10-min `scan()` can't help either — it calls `schedule()`, which uses `addIfAbsentAsync` and is a no-op for an entry already present.

**Effect:** after the first poll, the configured `initialPollIntervalMs`/`pollBackoffFactor` schedule is effectively ignored; every subsequent poll (and the error-retry path at line 168) is throttled to `leaseTimeoutMs`.

**Fix:** lower the hint whenever a new due time is pushed, e.g. at the top of `reschedule()`:
```java
nextJobTime.accumulateAndGet(nextPollTime, Math::min);
```
An over-eager wake-up is harmless — `firstEntry()` just finds nothing due and re-arms `nextJobTime` correctly.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
```diff
@@ -0,0 +1,340 @@
...
+            }
+
+            long now = System.currentTimeMillis();
+            long score = scoredEntry.getScore() == null ? now : scoredEntry.getScore().longValue();
+            if (score > now) {
+                nextJobTime.compareAndSet(previous, score);
+                break;
+            }
+
+            if (inFlightJobIds.size() >= settings.getMaxParallelJobs()) {
```
astsiapanay, line 139: When the loop breaks on the parallelism cap, `nextJobTime` is left unchanged (typically `<= now`, since we entered the loop). It happens to work — the next tick retries because the gate is still open — but the behavior is incidental. Worth a short comment, or explicitly keeping `nextJobTime` at "now" so the intent is clear. This also interacts with the `reschedule()` fix above: make sure hitting the cap keeps the gate open.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        return null;
+    }
+
+    @Nullable
+    private ClaimResult claim(ScoredEntry<String> entry) {
+        long leaseUntil = System.currentTimeMillis() + settings.getLeaseTimeoutMs();
+        long owner = ThreadLocalRandom.current().nextLong();
+        List<Object> result = script.eval(
```
astsiapanay, line 210: Minor/consistency: `claim()` uses the synchronous `script.eval`, while `reschedule()`/`complete()` use `evalAsync`. It's safe here because `tryClaimAndPoll` runs on a virtual thread (blocking one is cheap), but the inconsistency is easy to misread as "safe from the event loop." A one-line comment noting `claim` is only invoked from the virtual-thread claim loop would prevent a future caller from blocking the event loop.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
Oleksii-Klimov: Added a comment.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
Oleksii-Klimov: Fixed.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
Oleksii-Klimov: Fixed.

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobScheduler.java
Oleksii-Klimov: Renamed inside the scheduler.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov:
/deploy-review

>GitHub actions run: [29404226627](https://github.com/epam/ai-dial-ci/actions/runs/29404226627)

| [deploy-review](https://chat-ai-dial-core-pr-1613.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29404226627/job/87315692255) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29404226627/job/87323689280) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov:
/deploy-review

>GitHub actions run: [29415343240](https://github.com/epam/ai-dial-ci/actions/runs/29415343240)

| [deploy-review](https://chat-ai-dial-core-pr-1613.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29415343240/job/87351968423) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29415343240/job/87354192943) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/log/AnalyticsLogContext.java
```diff
@@ -0,0 +1,242 @@
...
+
+    private final String upstreamEndpoint;
+
+    private final int responseStatusCode;
+    private final Buffer responseBody;
+    private final String assembledStreamingResponse;
+
+    private final TokenUsage tokenUsage;
+
+    public static AnalyticsLogContext from(ProxyContext context, String assembledStreamingResponse) {
```
astsiapanay, line 66: Nit. We can overload the method to accept just ProxyContext

astsiapanay (dismissed):

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Added hash tags the keys.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
Oleksii-Klimov: Refactored.

server/src/main/java/com/epam/aidial/core/server/log/AnalyticsLogContext.java
Oleksii-Klimov: assembledStreamingResponse is different for chat completions and responses. Keeping it as is.

PR #1645, opened 2026-06-22
fix: handle URL-encoding asymmetry in AutoShareDeploymentFn #1632

### LINKED ISSUE #1632: 403 "Forbidden deployment" when a global interceptor is active and the deployment id contains a space (URL-encoding asymmetry in AutoShareDeploymentFn)
### Name and Version

ai-dial-core 0.43.2 (running image reports service.version 0.43.2). Affected code is unchanged on `development` and up to 0.44.4.

> Scope: only reproduced/tested with a **global interceptor**. A per-deployment interceptor uses the same `AutoShareDeploymentFn` code path and is expected to be affected too, but I have not verified that configuration.

1. Create a **private** application (e.g. a Quick App 2.0) whose deployment id contains a space, e.g. `applications/<bucket>/demo a_0.0.1`.
2. Configure a **global interceptor** (`globalInterceptors`) so it applies to the request.
3. Call `POST /openai/deployments/applications/<bucket>/demo%20a_0.0.1/chat/completions`.
4. Core returns `403` with body `Forbidden deployment: applications/<bucket>/demo a_0.0.1`.

Controls confirming the trigger is "global interceptor present **and** space in id":

| Interceptor   | Space in id | Result |
|---------------|-------------|--------|
| no            | no          | 200    |
| no            | yes         | 200    |
| yes (global)  | no          | 200    |
| yes (global)  | yes         | **403** |

Only the intersection fails.

The request should be processed (200), exactly as it is without the interceptor or without the space. A space (or any character that requires percent-encoding) in a deployment id must be handled consistently in both the direct and the interceptor code paths.

### What do you see instead?

`403 Forbidden deployment: applications/<bucket>/demo a_0.0.1`.

Sanitized Core log (the URL arrives percent-encoded as `demo%20a...`, but the error shows the decoded `demo a...`):

```
InterceptorController     Received request from client. Deployment: <interceptor>. request.uri: /openai/deployments/applications/<bucket>/demo%20a_0.0.1/chat/completions
InterceptorController     Connected to interceptor.
ProxyContext              Responding with error. Body: Forbidden deployment: applications/<bucket>/demo a_0.0.1   status=403
DeploymentPostController  Forbidden deployment applications/<bucket>/demo a_0.0.1   status=403
```

### Root cause

In the interceptor flow, `AutoShareDeploymentFn` grants the per-request API key READ access to the initial deployment. It builds the descriptor from the **raw** id:

```java
// AutoShareDeploymentFn#toResourceDescriptor
return ResourceDescriptorFactory.fromAnyUrl(id, proxy.getEncryptionService());
```

`id` is `context.getInitialDeployment()` — the **percent-decoded** form (literal space). `fromAnyUrl` → `fromUrl` → `fromEncoded` applies `UrlUtil.decodePath` per segment, and `decodePath` runs `new URI(segment)`, which throws `URISyntaxException` on a literal space. The exception is swallowed (`catch (Exception e) { return null; }`), so `apply()` returns early and the grant is **never** added to `apiKeyData.getAttachedDeployments()`.

On the interceptor callback, `DeploymentService#toResourceDescriptor` resolves the same deployment but **percent-encodes first** (`UrlUtil.encodePath(resourceUrl)`), builds the descriptor fine, then `AccessService.hasReadAccess` finds no grant for the per-request key → `throwForbiddenDeploymentError` → 403.

**The asymmetry is the bug:** `DeploymentService` calls `fromAnyUrl(encodePath(id))`; `AutoShareDeploymentFn` calls `fromAnyUrl(id)` on the raw id and cannot handle ids that require URL-encoding.

Minimal confirmation (plain Java): `new URI("demo a_0.0.1")` throws `URISyntaxException: Illegal character in path at index 4`; `new URI("demo%20a_0.0.1")` parses fine.

Relevant code (tag `0.43.2`):
- `AutoShareDeploymentFn#toResourceDescriptor` — https://github.com/epam/ai-dial-core/blob/0.43.2/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java#L52-L60
- `UrlUtil#decodePath` (`new URI`) — https://github.com/epam/ai-dial-core/blob/0.43.2/storage/src/main/java/com/epam/aidial/core/storage/util/UrlUtil.java#L61-L75
- `ResourceDescriptorFactory#fromEncoded` — https://github.com/epam/ai-dial-core/blob/0.43.2/server/src/main/java/com/epam/aidial/core/server/util/ResourceDescriptorFactory.java#L28-L34
- `DeploymentService#toResourceDescriptor` / `throwForbiddenDeploymentError` — https://github.com/epam/ai-dial-core/blob/0.43.2/server/src/main/java/com/epam/aidial/core/server/service/DeploymentService.java#L86-L107
- `AccessService` attached-deployments lookup — https://github.com/epam/ai-dial-core/blob/0.43.2/server/src/main/java/com/epam/aidial/core/server/security/AccessService.java#L215

### Proposed fix

Make `AutoShareDeploymentFn#toResourceDescriptor` consistent with `DeploymentService` by encoding the id before resolving:

```diff
diff --git a/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java b/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java
index fe6f0ca..9159260 100644
--- a/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java
@@ -13,6 +13,7 @@ import com.epam.aidial.core.server.util.ResourceDescriptorFactory;
 import com.epam.aidial.core.storage.http.HttpException;
 import com.epam.aidial.core.storage.http.HttpStatus;
 import com.epam.aidial.core.storage.resource.ResourceDescriptor;
+import com.epam.aidial.core.storage.util.UrlUtil;
 
 import javax.annotation.Nullable;
 
@@ -54,7 +55,7 @@ public class AutoShareDeploymentFn extends BaseRequestFunction<RequestObject> {
             return null;
         }
         try {
-            return ResourceDescriptorFactory.fromAnyUrl(id, proxy.getEncryptionService());
+            return ResourceDescriptorFactory.fromAnyUrl(UrlUtil.encodePath(id), proxy.getEncryptionService());
         } catch (Exception e) {
             return null;
         }
```

No-op for ids without special characters; fixes the space case. Verified on a running instance via hot class redefinition: with the patched method the same request returns 200.

### Additional notes (optional)
- Consider not silently swallowing the parse failure in `toResourceDescriptor`; it hid this issue entirely.
- Both `DeploymentService` and `AutoShareDeploymentFn` have their own `toResourceDescriptor`; a shared helper (`fromAnyUrl(encodePath(id))`) would stop them drifting apart again.

### Additional information

- The `AutoShareDeploymentFn` auto-share path was introduced in #1336 (fix for #1323), so this is an edge case not covered by that change.
- The proposed fix was verified on a running instance via hot class redefinition (Arthas `redefine`): with the patched method, the same request that returned 403 now returns 200.

astsiapanay:
Fixes a `403 Forbidden deployment` error that occurred when a global interceptor was active and a deployment id contained a character requiring percent-encoding (e.g. a space). The interceptor auto-share path (`AutoShareDeploymentFn`) built the resource descriptor from the raw, percent-decoded id, which caused `UrlUtil.decodePath` to throw `URISyntaxException`; the swallowed exception meant the per-request API key never received the READ grant, so the subsequent interceptor callback was rejected with 403.

### Applicable issues

- fixes #1632

### Description of changes

- `AutoShareDeploymentFn#toResourceDescriptor` now percent-encodes the id via `UrlUtil.encodePath(id)` before calling `ResourceDescriptorFactory.fromAnyUrl`, making it consistent with `DeploymentService#toResourceDescriptor`. This is a no-op for ids without special characters.
- Added regression test `testApply_WhenInitialDeploymentIdContainsSpace` verifying the auto-share grant is registered for a deployment id containing a space.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/AutoShareDeploymentFn.java
@@ -14,6 +14,7 @@ import com.epam.aidial.core.server.validation.ApplicationTypeResourceException;
 import com.epam.aidial.core.storage.http.HttpException;
 import com.epam.aidial.core.storage.http.HttpStatus;
 import com.epam.aidial.core.storage.resource.ResourceDescriptor;
+import com.epam.aidial.core.storage.util.UrlUtil;
 
 import javax.annotation.Nullable;
 
@@ -64,7 +65,7 @@ public class AutoShareDeploymentFn extends BaseRequestFunction<RequestObject> {
             return null;
         }
         try {
-            return ResourceDescriptorFactory.fromAnyUrl(id, proxy.getEncryptionService());
+            return ResourceDescriptorFactory.fromAnyUrl(UrlUtil.encodePath(id), proxy.getEncryptionService());
         } catch (Exception e) {
             return null;
         }
--- a/server/src/test/java/com/epam/aidial/core/server/function/AutoShareDeploymentFnTest.java
+++ b/server/src/test/java/com/epam/aidial/core/server/function/AutoShareDeploymentFnTest.java
@@ -85,6 +85,22 @@ public class AutoShareDeploymentFnTest {
     }
 
     @Test
+    public void testApply_WhenInitialDeploymentIdContainsSpace() {
+        when(encryptionService.decrypt(anyString())).thenReturn("/Users/user/");
+        when(proxy.getEncryptionService()).thenReturn(encryptionService);
+        when(proxy.getAccessService()).thenReturn(accessService);
+        when(context.getInitialDeployment()).thenReturn("applications/123/demo a_0.0.1");
+        when(accessService.hasReadAccess(any(ResourceDescriptor.class), eq(context))).thenReturn(true);
+        when(context.getProxyApiKeyData()).thenReturn(proxyApiKeyData);
+        when(proxy.getDeploymentService()).thenReturn(deploymentService);
+
+        assertFalse(fn.apply(EMPTY_OBJECT));
+        assertFalse(proxyApiKeyData.getAttachedDeployments().isEmpty());
+        assertEquals(ResourceAccessType.READ_ONLY,
+                proxyApiKeyData.getAttachedDeployments().get("applications/123/demo%20a_0.0.1").accessTypes());
+    }
+
+    @Test
     public void testApply_WhenInitialDeploymentIsApp() {
         when(encryptionService.decrypt(anyString())).thenReturn("/Users/user/");
         when(proxy.getEncryptionService()).thenReturn(encryptionService);


Oleksii-Klimov (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1646, opened 2026-06-23
fix: process compressed responses in streaming mode #1643

astsiapanay:
Enable response decompression on the proxy HttpClient so gzip-encoded upstream responses (including SSE streams) are inflated before DIAL parses, re-serializes, and proxies them. Previously a compressed streaming response was fed as gzip binary into the SSE parser, which emitted nothing and closed the connection without output.

### Applicable issues

- fixes #1643

### Description of changes

- Set `setDecompressionSupported(true)` on the shared proxy `HttpClient` in `AiDial.java`. Vert.x/Netty's `HttpContentDecompressor` decompresses `gzip`/`deflate` bodies incrementally (streaming stays streaming) and strips the `Content-Encoding` header, so the header copied to the client is correct and the SSE parser receives plaintext.
- This covers every consumer of the proxy client (chat completions, responses API, interceptors, MCP) instead of patching the streaming machinery per controller.
- Added `testStreaming_DecompressesGzipResponse` in `DeploymentPostApiTest`: the upstream returns a gzip-compressed chunked `text/event-stream` body; the test asserts the client receives decoded SSE content and that no `Content-Encoding` header leaks through.

### Checklist

- [X] Title of the pull request follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1643, opened 2026-06-22
DIAL Core fails to process compressed responses in streaming mode


Oleksii-Klimov:
### Name and Version

DIAL Core 0.44.4

### What steps will reproduce the bug?

1. Configure a new model to use another DIAL Core's chat completion endpoint.
2. Ensure that the second endpoint returns a compressed responses, for example with:

   ```http
   Content-Encoding: gzip
   ```

3. Send a streaming chat completion request through the first DIAL Core instance to the configured model:

   ```json
   {
     "model": "<configured-model-name>",
     "messages": [
       {
         "role": "user",
         "content": "Hello"
       }
     ],
     "stream": true
   }
   ```

4. Observe the response from the first DIAL Core instance.

### What is the expected behavior?

A valid SSE data.

### What do you see instead?

Connection is closed without producing an output.

PR #1650, opened 2026-06-23
chore: Release patch 0.44.5

DmytroZaichenkoDev:
Cherry-picks #1646 (`c27523e4`) onto `release-0.44`.

### Applicable issues

- backports fix for #1643

### Description of changes

Enables response decompression on the proxy `HttpClient` so gzip/deflate-encoded upstream responses (including SSE streams) are inflated before DIAL parses, re-serializes, and proxies them. Previously a compressed streaming response was fed as gzip binary into the SSE parser, which emitted nothing and closed the connection without output.

- `clientOptions.setDecompressionSupported(true)` on the shared proxy client in `AiDial.java`. Netty's `HttpContentDecompressor` decompresses incrementally (streaming stays streaming) and strips the `Content-Encoding` header, so the header copied to the client stays correct and the SSE parser receives plaintext.
- Covers every consumer of the proxy client (chat completions, responses API, interceptors, MCP) rather than patching streaming per controller.
- Includes `testStreaming_DecompressesGzipResponse` in `DeploymentPostApiTest`.

### Verification

- Cherry-pick applied cleanly onto `release-0.44` (no conflicts).
- Confirmed it also resolves the MCP `tools/list` gzipped-SSE hang on the 0.44 line: with this change a previously-buggy `McpProxyController` SSE path passes our gzip regression test; without it, the client fails decoding the gzip stream.
- Harmless alongside existing 0.44 gzip handling — the per-controller `Compression.decodeHttpBody` becomes a no-op (header already stripped), and `ResourceAuthorizationClient` is unaffected (separate `java.net.http.HttpClient`).

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/AiDial.java
+++ b/server/src/main/java/com/epam/aidial/core/server/AiDial.java
@@ -164,6 +164,8 @@ public class AiDial {
 
             vertx = Vertx.vertx(vertxOptions);
             HttpClientOptions clientOptions = new HttpClientOptions(settings("client"));
+            // upstream bodies (including SSE streams) are parsed/proxied, so they must be decoded
+            clientOptions.setDecompressionSupported(true);
             HttpProxySelector httpProxySelector = createHttpProxySelector(clientOptions);
             client = vertx.createHttpClient(clientOptions);
             WebSocketClientOptions webSocketClientOptions = new WebSocketClientOptions(settings("webSocketClient"));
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/28041228564)

astsiapanay (approved):

PR #1629, opened 2026-06-17
feat: external-service credentials for DIAL applications

DmytroZaichenkoDev:
## Summary

Lets DIAL applications use end-user (or app-level) credentials to call external services, mirroring the toolset credentials model. External services are defined inline on an application (`external_services` map with `auth_settings`); credentials are signed in per user/app, stored encrypted at rest, and retrieved by the app backend at runtime via the per-request key.

Relates to #1435.

## What's included

- **Inline definition + encryption-at-rest** — `external_services` on `Application`; `client_secret` encrypted at rest (per-bucket CEK, AAD-bound), preserve-on-omit on update, re-encrypt on publication/copy.
- **APIs**
  - `POST /v1/ops/external-service/{signin,signout}` — user/app sign-in (OAuth + API key)
  - `POST /v1/ops/external-service/credentials` — runtime retrieval via per-request key (app-binding + per-user isolation)
  - `GET/PUT/DELETE /v1/applications/{app_id}/external-services[/{id}]` — admin/owner management
- **Read surface** — deployment GET (`/openai/applications/{id}`) and the raw resource GET expose `external_services` with per-user sign-in status; secrets stripped (parity with toolset reads).
- **Validation** — first-time OAUTH create requires `client_secret` (`CREATE_STATIC_CLIENT`); updates may omit it (preserve-on-omit).

## Review guide (4 commits, bottom → top)

1. `feat: external-service credential model, scope locator, and resolution` — config model, `CredentialsLocatorFactory`, USER→APPLICATION resolution, status service.
2. `feat: encrypt external_services client_secret at rest + write-path hardening` — encryption-at-rest, create validation, status write-strip, config validation, publication double-encrypt fix.
3. `feat: external-service sign-in/out, credential retrieval, and management APIs` — controllers, DTOs, routes, integration tests.
4. `feat: expose external_services with per-user sign-in status on app reads` — status enrichment + secret stripping on app GETs.

Each commit compiles independently.

## Testing

- Integration/unit: `ExternalServiceCredentialsApiTest`, `PublicationApiTest`, `CustomApplicationApiTest`, `ResourceCredentialsServiceTest` (and others) — green; checkstyle clean.
- Manually verified end-to-end against a running core + mock app backend: USER/APP sign-in → retrieval, OAuth refresh, cross-app isolation, management CRUD + authz (owner / non-owner / admin), encryption-at-rest, app-sharing credential flow, create-validation, secret stripping, and adding services via both the application PUT and the management API.

## Out of scope / notes

[...]

PR #1677, opened 2026-07-01
feat: on-behalf-of external-service credentials for Scheduler #1649

### LINKED ISSUE #1649: Headless on-behalf-of credential retrieval for external services (Scheduler support)
## Background
DIAL apps can already call external services (e.g. Salesforce) using credentials a user signed in with — but **only while that user is actively using the app**. The app gets a short-lived per-request key during the request and uses it to fetch the user's token.

The **Scheduler** breaks this: it runs a user's task on a schedule, when the user is **offline**. There's no active request and no per-request key, so today there's no safe way for the Scheduler to obtain a user's external-service token to run their job.

## Goal
Allow a trusted background service (the Scheduler) to retrieve a user's external-service credential **headlessly** — without that user being online — in a way that is explicitly authorized and fails safely.

## How it works (high level)
1. **Setup (user online):** the user signs in to the external service and explicitly **consents to offline use**. DIAL stores their tokens as today, plus the consent.
2. **Firing (user offline):** the Scheduler calls a new endpoint with its own identity, naming the app, the external service, and the owner. DIAL checks three things, then returns the credential header. The Scheduler never sees refresh tokens — DIAL refreshes them server-side.
3. **Use:** the Scheduler attaches the returned header to its own call to the external service.

## The three safety checks
- **Authorized caller** — only services the app explicitly trusts can make headless calls for it.
- **Recorded consent** — only fires if the owner agreed to offline use; signing out revokes it.
- **Fail closed** — uses the owner's own credential only, never a silent fallback to a shared one; a missing/revoked grant returns a clear, actionable error.

## Scope
- New headless credential-retrieval endpoint; DIAL refreshes tokens server-side.
- Per-app declaration of which callers may act headlessly (DIAL key today; built to also accept cloud workload identity / MSI).
- Offline-consent opt-in at sign-in, revoked at sign-out.
- Both service-provisioning models: admin-provisioned services (each user signs into their own account) **and** users authoring/seeing their own external services, gated per app.
- Distinct audit events for headless retrieval with trace propagation.

## Boundary (owned elsewhere)
- The owner-identity token the Scheduler uses to call DIAL Core *as* the user is owned by the separate **Scheduler design**, not this issue. This work covers the external-service credential side only.

## Acceptance criteria
- [ ] A trusted service can retrieve an offline-consented user's external-service header while the user is offline.
- [ ] Expired tokens are refreshed transparently; the caller never handles refresh tokens.
- [ ] Unauthorized callers, missing consent, and missing/revoked credentials each return distinct, correct errors.
- [ ] Sign-out stops all further headless issuance for that user.
- [ ] Admin-provisioned and user-authored external services are both supported, isolated per user.
- [ ] Existing (online, per-request-key) flow is unchanged; headless is off by default for every app.

DmytroZaichenkoDev:
Extends the External Service Credentials feature (#1629) so a trusted background actor (the Scheduler) can retrieve an absent owner's external-service credential headlessly, and so regular users can author their own external services. Additive and default-off; existing app-scoped, admin-provisioned behavior is unchanged.

### Applicable issues

- fixes #1649

### Description of changes

**On-behalf-of (OBO) retrieval** — new `POST /v1/ops/external-service/obo-credentials` (project-key / workload-JWT; per-request key forbidden). Body `{url, owner_sub}` → downstream `{header_name, header_value, expires_at?}`. Gated by an actor check, offline consent, and fail-closed USER-level resolution:
- Actor gate: caller identity (SHA-256 of the DIAL key, or workload JWT `azp`) must equal the owning app's `app_identity`.
- Requires the owner's recorded `offline_usage_consent` (captured at sign-in, dropped at sign-out).
- USER-level only, no APP/GLOBAL fallback: missing → 404, no consent → 403, gate mismatch → 403, permanent refresh failure → 401.

**User-authored external services** (admin opt-in via `allow_user_external_services`):
- Stored in the author's own bucket (`Users/{userId}/external_services/applications/{app_id}/{id}`), `client_secret` encrypted under the owner's CEK; isolation rides on bucket ownership.
- Read-surface merge: management GET/list, the deployment view (`/openai/applications/{id}`) and the raw resource GET now show the caller's own definitions (secret-stripped).
- Delete cascade: removing a user-authored definition purges the author's own USER-level credential.

**Admin-config-only fields** on `Application`: `app_identity` and `allow_user_external_services` are honored only from the config file / `/v1/admin/apply`, stripped from the self-service `PUT`; `app_identity` is hidden from non-admin reads.

**Audit (design §8)** — each OBO call emits one distinct event on the `DIAL_OBO_AUDIT` logger (outcome, non-secret actor, `owner_sub`, application/service, `trace_id`, reason on failure); never logs credential material.

Verified end-to-end against a local stack (real keycloak JWTs + a DIAL project key) in addition to the automated suites.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -0,0 +1,48 @@
...
+
+    // Non-secret evidence of the calling actor: the DIAL key's project, or the workload JWT's azp (Azure v1: appid).
+    private static String actorEvidence(ProxyContext context) {
+        Key key = context.getKey();
+        if (key != null) {
+            return "project:" + key.getProject();
+        }
+        ExtractedClaims claims = context.getExtractedClaims();
+        String azp = claims == null ? null : claims.authorizedParty();
+        return azp == null ? "unknown" : "azp:" + azp;
```
astsiapanay: Should we throw exception if actor is unknown?

server/src/main/java/com/epam/aidial/core/server/service/UserExternalServiceService.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * User-authored external-service definitions (design §9/§10): each is its own resource in the author's
```
astsiapanay: Where is the design location?

```
(design §9/§10)
```

astsiapanay (commented):
## Automated multi-aspect review

Reviewed across code quality, silent-failure/error-handling, test coverage, comments, and type design. **Verdict: mergeable — no Critical defects.** The fail-closed security core is correct and well-tested (actor gate denies on null/mismatch; consent gated *before* refresh so a non-consented probe never rotates the owner's token; USER-only locator makes APP/GLOBAL fallback structurally impossible; 401/403/404 mapping correct; secret stripping centralized in `withoutSecrets()`; audit never logs credentials). Inline comments below, tagged **Important** / **Suggestion**.

### Test gaps worth adding
- Same DIAL key trusted by app A → OBO on app B whose `app_identity` is **non-null and non-matching** → 403 (current test only covers the `app_identity == null` branch).
- `owner_sub` naming a different real, non-consented user → 404; valid-shape `url` for a non-existent app/service → 404.
- Workload caller carrying only `appid` (no `azp`) → match (the Azure v1 branch is unexercised).
- Assert OBO OAuth `expires_at` is a plausible epoch-**seconds** value, not merely non-null (guards the ms→s math).
- User-authored **OAUTH** service: secret encrypted, stripped on reads, still resolvable via OBO.
- Inline-admin vs user-authored **id collision** → inline wins (`putIfAbsent` precedence rule).
- Anti-leak body assertions on the consent-denied and gate-mismatch 403 paths.

### Note on a suggestion NOT recommended
Marking `Application.appIdentity` Jackson `WRITE_ONLY` to auto-strip it from reads would be **unsafe**: the same `ProxyUtil.convertToString` serializes `Application` both to blob storage and to client responses, so `WRITE_ONLY` would drop `app_identity` from the stored blob and silently disable OBO. The manual read-surface scrubs in `ResourceController` are the correct pattern here.

### Strengths
Integration tests assert real side-effects (token-endpoint call counts, purge-via-reissue), not just status codes; centralized `withoutSecrets()`; behavior-with-data `authorizedParty()`; exhaustive audit-outcome `switch`; consent-before-refresh backed by a dedicated test.

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
```diff
@@ -74,14 +76,50 @@ public Future<?> listExternalServices(String appId) {
...
+    // Callers see inline (admin) definitions when they can manage them, unioned with their own user-authored
+    // ones (inline wins on id clash) — so a service authored before gaining write access stays visible.
+    private Map<String, ExternalService> manageableServices(ResolvedApp resolved, String appId) {
+        Map<String, ExternalService> services = new LinkedHashMap<>();
+        boolean inline = canManageInline(resolved);
+        if (inline && resolved.application.getExternalServices() != null) {
+            services.putAll(resolved.application.getExternalServices());
+        }
+        if (resolved.application.isAllowUserExternalServices()) {
+            Map<String, ExternalService> userAuthored = userExternalServiceService.list(context.getUserId(), appId);
```
astsiapanay: **Important — fail-open on null user identity.** These management paths call `userExternalServiceService.{list,get,put,delete}(context.getUserId(), …)` (here and lines 115, 140, 161) with no null guard — unlike the two read overlays (`ApplicationController.overlayUserAuthoredServices`, `ResourceController.overlayUserAuthoredServices`) which early-return when `getUserId() == null`. A userless principal (a project key with no user sub) on an app with `allow_user_external_services=true` falls into the user-authoring branch and derives `BucketBuilder.USER_BUCKET_PATTERN.formatted(null)` → literal `Users/null/`, a bucket shared across all userless callers (each blob holds an encrypted `client_secret`). A null owner identity should be a hard 403, not a silently shared namespace. Suggest rejecting null/blank owner before deriving the bucket (centralize in `UserExternalServiceService`).

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceCredentialsController.java
```diff
@@ -170,30 +178,109 @@ public Future<?> getCredentials() {
...
+
+        context.getRequest()
+                .body()
+                .compose(body -> taskExecutor.submit(() -> {
+                    OboCredentialsRequest request = ProxyUtil.convertToObject(body, OboCredentialsRequest.class);
+                    ValidationUtil.validate(request);
+
+                    try {
+                        String[] scope = CredentialsLocatorFactory.parseExternalServiceScope(request.getUrl());
+                        ResolvedExternalService resolved = resolveExternalService(request.getUrl(), request.getOwnerSub());
```
astsiapanay: **Important — actor gate runs *after* owner-scoped resolution (enumeration oracle).** `resolveExternalService(url, ownerSub)` runs before `AppIdentityMatcher.matches(...)` on line 212. For a user-authored service, resolution reads the *owner's private bucket* (`userExternalServiceService.get(ownerSub, …)`). An untrusted caller supplying an arbitrary `ownerSub`/`url` can distinguish 404 (no such service) from 403 (exists, not trusted) — a low-value oracle over who authored what, plus an owner-bucket read before authorization. The gate only needs `application.getAppIdentity()` (from the app definition, not the owner bucket). Suggest: resolve the application → run the actor gate → *then* resolve the external service / owner credential.

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceCredentialsController.java
```diff
@@ -170,30 +178,109 @@ public Future<?> getCredentials() {
...
+                        // Audit failures too — including a malformed url that fails scope parsing (best-effort ids).
+                        String[] scope = safeParseScope(request.getUrl());
+                        ExternalServiceAuditLog.oboRetrieval(context, scope[0], scope[1], request.getOwnerSub(), e);
+                        throw e;
                     }
-                    return response;
                 }))
                 .onSuccess(response -> context.respond(HttpStatus.OK, response))
-                .onFailure(error -> respondError("Can't get external service credentials", error));
+                .onFailure(error -> respondError("Can't get on-behalf-of external service credentials", error));
```
astsiapanay, line 317: **Suggestion — error not logged; internal message leaked to client (in `respondError`).** In `respondError` the `EncryptionException` branch returns `e.getMessage()` to the caller and logs nothing; the `HttpException` branch (5xx) is also unlogged. A 500 on the OBO path (crypto/config failure, upstream 5xx) then leaves no server-side record correlating it to the request/trace, and may expose internal detail. Log 5xx / crypto failures with the trace id and return a generic body for `EncryptionException`. (The OBO audit log does record `outcome=ERROR`, which softens this.)

server/src/main/java/com/epam/aidial/core/server/security/ExtractedClaims.java
```diff
@@ -1,9 +1,25 @@
...
+     * The token's authorized party — the calling workload's client id: the {@code azp} claim, or {@code appid}
+     * (Azure v1) as a fallback. {@code null} when neither is present.
+     */
+    public String authorizedParty() {
+        if (userClaims == null) {
+            return null;
+        }
+        JsonNode azp = userClaims.get("azp");
+        if (azp == null || azp.isNull()) {
+            azp = userClaims.get("appid");
```
astsiapanay: **Suggestion — `appid` fallback broadens the OBO trust surface (needs an explicit decision).** On Azure AD v1 tokens `appid` is frequently the *client app that requested the token* (e.g. the interactive chat UI), not a headless workload. If an app's `app_identity` (azp-form) ever collides with a UI client id present in ordinary users' tokens, those interactive users would pass the OBO actor gate. Either document that azp-form `app_identity` must be a dedicated workload client id, or drop the `appid` fallback if only true `azp` workloads are intended. (Also currently untested — see summary.)

server/src/main/java/com/epam/aidial/core/server/security/ExtractedClaims.java
```diff
@@ -1,9 +1,25 @@
...
+        }
+        return azp == null || azp.isNull() ? null : azp.asText();
```
astsiapanay: **Suggestion — treat blank as absent.** A present-but-blank `azp`/`appid` returns `""`, which `AppIdentityMatcher` could then match against a blank `app_identity`. Return `null` for empty/whitespace (e.g. `StringUtils.trimToNull(azp.asText())`).

server/src/main/java/com/epam/aidial/core/server/security/AppIdentityMatcher.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * The OBO actor gate: decides whether the caller's own identity matches an application's {@code app_identity}.
+ * The caller identity is derived from how it authenticated — a DIAL key (compared as the SHA-256 hex of the key)
+ * or an OAuth workload token (compared as its authorized party — {@code azp}, or Azure v1 {@code appid}).
+ */
+@UtilityClass
+public class AppIdentityMatcher {
+
+    public static boolean matches(ProxyContext context, String appIdentity) {
+        if (appIdentity == null) {
```
astsiapanay: **Suggestion — guard blank, not just null.** `if (appIdentity == null)` lets an empty-string `app_identity` reach the `equals` branches; an empty `azp` could then match it. Use `StringUtils.isBlank(appIdentity)` so blank ⇒ OBO disabled ⇒ never matches (fail closed). (Also: consider returning a verified-actor capability the OBO resolver *requires* rather than a `boolean`, so the gate can't be skipped by a future call site — larger, optional.)

config/src/main/java/com/epam/aidial/core/config/Application.java
```diff
@@ -43,6 +43,17 @@ public class Application extends Deployment {
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private Map<String, ExternalService> externalServices = new LinkedHashMap<>();
 
+    // The app's own actor identity (#1) for OBO credential retrieval: SHA-256 hex of its DIAL key, or
+    // (when wired) its workload client_id. The caller's derived identity must equal it. Absent ⇒ OBO off.
```
astsiapanay: **Suggestion — stale comment.** "(when wired)" implies the workload `client_id` comparison is not yet functional, but it is fully wired: `AppIdentityMatcher.matches` compares `appIdentity` against `claims.authorizedParty()` (`azp`/`appid`), and `testOboWorkloadIdentityMatch` exercises exactly this path. Drop "(when wired)" and the cryptic "(#1)".

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
```diff
@@ -223,6 +225,48 @@ public ResourceCredentials getRefreshedResourceCredentials(CredentialsLocator cr
         return null;
     }
 
+    /**
+     * Resolves the owner's USER-level credential only, with auto-refresh — and <b>no fallback</b> to
```
astsiapanay, line 229: **Suggestion — Javadoc return contract is incomplete.** The method does not always return `null` on the unhappy path: when a USER credential exists but `offlineUsageConsent` is false, it returns the *stored, un-refreshed* credential so the caller can reject on consent (403) rather than 404 — and deliberately does not rotate the refresh token. Worth stating in the class-facing Javadoc, since a caller reads only "returns null when the owner has no USER-level credential."

server/src/main/java/com/epam/aidial/core/server/service/UserExternalServiceService.java
```diff
@@ -0,0 +1,183 @@
...
+@RequiredArgsConstructor
+public class UserExternalServiceService {
+
+    private static final int LIST_PAGE_SIZE = 1000;
+
+    private final ResourceService resourceService;
+    private final ResourceAuthSettingsEncryptionService secretEncryptionService;
+    private final EncryptionService bucketEncryptionService;
+
+    public ResourceDescriptor descriptor(String ownerSub, String appPart, String serviceId) {
```
astsiapanay, line 45: **Suggestion — validate `serviceId` on every access, not just writes.** `descriptor()`/`folderDescriptor()` build the `ResourceDescriptor` directly (serviceId as name, `appPart.split("/")` as parents); the constructor does not reject `..`/empty segments. Only `put` validates `serviceId` (`SERVICE_ID_PATTERN`) — `get`/`delete` rely on write-path history. Not exploitable today (serviceId arrives as a single routed segment), but re-validating in `descriptor()` makes the isolation guarantee self-enforcing.

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -0,0 +1,48 @@
...
+        String outcome = switch (error) {
+            case null -> "SUCCESS";
+            case PermissionDeniedException ignored -> "DENIED";
+            case ResourceNotFoundException ignored -> "NOT_FOUND";
+            default -> "ERROR";
+        };
+        AUDIT.info("event=obo_credential_retrieval outcome={} actor={} owner_sub={} application_id={} "
+                        + "external_service_id={} trace_id={}{}",
+                outcome, actorEvidence(context), ownerSub, applicationId, externalServiceId, context.getTraceId(),
+                error == null ? "" : " reason=\"%s\"".formatted(error.getMessage()));
```
astsiapanay: **Suggestion (low) — make the no-secret guarantee explicit.** `reason` echoes `error.getMessage()`. Safe today (verified `HttpException.getMessage()` returns only `message`, never the response `body`), but the class docstring promises "never credential material" while this relies on an incidental property. Add a comment (or whitelist per exception type) so a future exception embedding token/secret text in its message doesn't silently land in the audit stream.

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -181,6 +181,7 @@ public Pair<ResourceItemMetadata, Application> putApplication(ResourceDescriptor
             Application existing = ProxyUtil.convertToObject(json, Application.class);
             verifySchemaRichApp(application, existing);
             prepareApplicationFunction(resource, application, existing);
+            prepareAdminManagedFields(application, existing, preserveForwardAuthToken);
```
astsiapanay: Wrong flag is passed to the method

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
```diff
@@ -103,41 +149,45 @@ public Future<?> putExternalService(String appId, String serviceId) {
...
 
-    private ResolvedApp resolveAndAuthorize(String appId) {
+    private void purgeCredentials(String appId, String serviceId, CredentialsLevel level) {
+        String scopeId = CredentialsLocatorFactory.APPLICATIONS_PREFIX + appId
+                + CredentialsLocatorFactory.EXTERNAL_SERVICES_SEPARATOR + serviceId;
+        CredentialsLocator locator = CredentialsLocatorFactory.fromExternalServiceScope(scopeId, context);
+        resourceCredentialsService.deleteResourceCredentialsAtLevel(locator, level);
+    }
+
+    private ResolvedApp resolveApp(String appId) {
```
astsiapanay, line 268: **Missing access-control gate (CONFIRMED)**

`resolveApp()` dropped the old `resolveAndAuthorize()`'s admin-or-write-access check (`hasAdminAccess(context) || hasWriteAccess(descriptor, context)`), which used to be enforced for *every* management operation on *every* app. Now it only loads the `Application`, with no ACL check.

`canManageInline()` still gates the admin/inline branch, but the new user-authored branch is gated only by `requireUserAuthoringAllowed()` (line 206), which checks solely `isAllowUserExternalServices()` — never any read/write/admin grant on the target app.

**Failure scenario:** User B, with zero access to User A's app `secret-app` (which has `allow_user_external_services=true`), can call `GET/PUT/DELETE /v1/applications/secret-app/external-services/{id}` and manage their own external-service resources scoped to that app. It also creates a 403-vs-404-vs-200 oracle revealing an arbitrary app's existence and its `allow_user_external_services` flag to callers with no access to it — something the old uniform pre-lookup 403 prevented.

Worth double-checking whether requiring at least *read* access to the target application before allowing user-authored service management was intended.

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -192,6 +196,22 @@ public Pair<ResourceItemMetadata, Application> putApplication(ResourceDescriptor
         return Pair.of(meta, application);
     }
 
+    // app_identity and allow_user_external_services are admin-managed. AUTHORITATIVE writes (config file /
+    // admin-apply) honor the incoming values; every other path inherits the stored values on update (so a
+    // read-modify-write can't wipe them) and strips on create.
+    private static void prepareAdminManagedFields(Application application, Application existing, AdminManagedFieldsWriteMode mode) {
```
astsiapanay, line 202: **`copyApplication()` bypasses `AdminManagedFieldsWriteMode` entirely (CONFIRMED)**

`prepareAdminManagedFields` (this method) is only invoked from `putApplication`. `copyApplication` (same file, ~line 303) — used by `ResourceOperationService.copyApplication/moveApplication` (plain user-triggered resource copy/move) and `PublicationService`'s review-copy step — writes the destination `Application` straight through `resourceService.computeResource` with no call to this method at all.

**Failure scenario:** a user with read access to any application that has `allow_user_external_services=true` (e.g. a public app) can use the ordinary resource-copy API to copy it into their own bucket. The copy inherits `allow_user_external_services=true` (and any `app_identity`) verbatim — self-granting the 'regular users may author external services' governance flag on an app they now fully own, entirely outside the `AUTHORITATIVE`/`INHERIT_ONLY` gate this PR introduces.

Suggest threading the same admin-managed-field stripping/inheriting into `copyApplication`, or centralizing it at the resource-write layer both methods funnel through.

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -0,0 +1,49 @@
...
+            case null -> "SUCCESS";
+            case PermissionDeniedException ignored -> "DENIED";
+            case ResourceNotFoundException ignored -> "NOT_FOUND";
+            default -> "ERROR";
+        };
+        // reason echoes only the exception message, never a response body or secret — keep it that way.
+        AUDIT.info("event=obo_credential_retrieval outcome={} actor={} owner_sub={} application_id={} "
+                        + "external_service_id={} trace_id={}{}",
+                outcome, actorEvidence(context), ownerSub, applicationId, externalServiceId, context.getTraceId(),
+                error == null ? "" : " reason=\"%s\"".formatted(error.getMessage()));
```
astsiapanay: **Audit log injection via unsanitized content (PLAUSIBLE)**

`OboCredentialsRequest.ownerSub` is validated only with `@NotBlank` (no charset/pattern restriction), and `error.getMessage()` can embed caller-supplied content — e.g. `ResourceNotFoundException`'s message embeds `request.getUrl()` verbatim via `"Credentials for %s not found"`. Both are interpolated raw into this `AUDIT.info(...)` call.

**Failure scenario:** a caller sends an `owner_sub` or `url` containing `\n` plus fabricated `outcome=SUCCESS ...` text. The resulting audit line contains embedded newlines that render as extra, forged-looking `DIAL_OBO_AUDIT` entries to any log viewer/SIEM that parses this stream line-by-line — undermining the audit trail's integrity for exactly the abuse-detection purpose it exists for. Worth stripping/escaping control characters from `ownerSub` and `error.getMessage()` before logging.

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    // Non-secret evidence of the calling actor: the DIAL key's project, or the workload JWT's azp.
+    private static String actorEvidence(ProxyContext context) {
```
astsiapanay, line 53: **Actor misattribution in the audit trail (PLAUSIBLE)**

`AppIdentityMatcher.matches()` checks the key-hash match first and only falls through to the `azp` claim comparison when a key is present but its hash doesn't match `app_identity` (or no key is present). If a request carries both a `Key` and `ExtractedClaims` and the gate passes via `azp` (key hash mismatched), `actorEvidence()` still short-circuits on `key != null` and logs `project:<project>` — the project of a key that did **not** satisfy the gate — instead of `azp:<azp>`, the identity that actually authorized the call.

**Failure scenario:** an operator investigating a suspicious OBO retrieval via the audit log attributes it to the wrong actor identity, since the logged evidence doesn't necessarily correspond to the identity that actually passed `AppIdentityMatcher.matches()`.

astsiapanay (commented):
Automated review findings (high-effort pass across 8 finder angles + direct verification against the PR head commit). Ranked most-severe first.

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final Logger AUDIT = LoggerFactory.getLogger("DIAL_OBO_AUDIT");
+
+    private ExternalServiceAuditLog() {
+    }
+
+    /** One event per OBO retrieval outcome; {@code error} is {@code null} on success. */
+    public static void oboRetrieval(ProxyContext context, String applicationId, String externalServiceId,
+                                    String ownerSub, RuntimeException error) {
+        String outcome = switch (error) {
```
astsiapanay, line 36: **Outcome classification collapses two different denial reasons (PLAUSIBLE)**

An `app_identity` gate mismatch and a stored-but-non-consented credential both throw `PermissionDeniedException` in `getOboCredentials`, so both map to `outcome=DENIED` here.

**Failure scenario:** an operator monitoring `DIAL_OBO_AUDIT` for identity-spoofing attempts (security-relevant) can't distinguish them from routine 'owner hasn't consented yet' denials (expected/frequent) without parsing the free-text `reason` field — reducing the dedicated audit stream's value for its stated abuse-detection purpose. Consider a distinct outcome (e.g. `IDENTITY_MISMATCH` vs `CONSENT_REQUIRED`).

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
```diff
@@ -223,6 +225,52 @@ public ResourceCredentials getRefreshedResourceCredentials(CredentialsLocator cr
...
+                || !Objects.equals(ownerSub, stored.getUserId())) {
+            return null;
+        }
+        // Gate consent BEFORE refreshing: an OBO probe against a non-consented credential must not rotate the
+        // owner's refresh token. Return the stored (un-refreshed) credential so the caller can reject on consent.
+        if (!stored.isOfflineUsageConsent()) {
+            return stored;
+        }
+
+        ResourceCredentials userCredentials = getAndRefreshCredentials(userDescriptor, authSettings);
```
astsiapanay, line 264: **Narrow TOCTOU on consent revocation (PLAUSIBLE)**

Consent is checked (line 260) before calling `getAndRefreshCredentials` here, but there's no re-check after the refresh completes. `getOboCredentials` only inspects `credentials.isOfflineUsageConsent()` on the object returned after refresh, which reflects the pre-refresh state read at line 260.

**Failure scenario:** if the owner revokes offline-usage consent (e.g. via sign-out) between the consent check and this refresh call completing, the freshly rotated token is still handed to the OBO caller — unlike the pre-refresh gate, which is explicitly commented as intentional to avoid rotating a non-consented credential's refresh token. Low likelihood, but worth an explicit comment if this is an accepted risk.

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
```diff
@@ -103,41 +149,45 @@ public Future<?> putExternalService(String appId, String serviceId) {
...
+            // Cascade: the author owns both the definition and its USER-level credentials, so purge them too.
+            purgeCredentials(appId, serviceId, CredentialsLevel.USER);
             return true;
         }).onSuccess(removed -> context.respond(HttpStatus.OK, removed))
                 .onFailure(error -> respondError("Can't delete external service", error));
         return Future.succeededFuture();
     }
 
-    private ResolvedApp resolveAndAuthorize(String appId) {
+    private void purgeCredentials(String appId, String serviceId, CredentialsLevel level) {
```
astsiapanay, line 256: **Possible double URL-decoding when rebuilding scope ids (PLAUSIBLE)**

`ControllerSelector` decodes the `appId`/`serviceId` path segments via `UrlUtil.decodePath` before invoking the controller. This method concatenates those already-decoded values raw into a scope string and passes it to `CredentialsLocatorFactory.fromExternalServiceScope`, whose `parseExternalServiceScope` calls `UrlUtil.decodePath` on the whole string again.

**Failure scenario:** an `appId`/`serviceId` that, after the first decode, contains a literal `%` followed by two hex digits gets corrupted by the second decode pass, causing the USER-level credential purge on delete (newly added for the user-authored cascade) to target the wrong resource id and silently fail to purge the author's credentials. This pattern predates the PR but is now exercised by the new USER-level cascade path too.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
DmytroZaichenkoDev: Fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
```diff
@@ -84,12 +88,10 @@ public ExternalServiceManagementController(Proxy proxy, ProxyContext context) {
     )
     public Future<?> listExternalServices(String appId) {
         taskExecutor.submit(() -> {
-            ResolvedApp resolved = resolveAndAuthorize(appId);
+            ResolvedApp resolved = resolveApp(appId);
```
astsiapanay, line 91: Please OpenAPI annotations

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
DmytroZaichenkoDev: Fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1678, opened 2026-07-02
feat: add POST /v1/ops/toolset/{bucket}/{path}/repair for DCR OAuth recovery

### LINKED ISSUE #1547: Auto re-register OAuth client when Authorization Server reports invalid_client for an MCP toolset
### Name and Version
0.43.1

When a toolset is created for an MCP server that uses OAuth, DIAL performs Dynamic Client Registration (RFC 7591) against the MCP's Authorization Server and persists the resulting `client_id` / `client_secret` alongside the toolset.

If the Authorization Server later drops that registration — e.g. after an AS restart with non-persistent storage, a TTL expiry on dynamically-registered clients, or an operator-initiated cleanup — every subsequent OAuth login attempt for that toolset fails with `invalid_client` / "client not found". The toolset becomes permanently unusable until it is deleted and re-created manually, even though nothing on the DIAL side has changed.

Detect `invalid_client` (and equivalent "unknown client" responses) from the Authorization Server during the OAuth flow for an MCP toolset, and automatically re-run Dynamic Client Registration against the same AS, replacing the stored credentials on the toolset. The user-facing login flow should then transparently retry with the new credentials so the toolset keeps working without manual intervention.

Scope:
- Applies to toolsets whose OAuth client was originally provisioned via DCR by DIAL.
- Triggered on `invalid_client` from the token and/or authorization endpoints.
- Re-registration uses the same metadata DIAL used on first registration.
- Old credentials are overwritten atomically on the toolset record.

Surfacing the error to the user and requiring manual toolset re-creation (current behavior) — rejected because it forces every end-user of the toolset to lose their integration whenever the AS cycles its client store.

### LINKED ISSUE #1581: Toolsets auth issue // auth path change
### Name and Version

0.43.2

If auth path for underlying MCP gets changed, then toolset stops working. This is required to either re-create toolset or find a way to update existing one triggering new path re-identification

_No response_

### What do you see instead?

Observe error 'Invalid token'

### Additional information

_No response_

karatkep:
## Summary

- Adds `POST /v1/ops/toolset/{bucket}/{path}/repair` endpoint that recovers broken OAuth state for dynamically-registered toolsets without requiring delete-and-recreate
- Fixes #1547: AS drops DCR client → `invalid_client` on authorize
- Fixes #1581: AS moves endpoints → stale discovery data

## Problem

DIAL stores OAuth discovery + DCR results at toolset-create time and never re-validates. When the AS drops the client or moves endpoints, auth fails at the browser→AS `/authorize` redirect — which Core never sees and cannot auto-fix. Recovery today requires deleting and recreating the toolset, which loses all existing credentials and configuration.

## Algorithm

1. **Validate** toolset is eligible: `dynamicallyRegistered == true` (static and legacy toolsets → 400)
2. **Pre-flight AS check**: discover AS metadata — if unreachable, return 424 Failed Dependency and abort (no DCR, no changes to authSettings). Signals that the AS is the failing dependency, not DIAL Core.
3. **Re-register**: issue a new DCR request unconditionally, store new `client_id`/`client_secret` and fresh endpoint URLs
4. **Clear credentials**: delete GLOBAL credentials and the calling user's personal credentials. Other users' personal credentials are not reachable by Core — they will fail at the AS on next use since they carry the old `client_id`.

## Key design decisions

- **Always re-registers**: USER-level credentials of other users live in private buckets and cannot be inspected. Since the caller invokes repair explicitly because something is broken, re-registration is always the safe choice. Worst case (client was actually alive): the caller and GLOBAL credentials are cleared immediately; other users' personal credentials will fail at the AS on next use.
- **Access control**: `hasWriteAccess()` — owner, admin on public/shared toolsets, and users with shared WRITE permission. Admin does **not** get access to other users' private toolsets.
- **Credential cleanup failure is non-fatal**: stale tokens are already invalid at the AS (wrong `client_id`); users will be prompted to re-sign-in regardless. A 500 here would cause callers to retry and orphan additional client registrations at the AS.
- **Atomicity**: toolset mutation uses `ResourceService.computeResource()` which holds its own distributed lock. No outer lock needed (adding one deadlocks against `computeResource`'s internal lock).
- **Eligibility**: `dynamicallyRegistered == true` only. `null` (legacy) and `false` (static) → 400. Prevents DIAL from re-registering operator-supplied static clients.

## Known limitations

- Legacy toolsets (`dynamicallyRegistered == null`) are not eligible — forward-only by design; clients must re-create toolsets to receive the flag
- Concurrent repairs on the same toolset: the last writer wins; the earlier registration is orphaned at the AS (minor resource leak, toolset remains functional)

## Test plan

- [x] `testRepairForbiddenWhenNoWriteAccess` — 403 when user has no write access
- [x] `testRepairForbiddenForAdminOnPrivateToolset` — 403 when admin tries to repair another user's private toolset
- [x] `testRepairAllowedForOwner` — 200 for toolset owner
- [x] `testRepairAllowedForAdminOnPublicToolset` — 200 for admin on published toolset
- [x] `testRepairAllowedForSharedWriteUser` — 200 for user with shared WRITE permission
- [x] `testRepairNotFoundForMissingToolset` — 404
- [x] `testRepairRejectedForStaticOauthToolset` — 400 (`dynamicallyRegistered=false`)
- [x] `testRepairRejectedForNonOauthToolset` — 400 (API_KEY toolset)
- [x] `testRepairSucceeds` — 200 `REREGISTERED` for valid DCR toolset
- [x] `testRepairClearsAllLevelsOnReregistration` — GLOBAL + USER both `SIGNED_OUT` after repair
- [x] `testRepairFailsWithFailedDependencyWhenDiscoveryReturnsNotFound` — AS unreachable → 424
- [x] `ToolSetRepairServiceTest.repairSucceedsEvenWhenCredentialCleanupFails` — cleanup exception is swallowed, repair returns successfully


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/data/RouteTemplate.java
```diff
@@ -218,6 +218,11 @@ public enum RouteTemplate {
         "/v1/ops/toolset/{operation}"
     ),
 
+    TOOL_SET_REPAIR(
```
astsiapanay, line 221: The endpoint could conflict with 
```
RESOURCE(
            "^/v1/(conversations|prompts|applications|toolsets)/(?<bucket>[a-zA-Z0-9]+)/(?<path>.*)$",
            "/v1/{resourceType}/{bucket}/{path}"
    ),
```

server/src/main/java/com/epam/aidial/core/server/service/ToolSetRepairService.java
```diff
@@ -0,0 +1,99 @@
...
+        // computeResource holds the distributed lock internally; an outer lock would deadlock against it.
+        resourceService.computeResource(resource, EtagHeader.ANY, author, json -> {
+            ToolSet stored = ProxyUtil.convertToObject(json, ToolSet.class);
+            if (stored == null) {
+                throw new ResourceNotFoundException("ToolSet is not found: " + resourceUrl);
+            }
+            ResourceAuthSettings storedSettings = stored.getAuthSettings();
+            authSettingsService.applyRegistration(storedSettings, registration);
+            storedSettings.setDynamicallyRegistered(true);
+            encryptionService.encrypt(resourceUrl, bucketInfo, storedSettings);
```
DmytroZaichenkoDev, line 87: double encryption here? we've just read settings from storage, where secrets already encrypted. worth adding test.

server/src/main/java/com/epam/aidial/core/server/service/ToolSetRepairService.java
```diff
@@ -0,0 +1,99 @@
...
+                    "ToolSet " + resourceUrl + " is not eligible for repair (dynamicallyRegistered != true)");
+        }
+
+        encryptionService.decrypt(resourceUrl, bucketInfo, authSettings);
+
+        // Pre-flight 502 guard: register() throws IllegalArgumentException (not HttpException) when AS
+        // is unreachable, which would surface as 400. This probe maps AS unavailability to 502 instead.
+        log.info("Repair: discovering AS metadata for toolset={}", resourceUrl);
+        if (registrationService.discoverMetadata(resourceUrl, toolSet.getEndpoint()) == null) {
+            throw new HttpException(HttpStatus.BAD_GATEWAY,
```
DmytroZaichenkoDev: I'm not sure 502 is correct choice here, since for dial-core client it looks like dial is unavailable, but it's not the case here.

server/src/main/java/com/epam/aidial/core/server/data/RouteTemplate.java
karatkep: Good catch, updated to
```
        "^/v1/ops/toolset/(?<bucket>[a-zA-Z0-9]+)/(?<path>.+)/repair$",
        "/v1/ops/toolset/{bucket}/{path}/repair"
```

server/src/main/java/com/epam/aidial/core/server/service/ToolSetRepairService.java
karatkep: Thank you, added
```
encryptionService.decrypt(resourceUrl, bucketInfo, storedSettings);
```

The same approach as for [ToolSetService.java](https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/service/ToolSetService.java#L103)

server/src/main/java/com/epam/aidial/core/server/service/ToolSetRepairService.java
karatkep: Thank you. Changed to **424 Failed Dependency** - this signals that the operation failed because a dependency (AS) is unavailable, not that DIAL Core itself is down.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

DmytroZaichenkoDev (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1683, opened 2026-07-02
fix: release local lock in LockService when Redis lock acquisition fails

siarhei-fedziukovich:
### Description of changes

A Redis failure thrown from the lock-script `EVAL` (e.g. `WriteRedisConnectionException` on a Valkey connection blip) escaped `LockService.lock()` **after** the per-key JVM-local `ReentrantLock` was already held, orphaning that lock permanently — it has no TTL and its owner thread dies with the exception. Every subsequent `lock()` on the same key then blocks forever.

Observed in a live environment (0.46.0-dev.14, 2 pods): one transient Valkey blip wedged the admin bucket locks on both pods simultaneously. From that moment every debounced merged-config rebuild (~1 per 60s file poll), `/v1/ops/config/reload`, and admin config write parked in the queue — a thread dump showed 377 blocked virtual threads and exactly one `Failed to rebuild merged config` warning per pod at the wedge start, then silence. Only a pod restart recovers.

The fix wraps the distributed spin-lock acquisition in try/catch and releases the local lock (and its refcount entry) before rethrowing, so a transient Redis error fails the single operation instead of permanently wedging the pod.

Regression test `LockServiceFailureTest` simulates the `EVAL` throwing once and asserts the exception propagates while the key remains acquirable by another thread; it reproduces the production hang (times out) without the fix.

`:storage:test`, `:server:test`, and checkstyle pass.

### Checklist

### DIFF
--- a/storage/src/main/java/com/epam/aidial/core/storage/service/LockService.java
+++ b/storage/src/main/java/com/epam/aidial/core/storage/service/LockService.java
@@ -62,13 +62,21 @@ public class LockService {
         LocalLock localLock = acquireLocalLock(id);
         localLock.lock();
 
-        long ttl = tryLock(id, owner);
-        long interval = WAIT_MIN;
-        // it seems the lock has been acquired by another instance of Core
-        while (ttl > 0) {
-            LockSupport.parkNanos(interval);
-            interval = Math.min(2 * interval, Math.min(WAIT_MAX, ttl + 1));
-            ttl = tryLock(id, owner);
+        // A Redis failure here must not leak the held local lock: it is a plain ReentrantLock,
+        // so a leaked hold would block every subsequent lock() on this key until pod restart.
+        try {
+            long ttl = tryLock(id, owner);
+            long interval = WAIT_MIN;
+            // it seems the lock has been acquired by another instance of Core
+            while (ttl > 0) {
+                LockSupport.parkNanos(interval);
+                interval = Math.min(2 * interval, Math.min(WAIT_MAX, ttl + 1));
+                ttl = tryLock(id, owner);
+            }
+        } catch (Throwable e) {
+            localLock.unlock();
+            releaseLocalLock(id);
+            throw e;
         }
 
         return () -> unlock(id, owner, localLock);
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

siarhei-fedziukovich:
Troubleshooting report: /v1/ops/config/reload hangs

Symptom: reload never returns (60s → 5min client timeouts), on both pods, every attempt; admin config PUTs also hang, everything else fine.

Root cause: a one-second Valkey connection blip (2026-07-02 16:13:00) made the lock-script EVAL throw inside LockService.lock() after the per-key JVM-local ReentrantLock was already held. That local lock has no TTL and was never released — every subsequent rebuild/reload/admin-write on the bucket key queued behind it forever (377 parked virtual threads ≈ uptime ÷ 60s poll). The unified-config change (v0.45+) made these locks be taken every 60s on every pod, turning the latent bug into a guaranteed permanent wedge.

Diagnosis steps (reusable runbook)

1. Scope with API probes (admin credential, curl -m 30 -w '%{http_code} %{time_total}s'):
  - GET /v1/admin/health/config, GET /v1/admin/config/file/models — in-memory, no locks → tests HTTP/auth/event-loop.
  - GET /v1/metadata/{models,roles,keys,...}/platform/ — listing, no locks → tests Redis/blob/executor.
  - PUT /v1/roles/platform/probe-test (delete after) — same locks as reload, no scan → tests the lock layer.
  - Our result: only PUT hung → locks, not the scan.
2. Thread dump with virtual threads — plain jstack omits them; use:
jcmd 1 Thread.dump_to_file -format=plain /tmp/threads.txt, then grep for LockService.lock (waiter count), MergedConfigStore.rebuild( (live rebuild?), and parkNanos + LockService (holder spinning on Redis?).
Our result: 377 waiters, zero live rebuild, zero spinners → holder dead → leaked local lock.
3. Correlate logs — one "Failed to rebuild merged config" (WriteRedisConnectionException on EVAL … lock:dial/platform/) per pod at wedge start, then silence. Timestamp = leak moment.
4. (If a live spinner had been found instead): redis-cli HGETALL lock:dial/platform/ — a deadline absurdly far in the future means a poisoned key → DEL it.

Oleksii-Klimov (approved):

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1690, opened 2026-07-06
fix: tolerate single-element 'resource' array in OAuth metadata #1682

### LINKED ISSUE #1682: Support single-element array for 'resource' in OAuth Protected Resource Metadata (RFC 9728)
During DCR / auth-metadata discovery, some MCP servers (e.g. GitLab) return the `resource` field of their OAuth 2.0 Protected Resource Metadata document as a single-element JSON array, e.g.:

```json
{
  "resource": ["https://gitlab.com"],
  "authorization_servers": ["https://gitlab.com"]
}
```

## Spec

[RFC 9728](https://www.rfc-editor.org/rfc/rfc9728.html) §2 defines `resource` as **REQUIRED** and a **single JSON string** (a URL). §3.2 states only multi-valued parameters are represented as arrays — `resource` is single-valued (unlike `authorization_servers`, which is an array). The MCP authorization spec references RFC 9728, so it inherits the same rule.

So the array form is **not spec-compliant**. GitLab is technically wrong here.

## Problem

DIAL Core parses the discovered metadata into a model whose `resource` is a plain string, so an array value fails deserialization and the entire discovery / DCR flow aborts. This makes OAuth / toolset creation against GitLab (and any other server that emits `resource` as an array) fail outright.

## Proposal

For interoperability (Postel's law), DIAL Core should tolerate a **single-element** array for the `resource` field on the client/parsing side and treat it as the equivalent scalar string. Genuine multi-element arrays should still be rejected, since the spec permits only one value. This is inbound-parsing only; the metadata DIAL itself publishes must remain a single string per spec.

## Repro

1. Start toolset creation / OAuth discovery against an MCP server whose `.well-known/oauth-protected-resource` returns `"resource"` as a single-element array (e.g. GitLab).
2. Discovery fails to parse the metadata and the flow aborts.

DmytroZaichenkoDev:
Some MCP servers (e.g. GitLab) return the `resource` field of their OAuth 2.0 Protected Resource Metadata as a single-element JSON array rather than the spec-required single string (RFC 9728), which fails deserialization and aborts the entire OAuth discovery / DCR flow. This tolerates a single-element array on the inbound-parsing side, treating it as the equivalent scalar; multi-element arrays are still rejected, and the metadata DIAL publishes is unchanged.

### Applicable issues

- fixes #1682

### Description of changes

- Accept a single-element `resource` array during Protected Resource Metadata parsing and unwrap it to the scalar string.
- Add unit and integration test coverage for the discovery flow.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1666, opened 2026-06-30
feat: collect request claims and headers into analytics logs

Pasichniuk:
### Applicable issues

- no applicable issue

### Description of changes

- Optional claims collection — When log.collectClaims is enabled, each analytics log entry gains a claims object (user_id, roles, project, user_hash, user_display_name), with null fields skipped so the JSON stays valid.
- Optional header collection — When log.collectHeaders is enabled, a headers object with the request headers is added to each log entry; headers matching the case-insensitive log.headersBlacklist (default authorization,api-key,cookie,proxy-authorization) are excluded, multi-value headers are joined, and oversized values are truncated with a >> marker.
- Configuration & wiring — Added the log.collectClaims / log.collectHeaders / log.headersBlacklist settings (defaults in aidial.settings.json, documented in the README), parsed in AiDial and passed into GfLogStore via its new constructor.
- Tests — Added GfLogStoreTest coverage for claims serialization, null-field skipping, header blacklisting/multi-value joining, and that claims+headers compose into valid log JSON.

### Checklist

- [x] Title of the pull request follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1701, opened 2026-07-09
feat: add regex-capable header allowlist for analytics logging

siarhei-fedziukovich:
### Applicable issues

- no applicable issue

### Description of changes

Follow-up to #1666. Header collection (`analytics.collectHeaders`) previously supported only an exact-name blocklist, which produced noisy output (`X-Stainless-*` SDK telemetry, transport headers like `Accept`/`Connection`/`Content-Length`) and gave no way to select just the useful headers (e.g. `traceparent` for OTel).

- **Regex matching** — every entry in `analytics.headersBlacklist` / `analytics.headersAllowlist` is now a case-insensitive, full-match regex (mirrors the `Route.paths` idiom). Literal names like `authorization` keep working since they match themselves; patterns like `x-stainless-.*` drop a whole family.
- **Optional allowlist** — new `analytics.headersAllowlist`: when present, only headers matching it are collected; the blacklist still wins over the allowlist. Omitting the key disables the allowlist (collect all non-blocked headers) — default behavior unchanged.
- **JSON array config** — `headersBlacklist` / `headersAllowlist` are now JSON arrays instead of a comma-separated string (a comma collides with regex quantifiers like `{1,3}`). Defaults in `aidial.settings.json` updated; README documents both keys with a recommended OTel allowlist example.
- **Tests** — extended `GfLogStoreTest` with regex-family-drop, allowlist-only, and blocklist-wins-over-allowlist cases.

> [!NOTE]
> Breaking config change: any deployment using a comma-separated `analytics.headersBlacklist` string must migrate to a JSON array.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1702, opened 2026-07-09
fix: follow MCP proxy 307/308 redirects instead of relaying them #1667

### LINKED ISSUE #1667: MCP proxy relays upstream /mcp -> /mcp/ redirect to caller instead of following it
### Description

When the MCP proxy forwards a request to an upstream MCP server and that server answers `POST /mcp` with a redirect to `/mcp/` (HTTP `307`/`308`), DIAL Core relays the redirect response **verbatim** to the caller instead of following it. The caller (e.g. QuickApp) receives a `3xx` + `Location` header for what should be an MCP response and doesn't know how to handle it, so the MCP call fails.

Many MCP server frameworks (Starlette/FastMCP and similar) perform automatic trailing-slash redirects: a request to `/mcp` is answered with a `307 Temporary Redirect` to `/mcp/`. DIAL Core should follow this redirect itself rather than surfacing it to the client.

### Steps to reproduce

1. Configure an MCP application/toolset whose upstream endpoint is served by a framework that trailing-slash-redirects (`/mcp` -> `/mcp/`, e.g. Starlette/FastMCP).
2. Issue an MCP call through DIAL Core (`POST /v1/deployments/{id}/mcp` or `POST /v1/toolset/{id}/mcp`).
3. Observe DIAL Core returns the upstream's `307`/`308` + `Location: .../mcp/` to the client instead of the MCP response.

### Expected behavior

DIAL Core follows the upstream redirect (re-issuing the request, preserving method and body for `307`/`308`) and returns the final MCP response to the caller. The redirect should not leak to the client.

### Suggested considerations for the fix

- Follow `307`/`308` (which preserve method + body — the path-normalization case). Be cautious with `301`/`302`/`303`, where standard semantics downgrade a `POST` to `GET`.
- Cap the number of redirects followed to avoid loops.
- Restrict automatic following to the same origin (scheme/host/port), or otherwise avoid forwarding the `Authorization` header to a different host.

astsiapanay:
DIAL Core's MCP proxy was relaying upstream `307`/`308` trailing-slash redirects (e.g. Starlette/FastMCP redirecting `/mcp` -> `/mcp/`) verbatim to the caller instead of following them, breaking MCP calls. Core now follows same-origin `307`/`308` redirects internally, capped, preserving method and body; other redirect codes (`301`/`302`/`303`) continue to be relayed unchanged.

### Applicable issues

- fixes #1667

### Description of changes

- `McpProxyController`: added `tryFollowRedirect` which, on a `307`/`308` response, resolves the `Location` header against the current request URI, verifies it's same-origin (scheme/host/port) as the upstream endpoint, and internally re-issues the request (same method + buffered body) up to 5 hops. Falls back to the existing verbatim relay when the `Location` is missing/malformed, cross-origin, or the redirect cap is hit — so `Authorization`/API-key headers are never sent to a different host.
- `sendRequest()` was split to accept an explicit absolute URI so redirected requests can target the `Location` instead of the original upstream endpoint.
- Added tests in `ToolSetApiTest` covering: following `307`/`308`, not following `302`, refusing cross-origin redirects, and capping redirect loops.

### Checklist


server/src/main/java/com/epam/aidial/core/server/controller/McpProxyController.java
```diff
@@ -289,6 +302,54 @@ private boolean canRetry(UpstreamRoute route) {
...
+        if (location == null || redirectCount >= MAX_MCP_REDIRECTS) {
+            return false;
+        }
+
+        URI currentUri;
+        URI redirectUri;
+        try {
+            currentUri = new URI(context.getProxyRequest().absoluteURI());
+            redirectUri = currentUri.resolve(location);
+        } catch (URISyntaxException e) {
```
DmytroZaichenkoDev, line 326: **Bug: a malformed `Location` escapes this catch and the client request hangs.**

`URI.resolve(String)` is equivalent to `resolve(URI.create(str))`, and `URI.create` wraps the `URISyntaxException` in an unchecked `IllegalArgumentException` — verified on JDK 21:

```java
URI.create("http://localhost:9876/mcp").resolve("http://[malformed")
// throws java.lang.IllegalArgumentException, not URISyntaxException
```

So an upstream answering `307` with a garbage `Location` (e.g. `http://[bad`) bypasses this catch, the exception propagates out of the Vert.x `onSuccess` handler, and the client gets no response at all (hangs until timeout) instead of the documented fallback of relaying the redirect verbatim. Since `Location` is upstream-controlled input, this should be fixed before merge.

Two clean options:
- `catch (URISyntaxException | IllegalArgumentException e)`, or
- resolve via `currentUri.resolve(new URI(location))` so both statements throw the checked `URISyntaxException`.

A companion test with a malformed `Location` asserting the `307` is relayed as-is would lock this in.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/McpProxyController.java
DmytroZaichenkoDev: Confirmed at runtime against this branch (head `d439a9b`), not just from the JDK docs. Repro test added locally:

```java
TestWebServer.Handler handler = request ->
        new MockResponse().setResponseCode(307).setHeader("Location", "http://[malformed");
```

Result: the client never receives a response (read timeout after 10s), and the server log shows the escaped exception exactly at this line:

```
Unhandled exception (io.vertx.core.impl.ContextImpl)
java.lang.IllegalArgumentException: Expected closing bracket for IPv6 address at index 17: http://[malformed
	at java.net.URI.create(URI.java:932)
	at java.net.URI.resolve(URI.java:1122)
	at McpProxyController.tryFollowRedirect(McpProxyController.java:322)
	at McpProxyController.handleProxyResponse(McpProxyController.java:243)
```

The PR's own redirect tests (follow-307, cross-origin refusal, loop cap) pass in the same environment, so this is the only gap.

astsiapanay: fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/McpProxyController.java
DmytroZaichenkoDev: Verified on `9affaa6f`: my independent repro (which previously hung for the full 10s client timeout) now passes in ~0.3s — the malformed `Location` is relayed as a verbatim `307` and the server log is clean. All 7 redirect tests green locally. Thanks for the quick fix — resolving.

DmytroZaichenkoDev (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [29027580617](https://github.com/epam/ai-dial-ci/actions/runs/29027580617)

| [deploy-review](https://chat-ai-dial-core-pr-1702.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29027580617/job/86151340194) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/29027580617/job/86154980117) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

Issue #1713, opened 2026-07-14
Upstreams might be set to null in ApplicationController


astsiapanay:
### Name and Version

core 0.45

### What steps will reproduce the bug?

1. List or get applications from DIAL Config
2. Run chat completion request or application route

### What do you see instead?

The error appears in logs

```
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because "upstreams" is null
	at com.epam.aidial.core.server.upstream.TieredBalancer.buildTiers(TieredBalancer.java:126)
	at com.epam.aidial.core.server.upstream.TieredBalancer.<init>(TieredBalancer.java:33)
	at com.epam.aidial.core.server.upstream.UpstreamRouteProvider.lambda$get$1(UpstreamRouteProvider.java:113)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
	at com.epam.aidial.core.server.upstream.UpstreamRouteProvider.get(UpstreamRouteProvider.java:107)
	at com.epam.aidial.core.server.upstream.UpstreamRouteProvider.get(UpstreamRouteProvider.java:103)
	at com.epam.aidial.core.server.controller.route.BaseRouteController.lambda$handleRoute$2(BaseRouteController.java:75)
	at io.vertx.core.impl.future.Composition.onSuccess(Composition.java:38)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.SucceededFuture.addListener(SucceededFuture.java:88)
	at io.vertx.core.impl.future.FutureBase.compose(FutureBase.java:90)
	at io.vertx.core.impl.future.SucceededFuture.compose(SucceededFuture.java:27)
	at io.vertx.core.Future.compose(Future.java:374)
	at com.epam.aidial.core.server.controller.route.BaseRouteController.handleRoute(BaseRouteController.java:67)
	at io.vertx.core.impl.future.Composition.onSuccess(Composition.java:38)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.Mapping.onSuccess(Mapping.java:40)
	at io.vertx.core.impl.future.FutureBase.lambda$emitSuccess$0(FutureBase.java:60)
[...]

PR #1718, opened 2026-07-14
fix: upstreams might be set to null in ApplicationController #1713

astsiapanay:


Oleksii-Klimov (approved):

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/29337338274)

PR #1723, opened 2026-07-15
fix: handle schema loading errors

### LINKED ISSUE #1722: GET /openai/applications fails when one schema-rich application has an invalid schema endpoint
### Name and Version

0.45.2

* Add a schema-rich application with an invalid schema endpoint, for example - http://non-existent/application-type-schema
* Call: GET /openai/applications
* Observe the response

_No response_

### What do you see instead?

500: Internal error

### Additional information

A single incorrectly configured schema-rich application can break the entire application listing endpoint: GET /openai/applications.

The issue was observed when an application was configured with a schema endpoint that does not exist or is unreachable. Instead of handling the failure for that specific application, DIAL Core fails the whole application listing request, preventing clients from receiving the list of available applications.

The endpoint should be resilient to misconfiguration of an individual application. One broken application configuration should not prevent other valid applications from being returned.

Oleksii-Klimov:


ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/29408842862)

server/src/test/java/com/epam/aidial/core/server/CustomApplicationApiTest.java
```diff
@@ -1534,20 +1534,7 @@ void testOpenAiApplicationWithTypeSchemaGet_ReturnedInvalid_WhenAppDoesNotConfor
...
-                         "maxRetryAttempts" : 1,
-                         "order" : 5,
-                         "permissions" : [ "WRITE" ],
-                         "attachmentPaths" : {
-                           "requestBody" : [ ],
-                           "responseBody" : [ ]
-                         }
-                       }
-                     }
+                     "routes" : { }
```
astsiapanay, line 1537: Why did you delete routes?

server/src/main/java/com/epam/aidial/core/server/service/ApplicationSchemaService.java
```diff
@@ -465,7 +471,7 @@ public Application modifySchemaRichApplication(Application application, boolean
 
     @Nullable
     @SneakyThrows
-    public Map<String, Route> getRoutes(Application application) {
+    public LinkedHashMap<String, Route> getRoutes(Application application) {
```
astsiapanay, line 474: Why class implementation instead of interface?

Oleksii-Klimov: The Application class uses a LinkedHashMap for routes to maintain their order.

server/src/test/java/com/epam/aidial/core/server/CustomApplicationApiTest.java
Oleksii-Klimov: It’s a failed application. I don’t think we guarantee the correctness of its fields.

astsiapanay (approved):

PR #1724, opened 2026-07-15
feat: sweep/GC reconciliation for complex resources #1639

### LINKED ISSUE #1639: Folder-as-resource: GC / reconciliation sweep (version GC + deleting tombstones)
## Parent

#1633 — DIAL Folder As Resource (EPIC)

## What to build

A periodic sweep scanner that reclaims storage and finishes interrupted deletes. This replaces the parent design doc's original "walk markers / prefix-orphan inference" sketch with a small **reference registry**, so the sweep can enumerate every complex resource in the system without walking every bucket.

- **Reference registry**: on first whole-resource creation, `ComplexResourceService.putFolder` writes a pointer file `complex_resource_refs/{uuid}.json` (`{"url": "<resource url>"}`) **before** committing the `.dial-resource` marker — crash-safe, a resource can never exist without a reference. Written via plain `BlobStorage` calls, not through `ResourceService`/`ResourceDescriptor` (same treatment as the existing `.dial-tmp` / `TEMP_FOLDER` prefix). Not written on a subsequent whole-resource overwrite of an already-existing resource.
- **`deleteFolder` no longer reclaims inline**: it only flips the marker to `state: deleting` (tombstone landed in #1635) and stops — no more best-effort inline `deleteVersion` + `deleteResource` in the request path. The version tree, marker, and reference now live until the sweep reclaims them.
- **`ComplexResourceSweepService`** (new): a periodic job (`TimerService`, same idiom as `ResourceService.sync()` / `cleanupTempFolder`) that walks `complex_resource_refs` in batches. Per reference:
  - marker absent → under the bucket lock, re-confirm absence, then delete the dangling reference (the bucket lock rules out a concurrent in-flight create).
  - marker `state: deleting` and past `gracePeriod` → reclaim (delete the `v/{versionId}` tree + the marker) under the per-resource lock, then delete the reference.
  - marker `active` and its current version past `gracePeriod` → GC any `v/{versionId}` siblings other than `currentVersion`, under the per-resource lock (safety net for `copyOnWrite`'s inline old-version delete when that previously failed).
- **Progress state** in Redis (`complex_resource_scanner_state`): `{ lastRun, nextToken }`. A full pass starts at most once per `fullScanPeriod` (default 24h), gated only when `nextToken` is `null`; mid-pass, each tick just continues from `nextToken`.
- **Concurrency**: cursor *claim* (advancing `nextToken`, brief and exclusive via `LockService.tryLock`) is decoupled from batch *processing* (parallelizable). Processing is bounded by a new `activeBatches` setting enforced **per instance**, via a plain in-process `AtomicInteger` — not a cluster-wide total. Each instance gets its own independent budget, so total sweep concurrency scales with the number of instances (`activeBatches × liveInstances`, roughly) and no single instance can starve the others out of their share. Vert.x's periodic timer fires unconditionally every `period` regardless of whether a previous tick's batch is still processing (same as the existing `sync()`/`cleanupTempFolder` jobs), so an instance can legitimately have more than one batch in flight at once — `activeBatches` is exactly the cap on that.
- **`gracePeriod`**: the reused per-resource/bucket locks already exclude concurrent *writers*; they don't protect an in-flight *reader* (`getFileStream` / `downloadArchive`) that started before a tombstone or version supersession. `gracePeriod` (default 1h, mirrors the existing `TEMP_FILE_TTL` precedent) is the deliberate, time-based guard for that case — checked against `marker.deletedAt` for reclaim, `marker.updatedAt` for version GC.
- New config block `complexResourceSweep`: `period`, `batch`, `fullScanPeriod`, `activeBatches`, `gracePeriod`.
- New metrics: `dial_complex_resource_sweep_batches_total`, `dial_complex_resource_sweep_reclaimed_total`, `dial_complex_resource_sweep_orphan_versions_total`, `dial_complex_resource_sweep_dangling_refs_total`.

## Acceptance criteria

- [ ] A reference file is written to `complex_resource_refs/` on first whole-resource creation only (not on subsequent overwrites of an existing resource), before the `.dial-resource` marker is committed.
- [ ] `deleteFolder` only tombstones the marker; it no longer synchronously deletes the version tree or the marker itself.
- [ ] The sweep reclaims a `deleting` marker's tree + marker, then deletes its reference, only once `gracePeriod` has elapsed since `deletedAt`.
- [ ] The sweep GCs superseded `v/{versionId}` prefixes of an `active` resource (the `copyOnWrite` safety net), only once `gracePeriod` has elapsed since `updatedAt`.
- [ ] The sweep deletes a dangling reference (marker never created) only after re-confirming absence under the bucket lock, so it never races a concurrent create.
- [ ] A full scan runs at most once per `fullScanPeriod`; mid-pass progress survives across ticks via `nextToken`.
- [ ] At most `activeBatches` batches are mid-processing **per instance** at once (not a cluster-wide cap — two instances each configured with `activeBatches=1` run one batch apiece concurrently); cursor claims never hand out overlapping slices regardless of which instance makes them.
- [ ] Tests cover: ref-before-marker crash tolerance, grace-period boundary behavior for both reclaim and version GC, dangling-ref bucket-lock safety, and the per-instance `AtomicInteger` admission behavior under concurrent ticks.

## Blocked by

- #1634
- #1635

astsiapanay:
Adds a periodic sweep scanner that reclaims storage for folder-as-resource complex resources (e.g. skills): it finishes deleting tombstoned resources and GCs orphaned versions left behind by interrupted writes.

### Applicable issues

- fixes #1639

### Description of changes

- Added `ComplexResourceSweepService`, a periodic job (`TimerService`) that walks a new `complex_resource_refs` reference registry in batches, so it can enumerate every complex resource without walking every bucket.
- `ComplexResourceService.put` now writes a `complex_resource_refs/{uuid}.json` pointer before committing the marker on first creation only (crash-safe: a resource can never exist without a reference).
- `ComplexResourceService.delete` no longer reclaims storage inline — it only tombstones the marker (`state: deleting`); the version tree, marker, and reference are reclaimed later by the sweep, once a configurable grace period has elapsed.
- Added package-visible `reclaimDeletingResource`/`gcObsoleteVersions`/`readMarkerForSweep` methods used only by the sweep.
- Fixed `listChildren` (metadata listing) to also exclude a `deleting` marker rather than relying on marker-file presence alone, since the marker can now outlive the DELETE response until swept.
- Cursor advancement over `complex_resource_refs` is a brief, cluster-wide exclusive claim decoupled from per-item processing, which is bounded by an `activeBatches` setting enforced per-instance (a plain in-process counter, not a distributed limit).
- New `complexResourceSweep` config block (`period`, `batch`, `activeBatches`, `gracePeriod`) and 4 Micrometer counters (`dial_complex_resource_sweep_{batches,reclaimed,orphan_versions,dangling_refs}_total`).
- Added unit tests covering reference lifecycle, grace-period gating, dangling-reference cleanup, and per-instance concurrency behavior.
- Updated README with the new settings.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
@@ -687,7 +687,11 @@ public class ComplexResourceService {
      */
     boolean reclaimDeletingResource(ResourceDescriptor resource, long gracePeriodMs) {
         ResourceDescriptor marker = markerDescriptor(resource);
-        try (LockService.Lock ignore = resourceService.lockResource(marker)) {
+        try (LockService.Lock lock = resourceService.tryLockResource(marker)) {
+            if (lock == null) {
+                log.warn("Lock is not acquired to delete inactive resource {}", resource.getDecodedUrl());
+                return false;
+            }
             FolderResourceMarker current = readMarker(marker, false);
             if (current == null || !STATE_DELETING.equals(current.getState())) {
                 return false;
@@ -714,7 +718,10 @@ public class ComplexResourceService {
      */
     boolean gcObsoleteVersions(ResourceDescriptor resource, long gracePeriodMs) {
         ResourceDescriptor marker = markerDescriptor(resource);
-        try (LockService.Lock ignore = resourceService.lockResource(marker)) {
+        try (LockService.Lock lock = resourceService.tryLockResource(marker)) {
+            if (lock == null) {
+                return false;
+            }
             FolderResourceMarker current = readMarker(marker, false);
             if (!isActive(current)) {
                 return false;
--- a/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
@@ -1,9 +1,11 @@
 package com.epam.aidial.core.server.service.resource;
 
+import com.epam.aidial.core.server.data.folder.ComplexResourceRef;
 import com.epam.aidial.core.server.data.folder.FolderResourceMarker;
 import com.epam.aidial.core.server.service.InvitationService;
 import com.epam.aidial.core.server.service.ShareService;
 import com.epam.aidial.core.server.util.ProxyUtil;
+import com.epam.aidial.core.storage.blobstore.BlobStorage;
 import com.epam.aidial.core.storage.blobstore.BlobStorageUtil;
 import com.epam.aidial.core.storage.data.FileMetadata;
 import com.epam.aidial.core.storage.data.MetadataBase;
@@ -21,7 +23,6 @@ import io.vertx.core.buffer.Buffer;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.mutable.MutableObject;
 
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
@@ -55,13 +56,20 @@ public class ComplexResourceService {
 
     public static final String MARKER_NAME = ".dial-resource";
     public static final String FOLDER_MARKER_NAME = ".dial-folder";
+    /**
+     * Root-level (not bucket-scoped) prefix for the sweep-enumeration reference registry, sibling to the
+     * {@code .dial-tmp} temp-folder prefix. Manipulated via raw {@link BlobStorage} calls, not through
+     * {@link ResourceService}, since it is infrastructure bookkeeping rather than an access-controlled resource.
+     */
+    public static final String COMPLEX_RESOURCE_REFS_FOLDER = "complex_resource_refs";
 
     private static final String VERSION_PREFIX = "v";
     private static final String FILES_SEGMENT = "files";
     private static final String FOLDER_TYPE = "folder";
     private static final int SCHEMA_VERSION = 1;
     private static final String STATE_ACTIVE = "active";
-    private static final String STATE_DELETING = "deleting";
+    // Package-visible: ComplexResourceSweepService compares the raw marker state read via readMarkerForSweep.
+    static final String STATE_DELETING = "deleting";
     private static final int PAGE_SIZE = 1000;
     // A file inside a resource must not be named after a structural token.
     private static final Set<String> RESERVED_SEGMENTS = Set.of(VERSION_PREFIX, MARKER_NAME, FOLDER_MARKER_NAME);
@@ -72,13 +80,16 @@ public class ComplexResourceService {
     private final LockService lockService;
     private final ShareService shareService;
     private final InvitationService invitationService;
+    private final BlobStorage blobStorage;
 
     public ComplexResourceService(ResourceService resourceService, LockService lockService,
-                                 ShareService shareService, InvitationService invitationService) {
+                                 ShareService shareService, InvitationService invitationService,
+                                 BlobStorage blobStorage) {
         this.resourceService = resourceService;
         this.lockService = lockService;
         this.shareService = shareService;
         this.invitationService = invitationService;
+        this.blobStorage = blobStorage;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceSweepService.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+
+    @Nullable
+    private ComplexResourceRef loadRef(String path) {
+        Blob blob = blobStorage.load(path);
+        if (blob == null) {
+            return null;
+        }
+        try (InputStream input = blob.getPayload().openStream()) {
```
Oleksii-Klimov, line 176: The payload is Closable too.

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
```diff
@@ -629,11 +663,88 @@ public void deleteFolder(ResourceDescriptor resource, EtagHeader etag) {
...
+     *         (the caller should retry on a later pass)
+     */
+    boolean reclaimDeletingResource(ResourceDescriptor resource, long gracePeriodMs) {
+        ResourceDescriptor marker = markerDescriptor(resource);
+        try (LockService.Lock ignore = resourceService.lockResource(marker)) {
+            FolderResourceMarker current = readMarker(marker, false);
+            if (current == null || !STATE_DELETING.equals(current.getState())) {
+                return false;
+            }
+            long deletedAt = current.getDeletedAt() == null ? 0L : current.getDeletedAt();
```
Oleksii-Klimov, line 699: Should it be defaulted to now?

astsiapanay: I guess no. Otherwise a resource will be never deleted.

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceSweepService.java
astsiapanay: Right it's. The actual resource which needs to be released is a stream itself. We can enclose the payload in the try block instead of the stream if you prefer.
Let me know if that's worth doing.

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
```diff
@@ -629,11 +663,88 @@ public void deleteFolder(ResourceDescriptor resource, EtagHeader etag) {
...
+     * {@code ComplexResourceSweepService}, after the marker has been reclaimed the caller deletes the
+     * reference.
+     *
+     * @return {@code true} if the resource was reclaimed; {@code false} if the marker has since become
+     *         {@code active} again (a recreate raced ahead of the sweep) or is still inside the grace window
+     *         (the caller should retry on a later pass)
+     */
+    boolean reclaimDeletingResource(ResourceDescriptor resource, long gracePeriodMs) {
+        ResourceDescriptor marker = markerDescriptor(resource);
+        try (LockService.Lock ignore = resourceService.lockResource(marker)) {
```
Oleksii-Klimov: Should it rather be a non-blocking tryLockResource?

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
```diff
@@ -629,11 +663,88 @@ public void deleteFolder(ResourceDescriptor resource, EtagHeader etag) {
...
+     * GCs {@code v/{versionId}} siblings of an {@code active} resource other than its current version, once
+     * {@code gracePeriodMs} has elapsed since the marker was last updated. Safety net for {@code copyOnWrite}'s
+     * inline old-version delete when that previously failed. Package-visible: called only by
+     * {@code ComplexResourceSweepService}.
+     *
+     * @return {@code true} if any orphan version was GC'd
+     */
+    boolean gcObsoleteVersions(ResourceDescriptor resource, long gracePeriodMs) {
+        ResourceDescriptor marker = markerDescriptor(resource);
+        try (LockService.Lock ignore = resourceService.lockResource(marker)) {
```
Oleksii-Klimov: tryLockResource?

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
astsiapanay: We may try to lock and the attempt is failed. The resource will be examined in the next run only. So we introduce a delay in this case. At another hand the number of resources to be checked is high. More likely if the resource is active we may get into a situation when the resource is locked. And less likely if resource is marked as deleted it shouldn't be used by anyone.

astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
astsiapanay: fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1727, opened 2026-07-16
fix: allow admins to set app_identity via applications PUT API #1649

### LINKED ISSUE #1649: Headless on-behalf-of credential retrieval for external services (Scheduler support)

DmytroZaichenkoDev:
Admin PUT to the public bucket is now authoritative for the admin-managed application fields (`app_identity`, `allow_user_external_services`), and `app_identity` becomes readable on raw resource GET for admins. This unblocks the Admin UI, which manages public-bucket applications through the plain applications PUT API (Core-direct) and therefore could never grant or display OBO trust with admin/apply as the only authoritative channel.

### Applicable issues

- fixes #1649

### Description of changes

- Replaced the all-or-nothing `AdminManagedFieldsWriteMode` enum with a per-field `AdminManagedFieldsWrite` policy record: config file / admin-apply honor both fields (unchanged), an admin PUT to the public bucket honors exactly the fields **present in the request body**, every other write (non-admin, user-bucket, copy/move, publication) honors neither.
- Presence-based semantics on the admin PUT: an omitted field inherits the stored value, so a read-modify-write client (Admin UI edit, chat UI admin edit) can never wipe a grant; present-as-`null`/`false` is an explicit revoke.
- `app_identity` is returned on raw resource reads to admins only — it is verification material (a key hash or workload `client_id`), not a presentable credential; `client_secret`/`code_verifier` stripping stays unconditional for all roles. Deployment views and non-admin reads keep stripping it.
- User-bucket writes stay strip/inherit even for admins, preserving the invariant that OBO grants can only exist on config apps and public-bucket apps.
- Tests: admin public PUT grant proven end-to-end through the OBO gate; explicit revoke; wipe-safety on field-omitting edits; user-bucket admin PUT still stripped; admin read exposure and non-admin read stripping.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/AdminManagedFieldsWrite.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Per-field write policy for the admin-managed application fields ({@code app_identity},
+ * {@code allow_user_external_services}): {@code honor*} takes the incoming value as authoritative,
+ * otherwise the stored value is inherited on update and stripped on create. Producers:
+ * config file / admin-apply honor both (declarative desired state); an admin PUT to the public
+ * bucket honors exactly the fields present in the request body (so a read-modify-write client that
+ * omits them can never wipe a grant, while present-as-null explicitly clears); every other write —
+ * non-admin PUT, any user-bucket write, copy/move, publication — honors neither.
+ */
+public record AdminManagedFieldsWrite(boolean honorAppIdentity, boolean honorAllowUserExternalServices) {
```
astsiapanay: The class's name should be a noun.

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
```diff
@@ -196,19 +196,14 @@ public Pair<ResourceItemMetadata, Application> putApplication(ResourceDescriptor
...
 
-    // app_identity and allow_user_external_services are admin-managed. AUTHORITATIVE writes (config file /
-    // admin-apply) honor the incoming values; every other path inherits the stored values on update (so a
-    // read-modify-write can't wipe them) and strips on create.
-    private static void prepareAdminManagedFields(Application application, Application existing, AdminManagedFieldsWriteMode mode) {
-        if (mode == AdminManagedFieldsWriteMode.AUTHORITATIVE) {
-            return;
+    // app_identity and allow_user_external_services are admin-managed: a field the policy does not honor is
+    // inherited from the stored value on update (so a read-modify-write can't wipe it) and stripped on create.
+    private static void prepareAdminManagedFields(Application application, Application existing, AdminManagedFieldsWrite write) {
```
astsiapanay: The argument name `write` confuses.

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
```diff
@@ -770,14 +772,23 @@ private Future<?> putResource(ResourceDescriptor descriptor) {
...
+                JsonNode bodyJson = ProxyUtil.parseTree(pair.getValue());
                 ExternalServicesWriteMode externalServicesWriteMode =
-                        ProxyUtil.hasTopLevelField(pair.getValue(), "external_services", "externalServices")
+                        ProxyUtil.hasTopLevelField(bodyJson, "external_services", "externalServices")
                                 ? ExternalServicesWriteMode.OVERRIDE
                                 : ExternalServicesWriteMode.PRESERVE_IF_OMITTED;
+                // Admin writes to public/ are authoritative for the admin-managed governance fields, but only
+                // for fields present in the body — an omitted field inherits, so a read-modify-write client
+                // can't wipe a grant, while present-as-null explicitly clears it.
+                AdminManagedFieldsWrite adminManagedFieldsWrite = adminPublicWrite
```
astsiapanay: The logic should be moved to AdminManagedFieldsWrite

server/src/main/java/com/epam/aidial/core/server/service/AdminManagedFieldsWrite.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/service/ApplicationService.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/ResourceController.java
DmytroZaichenkoDev: Fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1742, opened 2026-07-20
fix: release pooled ByteBuf in SseParser #1741

### LINKED ISSUE #1741: Netty pooled ByteBuf leak in SseParser causes unbounded PoolChunk growth
## Summary

A heap dump showed 673 instances of `io.netty.buffer.PoolChunk` (4MB each, ~2.7GB total) with counts growing over time. Root cause traced to a pooled Netty `ByteBuf` that is allocated per SSE stream and never released.

## Root cause

`SseParser` allocates a raw pooled `ByteBuf` directly from the default allocator:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/sse/SseParser.java#L35

```java
chunkBuffer = ByteBufAllocator.DEFAULT.heapBuffer(initialChunkBufferSize, Integer.MAX_VALUE);
```

It is written to and `.clear()`'d during parsing, but **`.release()` is never called** anywhere in the class, and a module-wide grep for `.release(` across `server/src/main/java` returns zero matches.

A new `SseParser` is created per streaming SSE request/response via:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java#L58

`BufferingReadStream` itself is constructed fresh for essentially every proxied SSE-capable response (chat completions, Responses API, Anthropic messages, MCP proxy, interceptors, route handler — the hottest paths in the proxy).

## Why this causes unbounded PoolChunk growth

For a pooled `ByteBuf`, backing memory is only returned to its `PoolArena` via an explicit `.release()`. Letting the object become unreachable and GC'd does **not** free the region — the arena's internal bitmap still marks that subpage as "in use" indefinitely. Under sustained streaming traffic, each abandoned per-request buffer permanently pins a sliver of some chunk, so chunks can never be fully reclaimed/reused and the arena is forced to keep allocating new 4MB chunks to satisfy new requests. This matches the observed symptom of chunk count growing with request volume rather than settling into a fixed steady-state.

Note: the 4MB chunk size itself is just Netty's stock default (`pageSize=8KB × 2^maxOrder(9)`) — no custom allocator tuning exists in this repo, so chunk size is not itself a signal of misconfiguration; the leak is specifically the missing `release()`.

## Suggested fix

1. Add a `close()` method to `SseParser` that calls `chunkBuffer.release()`.
2. In `BufferingReadStream`, call `eventStreamParser.close()` on every path that ends the stream:
   - In `handleEnd`, after `eventStreamParser.finish()`.
   - In `handleException`, which currently doesn't touch `eventStreamParser` at all despite the buffer having been allocated.
   - Guard so `close()` fires exactly once even if both paths could theoretically be reached.

## Verification suggestions

- Unit test asserting `chunkBuffer.refCnt() == 0` after `close()`, on both the normal-completion and exception paths.
- `./gradlew :server:test --tests "*SseParser*"` / `--tests "*BufferingReadStream*"`.
- End-to-end: send repeated streaming chat-completion requests locally and confirm arena/PoolChunk usage no longer grows unboundedly after the fix (e.g. via `-Dio.netty.leakDetection.level=paranoid` or heap dump diffing).

astsiapanay:
Fixes a Netty pooled `ByteBuf` leak in `SseParser` that caused unbounded `PoolChunk` growth under sustained SSE streaming traffic (chat completions, Responses API, Anthropic messages, MCP proxy, etc.).

### Applicable issues

- fixes #1741

### Description of changes

- `SseParser`: added a `close()` method that releases the internal pooled `chunkBuffer` (previously never released after allocation via `ByteBufAllocator.DEFAULT.heapBuffer(...)`).
- `BufferingReadStream`: calls `eventStreamParser.close()` on every path that ends the stream — after `finish()` in `handleEnd`, and in `handleException` — guarded by a flag so it fires exactly once regardless of how the stream terminates.
- `SseParserTest`: added a test verifying the buffer's ref count drops to 0 after `close()`, and that a second `close()` correctly throws `IllegalReferenceCountException`.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/sse/SseParser.java
+++ b/server/src/main/java/com/epam/aidial/core/server/sse/SseParser.java
@@ -1,5 +1,6 @@
 package com.epam.aidial.core.server.sse;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.vertx.core.buffer.Buffer;
@@ -70,6 +71,18 @@ public class SseParser {
         } catch (Throwable error) {
             log.error("Error occurred at finishing SSE stream", error);
         }
+    }
+
+    /**
+     * Releases the pooled buffer backing this parser. Must be called exactly once when the parser is no longer used.
+     */
+    public void close() {
+        chunkBuffer.release();
+    }
+
+    @VisibleForTesting
+    int chunkBufferRefCnt() {
+        return chunkBuffer.refCnt();
     }
 
     private void processLine() {
--- a/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java
+++ b/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java
@@ -36,6 +36,7 @@ public class BufferingReadStream implements ReadStream<Buffer> {
     private Throwable error;
     private boolean ended;
     private boolean reset;
+    private boolean eventStreamParserClosed;
     private final SseParser eventStreamParser;
     private final BaseEventListener eventListener;
     // promise on input stream is completed
@@ -167,6 +168,7 @@ public class BufferingReadStream implements ReadStream<Buffer> {
             handleEndInternal(ignored);
         } else {
             eventStreamParser.finish();
+            closeEventStreamParser();
         }
     }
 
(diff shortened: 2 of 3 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

PR #1751, opened 2026-07-23
feat: support interceptors for responses api

### LINKED ISSUE #1748: Support interceptors for Responses API
### Name and Version

0.45.6

DIAL Core currently supports interceptors for Chat Completions API, but does not call interceptors for Responses API requests. This prevents Responses API calls from being intercepted.

Update DIAL Core to call the appropriate Responses API endpoint on configured interceptors when processing Responses API requests.

_No response_

Oleksii-Klimov:


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/function/CollectRequestStandardAttachmentsFn.java
```diff
@@ -6,9 +6,9 @@
 
 import java.util.Set;
 
-public class CollectRequestChatCompletionAttachmentsFn extends CollectRequestAttachmentsFn {
```
astsiapanay, line 9: Why did you rename the class?

It's clear from the class's name what the class is intended for

server/src/main/java/com/epam/aidial/core/server/controller/BaseInterceptorController.java
```diff
@@ -37,15 +43,108 @@ public class InterceptorController extends BaseDeploymentPostController {
     private static final Pattern DEPLOYMENT_PATH = Pattern.compile("(/openai/deployments/)([^/]+)(/.*)");
 
     private final List<BaseRequestFunction<RequestObject>> enhancementFunctions;
-
-    public InterceptorController(Proxy proxy, ProxyContext context) {
+    private final int interceptorIndex;
```
astsiapanay, line 33: I would keep the base class of interceptor controller and extend it for specific use cases: chat completions, responses and etc.

There are generic methods and specialized methods need to be overridden for concrete use case.

Oleksii-Klimov: Done.

server/src/main/java/com/epam/aidial/core/server/function/CollectRequestStandardAttachmentsFn.java
Oleksii-Klimov: It works for both APIs: Chat Completions and Responses. Maybe eventually, it will work for other APIs too. Having “ChatCompletion” in the name could cause confusion.

astsiapanay: Please add javadoc. It's worth having explanation

server/src/main/java/com/epam/aidial/core/server/controller/BaseInterceptorController.java
```diff
@@ -26,26 +23,49 @@
...
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.IOException;
 import java.util.List;
-import java.util.function.Supplier;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 @Slf4j
-public class InterceptorController extends BaseDeploymentPostController {
```
astsiapanay, line 35: I would rename to BaseInterceptorController since the class is abstract to make class naming consistent.

astsiapanay (dismissed):

astsiapanay (commented):
resolve please comments

server/src/main/java/com/epam/aidial/core/server/controller/BaseInterceptorController.java
Oleksii-Klimov: Renamed.

server/src/main/java/com/epam/aidial/core/server/function/CollectRequestStandardAttachmentsFn.java
Oleksii-Klimov: Added.

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1752, opened 2026-07-24
feat: support deployment_name alias for interfaces base_url routing

khshanovskyi:
### Applicable issues

- 

### Description of changes

Before the `interfaces`/`base_url` routing model, admins could publish a lightweight "alias" deployment (its own displayName/description/iconUrl) whose `endpoint` pointed at a different, "real" deployment's fully-qualified route. Since `endpoint` is forwarded verbatim, this worked as a two-hop redirect.

The `interfaces` flow forwards to `base_url` + the exact ingress path instead, so an alias declared this way forwarded the request under its own deployment id again. When `base_url` pointed back at DIAL Core itself, this caused an infinite loop (Core kept re-resolving the alias and re-forwarding to itself).

Adds an optional `deployment_name` field to `interfaces.<type>` that routes the request as if it had been sent to a different deployment name:
- `openaiChatCompletions` carries the deployment id in the ingress path (`/openai/deployments/{name}/...`), so that path segment is rewritten to `deployment_name` before being appended to `base_url`.
- `openaiResponses` and `anthropicMessages` carry the deployment id in the request body's `model` field instead, so `model` is rewritten to `deployment_name` there before forwarding.

Supported for models and applications (chat completions only for applications, matching their existing interface support). Self-referencing/cyclic `deployment_name` values are not currently detected — documented as a known limitation, not addressed by this PR.

### Checklist

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/Deployment.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Deployment.java
@@ -115,6 +115,15 @@ public abstract class Deployment extends RoleBasedEntity {
     }
 
     /**
+     * Deployment id this interface should route to instead of this deployment's own id, or null when not
+     * declared. Only meaningful alongside {@link #getInterfaceBaseUrl}.
+     */
+    public String getInterfaceDeploymentName(InterfaceType type) {
+        DeploymentInterface deploymentInterface = interfaces == null ? null : interfaces.get(type.getValue());
+        return deploymentInterface == null ? null : deploymentInterface.getDeploymentName();
+    }
+
+    /**
      * Legacy fully-qualified endpoint configured for the type, or null.
      */
     public String getLegacyEndpoint(InterfaceType type) {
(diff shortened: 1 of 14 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/Deployment.java
```diff
@@ -114,6 +114,15 @@ public String getInterfaceBaseUrl(InterfaceType type) {
         return url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
     }
 
+    /**
+     * Deployment id this interface should route to instead of this deployment's own id, or null when not
+     * declared. Only meaningful alongside {@link #getInterfaceBaseUrl}.
+     */
+    public String getInterfaceDeploymentName(InterfaceType type) {
```
astsiapanay, line 123: Annotate with `@Nullable`

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1755, opened 2026-07-27
fix: reuse shared HttpClient for MCP transports #1754

### LINKED ISSUE #1754: Unbounded java.net.http.HttpClient creation in MCP toolset flows leaks SelectorManager/Worker threads under load
## Summary

Under performance-testing load, heap dumps show an excessive number of `HttpClient-N-SelectorManager` and `HttpClient-N-Worker-*` daemon threads. Root cause: two code paths build a brand-new `java.net.http.HttpClient` (via the MCP Java SDK's `HttpClientStreamableHttpTransport`) on **every incoming HTTP request**, instead of reusing a shared/pooled client.

## Where

- [`ToolSetToolsController.fetchTools()`](server/src/main/java/com/epam/aidial/core/server/controller/ToolSetToolsController.java#L184-L195) — builds a new transport/client on every `GET /v1/toolset/{toolset_name}/tools` and `GET /v1/toolset/{toolset_name}/allowed-tools` request.
- [`SecuredResourceService.validateApiKey()`](server/src/main/java/com/epam/aidial/core/server/service/SecuredResourceService.java#L89-L95) — same pattern on toolset sign-in (lower frequency, same root cause).

```java
HttpClientStreamableHttpTransport transport = HttpClientStreamableHttpTransport
        .builder(upstream.getEndpoint())
        .jsonMapper(McpClientUtils.MCP_JSON_MAPPER)
        .httpRequestCustomizer((builder, method, endpoint, body, transportContext) ->
                customizeRequest(builder, deployment))
        .build();
```

## Why this leaks threads

Decompiling `io.modelcontextprotocol.sdk:mcp-core:1.1.2` (`HttpClientStreamableHttpTransport$Builder.build()`) confirms it **always** calls `clientBuilder.connectTimeout(...).build()` internally, constructing a fresh `java.net.http.HttpClient` on every call — there is no builder API to inject an already-built, shared `HttpClient` (only a `HttpClient.Builder`, which gets rebuilt from scratch regardless of what's passed in).

Each `java.net.http.HttpClient` instance eagerly spins up its own dedicated `SelectorManager` thread plus an internal cached-executor worker pool (daemon threads in modern JDKs). Nothing in the transport's `closeGracefully()` (nor `McpSyncClient.close()`, called via try-with-resources at both call sites) touches the underlying `HttpClient` — only the MCP session/transport layer is torn down. These threads are reclaimed only when the `HttpClient` object becomes phantom-reachable and is garbage-collected, not when the request completes.

Under sustained load, requests to the toolset-tools endpoints arrive faster than GC reclaims the discarded clients, so live SelectorManager + Worker thread counts grow with concurrent/recent request volume instead of staying bounded — matching the heap-dump symptom.

For contrast, the other two `java.net.http.HttpClient` users in the codebase (`ApplicationSchemaService`, `ResourceAuthorizationClient`) build their client once as singletons in `AiDial.start()` and are not implicated.

## Suggested fix

Cache/reuse the `McpSyncClient` (and its underlying transport/`HttpClient`) per toolset + upstream endpoint instead of building one per incoming HTTP request, e.g. a bounded, idle-evicting cache. Per-request auth is already applied via the `httpRequestCustomizer` callback, which is invoked on every outgoing HTTP call rather than baked in at build time, so it can be adapted to look up the current request context dynamically — making safe client reuse across requests feasible without losing per-request auth headers.

## Impact

Thread exhaustion / excessive memory and scheduling overhead under sustained concurrent load on toolset tool-listing endpoints, visible as a large, growing population of `HttpClient-*-SelectorManager` and `HttpClient-*-Worker-*` daemon threads in heap dumps.

astsiapanay:
Under sustained load, `ToolSetToolsController.fetchTools()` and `SecuredResourceService.validateApiKey()` each built a brand-new `java.net.http.HttpClient` (via the MCP SDK's `HttpClientStreamableHttpTransport`) on every request, leaking `SelectorManager`/worker threads that are only reclaimed on GC. This introduces a shared, process-lifetime `HttpClient` that both call sites reuse instead.

### Applicable issues

- fixes #1754

### Description of changes

- Added `McpHttpClientBuilderService`, which owns a single shared `java.net.http.HttpClient` and exposes an `HttpClient.Builder` whose `build()` always returns that same instance (other builder calls are no-ops), so it can be plugged into the MCP SDK's transport builder without the SDK constructing a fresh client per call.
- Wired the new service into `AiDial`/`Proxy`, and updated `ToolSetToolsController.fetchTools()` and `SecuredResourceService.validateApiKey()` to build their `HttpClientStreamableHttpTransport` via `.clientBuilder(mcpHttpClientBuilderService.httpClientBuilder())`, reusing the shared client instead of allocating a new one per request.
- Renamed the static settings key `mcpClientCache` → `mcpHttpClient` (in both `aidial.settings.json` files and the lookup in `AiDial`) to match the new class name, and documented `mcpHttpClient.connectTimeout` in the README settings table.

### Checklist


DmytroZaichenkoDev (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [30260561577](https://github.com/epam/ai-dial-ci/actions/runs/30260561577)

| [deploy-review](https://chat-ai-dial-core-pr-1755.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30260561577/job/89959193025) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30260561577/job/89961879358) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1759, opened 2026-07-27
fix: stabilize flaky ResponsesControllerTest deep-stub race

DmytroZaichenkoDev:
`ResponsesControllerTest.testInterceptorReentryDispatchesToNextInterceptor` fails intermittently in CI with `WrongTypeOfReturnValue: TokenStatsTracker cannot be returned by getDeploymentService()`. It is a test-only race, unrelated to any production change — it most recently reddened a docs-only PR.

### Applicable issues

- N/A (CI flake)

### Description of changes

- The two interceptor-reentry tests complete the `VertxTestContext` from inside the `httpClient.request(...)` answer, so `await()` returns while the controller is still running its error path on a Vert.x thread.
- The test thread then evaluated `verify(proxy.getDeploymentService(), ...)`, resolving a `RETURNS_DEEP_STUBS` child mock on the `proxy` mock concurrently with the controller resolving `proxy.getTokenStatsTracker()`. Deep-stub resolution mutates the mock's invocation container, so the concurrent lookup returned the wrong child mock.
- Fix: resolve `proxy.getDeploymentService()` once during setup, before `controller.handle()`, and verify against that instance. The deep stub is then created while only the test thread is running, and the Vert.x thread only reads existing stubbings.
- Test-only change; no production code touched.

### Checklist

### DIFF
--- a/server/src/test/java/com/epam/aidial/core/server/controller/ResponsesControllerTest.java
+++ b/server/src/test/java/com/epam/aidial/core/server/controller/ResponsesControllerTest.java
@@ -17,6 +17,7 @@ import com.epam.aidial.core.server.data.ResponseMapping;
 import com.epam.aidial.core.server.limiter.RateLimitResult;
 import com.epam.aidial.core.server.security.ApiKeyStore;
 import com.epam.aidial.core.server.service.ConsentService;
+import com.epam.aidial.core.server.service.DeploymentService;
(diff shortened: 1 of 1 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

Issue #1618, opened 2026-06-12
Support Anthropic API


khshanovskyi:
### Name and Version

0.45

### What is the problem this feature will solve?

To make Claude Code to work through DIAL Core with Anthropic API supported models in `pass-through` mode.

### What is the feature you are proposing to solve the problem?

Support 2 endpoints:
- `{core}/anthropic/v1/messages`
- `{core}/anthropic/v1/messages/count_tokents`
It should work in `pass-through` mode with models that support Anthropic API (call `{core}/anthropic/v1/messages` <-> core calls `{adapter}/anthropic/v1/messages` <-> adapter calls model `{host}/anthropic/v1/messages`)

To work in `pass-through` mode we need to support on the Bedrock adapter Anthropic API (PR https://github.com/epam/ai-dial-adapter-bedrock/pull/433)

PR #1760, opened 2026-07-28
fix: resolve correct upstream endpoint for interfaces-based deployments called through an interceptor

khshanovskyi:
### Applicable issues

- feat #1618

### Description of changes

Deployments configured with the new `interfaces.openaiChatCompletions.base_url` (instead of the legacy `endpoint` field) returned 404 from the upstream whenever the deployment also had `interceptors` configured. After the interceptor chain finished, Core built the final upstream URL by appending the *inbound request's raw URI* to `base_url`. That inbound request is the interceptor's callback into Core, whose path carries the literal pseudo deployment id `interceptor` (`/openai/deployments/interceptor/chat/completions`) rather than the real deployment name, so Core dialed `base_url + /openai/deployments/interceptor/chat/completions` instead of the deployment's actual path. Legacy `endpoint`-configured deployments were unaffected because that flow ignores the inbound path entirely and uses the configured absolute URL as-is.

`Deployment.rewriteDeploymentPathSegment` only rewrote the `/deployments/{id}/` path segment when an explicit `deployment_name` alias was configured, and even then it matched against the deployment's own name — which never appears in the interceptor callback path. The fix makes the rewrite unconditional: it now always normalizes whatever id sits between `/deployments/` and the next `/` to the correct target deployment (alias if declared, otherwise the deployment's own name), using a regex match-and-replace instead of manual index arithmetic. This mirrors the equivalent rewrite already used for the Core-to-interceptor hop in `ChatCompletionInterceptorController`.

### Checklist

- [x] All tests pass (DeploymentTest 16/16 incl. new regression test, ChatCompletionInterceptorControllerTest, ResponsesInterceptorControllerTest, ResponsesControllerTest, DeploymentPostControllerTest, checkstyle clean)

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/Deployment.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Deployment.java
@@ -9,11 +9,16 @@ import lombok.EqualsAndHashCode;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 @Data
 @EqualsAndHashCode(callSuper = true)
 public abstract class Deployment extends RoleBasedEntity {
+
+    private static final Pattern DEPLOYMENT_SEGMENT = Pattern.compile("/deployments/[^/]+/");
+
     private String endpoint;
     private String responsesEndpoint;
     /**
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1764, opened 2026-07-28
fix: rollback deploymentName for interfaces

khshanovskyi:
### Applicable issues

-

### Description of changes

Rolls back #1752's `deployment_name` field on `interfaces.<type>`. Recursive/loopback calls from Core back to Core (an alias's `base_url` pointing at DIAL Core itself) are not something we should support. The use case it targeted — calling a deployment under a different id/name — is already covered by the existing `overrideName` param for chat completions, so the extra indirection isn't needed.

Keeps the `/deployments/{id}/` ingress path rewrite for the interceptor pseudo-id case (fixed separately in #1760), just always rewriting back to the deployment's own name instead of an aliased one.

### Checklist

- [x] All tests pass (DeploymentTest, DeploymentPostApiTest, MessagesApiTest, DeploymentPostControllerTest, ResponsesControllerTest all green; checkstyle clean; OpenAPI spec regenerated and verified in sync)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1773, opened 2026-07-30
feat: materialize applications/toolsets into the platform bucket

### LINKED ISSUE #1782: Slice A — Materialize config-managed applications/toolsets in the platform bucket
# Slice A — Materialize config-managed applications/toolsets in the `platform` bucket

Part of #1781.

## Goal

Make config-managed applications and toolsets first-class, API-managed config
entities stored in the **`platform`** bucket and materialized into the in-memory
`Config` (like models/roles/interceptors). This slice keeps them **canonically
addressed** — short-name addressing is added in the Slice B+C issue. User-published
apps/toolsets in the `public` bucket are untouched; the two buckets run
simultaneously.

Why `platform` and not `public`: (a) `public` cannot disambiguate a migrated config
app from a user app in listings; (b) migrating into `public` would overwrite a
same-named user-published app (`applications/public/{name}` collision → data loss);
(c) config apps declare access via `userRoles` (`Deployment.hasAccess`, same path as
models), whereas `public` apps use folder rules (`AccessService`→`RuleService`) —
`platform` preserves the `userRoles` semantics.

## Scope

- **`MergedConfigStore.MANAGED_TYPES`**: add `APPLICATION` and `TOOL_SET` (resolved to
  the `platform` bucket). Extend `deserializeReplicaEntity`, `cloneTypeMap`,
  `peekEntity`, `putEntityInPlace`, `removeEntityInPlace`, `removeAddedEntity` with
  the two cases.
- **`EntityBucketBinding`**: allow `PLATFORM_BUCKET` for `applications`/`toolsets`
  (alongside the existing `PUBLIC_BUCKET`).
- **`PlatformEntityLocationStrategy`**: add `APPLICATION`/`TOOL_SET` to `resolveBucket`
  → `PLATFORM_BUCKET` (currently throws for them).
- **Routing (`ControllerSelector` / `RouteTemplate`)**: add dedicated routes matching
  only the literal `platform` bucket —
  `^/v1/(applications|toolsets)/platform/(?<path>.*)$` and the `/v1/metadata/...`
  twin — wired to the config-resource controller for all four verbs (the metadata
  twin wires all four so it can return `405 Allow: GET` rather than a router 404).
  Register **before** the generic resource routes. The generic resource routes and
  the existing config-resource routes are otherwise unchanged.
- **`ConfigResourceController`**: its four entity-type switches (GET dispatch,
  descriptor build, `WriteSpec`, entity-class map) lack `APPLICATION`/`TOOL_SET`
  cases. Add them, but the write/read cases **must delegate to `ApplicationService` /
  `ToolSetService`** (with a `platform` descriptor), **not** the generic raw-blob +
  `secretFieldProcessor` `WriteSpec` path. See "Secret handling" below.
- **`ConfigResourceMetadataController`**: generic (gates via
  `EntityBucketBinding.isAllowed` only) — no per-type change once the bucket allowlist
  permits the two types.
- **`AdminApplyController`**: `applyApplication`/`applyToolSet`/`entityId` currently
  hardcode `PUBLIC_BUCKET`. Route apps/toolsets to `platform` (keep going through
  `ApplicationService`/`ToolSetService`; only the bucket changes). This changes the
  reported `entityId` shape from `applications/public/{name}` to
  `applications/platform/{name}`. The user-published-app write path on
  `ResourceController` is unchanged.

## Secret handling (do not use the generic config-write path)

Applications and toolsets are **not** plain config types:

- `Application` has no `@EncryptedField`, so `secretFieldProcessor` is a no-op for it;
  external-service secrets go through `ExternalServiceService` (`processOnWrite` on
  write, `decryptSecrets` on read). `ApplicationService.putApplication` also performs
  admin-managed-field governance (use `AdminManagedFieldsWriteMode.AUTHORITATIVE`),
  function preparation, and schema validation.
- `ToolSet extends SecuredResource` and carries `authSettings` (incl. `clientSecret`),
  encrypted by `ResourceAuthSettingsEncryptionService` (keyed by `resource.getUrl()`)
  inside `ToolSetService.putToolSet`.

**Materialization decrypt gap:** `MergedConfigStore.rebuild` currently only calls
`secretFieldProcessor.decryptFields`, which covers neither of the above. A `platform`
app/toolset with secrets would otherwise materialize into `Config` with **encrypted**
credentials. The app/toolset materialization path must additionally run the
type-specific decrypt against the `platform` descriptor (equivalently, read them via
`ApplicationService.getApplicationWithDecryptedSecrets` / `ToolSetService`'s
decrypting read). Both schemes are **keyed by the resource URL**, so write and decrypt
must use the same `platform` URL for a given entity.

## Notes

- `isPublicOrReview` is `false` for `platform` (correct — the branches it gates are
  public-publication cleanup that endpoint-based config apps don't have). Consider a
  guard rejecting `function`-type apps on the `platform` write path (a function app
  there would leak its target folder on delete; config apps are endpoint-based).
- `FileConfigController` already maps `APPLICATION`/`TOOL_SET`; no change.

## Testing

- **Unit — `EntityBucketBindingTest`**: `isAllowed("applications"/"toolsets",
  "platform")` → true; `"public"` still true; others false.
- **Unit — `ControllerSelectorTest`**: `/v1/(applications|toolsets)/platform/...` (and
  the `/metadata/` twin) route to the config-resource controller; `public`/user-bucket
  paths still route to the generic resource controller; `GET /v1/models/public/gpt-4`
  still returns the bodyless 404.
- **Unit — new small `PlatformEntityLocationStrategy` test** (or fold into
  `MergedConfigStoreTest`): resolves `APPLICATION`/`TOOL_SET` to `platform`.
- **Integration — `AdminApplyApiTest`**: applying an app/toolset now targets
  `platform` (assert the `entityId` shape changed to `.../platform/...`).
- **Integration — new `PlatformAppToolsetApiTest`**: `GET/PUT/DELETE
  /v1/applications/platform/{name}` reach config-resource semantics;
  `/v1/applications/public/{name}` + user-bucket paths still reach the generic
  resource controller; `platform` apps/toolsets listed unconditionally while `public`
  stays gated by `includeCustomApps`; a same-named `platform` + `public` pair resolves
  `platform`.
- **Integration** — secret round-trip: a `platform` application with an external
  service and a `platform` toolset with `authSettings.clientSecret` are stored
  encrypted and surface decrypted in the runtime `Config`.

## Out of scope

Short-name inbound/outbound addressing for these entities — that is the Slice B+C
issue. After this slice, `platform` apps/toolsets are addressed by canonical id only.

KirylKurnosenka:
### Applicable issues

- fixes #1782 

Adds admin-only CRUD (GET/PUT/DELETE) for applications and toolsets under the platform bucket, alongside models/interceptors/roles/keys/routes:

- MergedConfigStore materializes platform apps/toolsets into the merged Config via the existing MANAGED_TYPES blob-scan/replica-fast-path machinery, decrypting external-service/auth_settings secrets through ExternalServiceService/ResourceAuthSettingsEncryptionService.
- New platform-only routes (ConfigResourceController/ConfigResourceMetadataController) delegate writes to ApplicationService/ToolSetService rather than the generic raw-blob path, since those services own atomic writes, function lifecycle bookkeeping, admin-managed-field governance, and credential purge on delete.
- ApplicationService preserves userRoles for platform-bucket apps (their access model is userRoles itself, unlike public-bucket user-published apps) and rejects function-type apps on platform.
- AdminApplyController (`POST /v1/admin/apply`) manifest `name` is now the canonical resource id (`<kind-segment>/<bucket>/<name>`, e.g. `models/platform/gpt-4` or `applications/public/my-app`). For every kind except Application/ToolSet the bucket must be `platform`; Application/ToolSet manifests may target either `platform` or `public`, so the client picks the bucket explicitly instead of it always defaulting to platform — this keeps OBO/user-authored-external-service workflows (which require the public bucket) working. Manifests targeting the `public` bucket are written through the normal per-resource path only and are **not** materialized into MergedConfigStore (that stays platform-bucket-only, matching MergedConfigStore's rebuild-time scan).
- AccessService.getAdminAccess now grants admin ALL access over platform-bucket **Application/ToolSet** resources specifically, closing a gap where external-service sign-in/credentials controllers (which check access via AccessService.lookupPermissions, not AdminRoleAuthorizationService) would 403 for platform-bucket applications. Other platform-bucket resource types (models, interceptors, roles, routes, keys) are unaffected — they're already fully admin-gated via AdminRoleAuthorizationService.
- Fixed a latent ConfigPostProcessor bug where canonical-id-keyed toolset names (containing '/') were incorrectly rejected as invalid; the relaxed key-validation is scoped to ToolSet names only, so external-service ids on Applications keep the strict no-slash check.
- OpenAPI docs for the new platform routes, and generator exclusion-list updates for their 405-only verbs.

### Applicable issues

- fixes #

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -80,7 +85,9 @@ public final class MergedConfigStore implements ConfigStore {
             ResourceTypes.INTERCEPTOR,
             ResourceTypes.ROLE,
             ResourceTypes.PROJECT_KEY,
-            ResourceTypes.ROUTE);
+            ResourceTypes.ROUTE,
+            ResourceTypes.APPLICATION,
```
astsiapanay, line 89: `MANAGED_TYPES` feeds two code paths, and only one of them is bucket-scoped:

- `rebuild()` (L1018-1024) iterates `locationStrategy.resolveBucket(type, scope)`, which always returns `platform`.
- `onResourceEvent()` (L259-287) filters on the first URL segment (`isManagedEventUrl`, derived from `MANAGED_TYPES`) and then on `MANAGED_TYPES.contains(descriptor.getType())` — no bucket check anywhere.

`ResourceDescriptorFactory.fromAnyUrl(url, null)` resolves `public` and `platform` and throws for encrypted user buckets, so user-bucket app writes are harmlessly dropped but **public-bucket ones are applied**.

Scenario (needs ≥2 pods): an admin approves a publication on pod A → `applications/public/my-app` is written and `ResourceService.publishEvent` broadcasts it. Pod B is not the sender, so it runs `applyReplicaEvent` → `applyEntityWrite(APPLICATION, "applications/public/my-app", app)`.

Public-bucket apps have `userRoles == null` — `prepareApplication` strips it, correctly, because their visibility is governed by `RuleService` publication rules. But `RoleBasedEntity.hasAccess` returns `true` when `userRoles` is null, so on pod B that app is now visible to **every** authenticated caller through `DeploymentController.getApplications()`, `ApplicationController#getApplications` and `Config.selectDeployment` — bypassing the publication rules — and it is returned twice, since those listings also merge `deploymentService.listDeployments`. Same mechanism for public toolsets via `ToolSetController.mergeToolsets`.

Extra effect on toolsets: a leaked `toolsets/public/x` key makes `Config.isDeploymentExists("toolsets/public/x")` true, which flips `CredentialsLocatorFactory.fromAnyUrl` onto the static-config branch and points OAuth credential lookups at a different path than the one sign-in wrote on pod A.

The entries vanish again on pod B's next full rebuild (platform-only scan), so visibility flaps rather than staying broken.

Suggested fix — scope the event filter the way `rebuild()` is scoped:

```java
String managedBucket = locationStrategy.resolveBucket(type, EntityLocationStrategy.PLATFORM_SCOPE);
if (!managedBucket.equals(descriptor.getBucketName())) {
    return;
}
```

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -396,11 +413,32 @@ private static Object deserializeReplicaEntity(ResourceTypes type, JsonNode node
...
+    /**
+     * Decrypts a managed entity read from blob storage before it enters the merged {@link Config}.
+     * {@code APPLICATION}/{@code TOOL_SET} secrets are not {@code @EncryptedField}-annotated — they
+     * are encrypted per-resource by their own write paths ({@link ExternalServiceService}/
+     * {@link ResourceAuthSettingsEncryptionService}), keyed by the entity's own bucket — so they
+     * must be decrypted the same way here rather than via {@link SecretFieldProcessor}, which is a
+     * no-op for these two types (no {@code @EncryptedField} fields). Every other managed type keeps
+     * going through {@link SecretFieldProcessor}.
+     */
+    private void decryptManagedEntity(ResourceTypes type, Object entity, ResourceDescriptor descriptor) {
```
astsiapanay, line 443: Holding APPLICATION/TOOL_SET decrypted in the merged `Config` is deliberate, and `ConfigResourceController` does redact on per-entity GET via the new `redactExternalServiceSecrets`/`redactAuthSettingsSecrets`. But the merged `Config` is also serialized wholesale by `ConfigController#handle`:

```java
.onSuccess(config -> context.respond(HttpStatus.OK, config))   // POST /v1/ops/config/reload
```

So `POST /v1/ops/config/reload` returns the plaintext `external_services.*.auth_settings.client_secret` and toolset `auth_settings.client_secret`/`code_verifier` in its response body — the exact values the GET path strips.

Before this change the merged Config's applications/toolsets came only from the config file, where those values are already plaintext, so the endpoint revealed nothing new. This is the first time secrets that are encrypted at rest become readable through it, which cuts against the "no plaintext-reveal path" note in `projectItem`.

Admin-only, so not privilege escalation — but worth closing, either by redacting in the reload response or by keeping the reveal decision in one place.

server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
```diff
@@ -185,6 +185,7 @@ Map<ResourceDescriptor, Set<ResourceAccessType>> getAdminAccess(
         if (hasAdminAccess(context)) {
             return resources.stream()
                     .filter(resource -> resource.isPublic()
+                            || isPlatformBucketAppOrToolSet(resource)
```
astsiapanay: This grant makes admin work, but end users stay locked out, which undercuts preserving `userRoles` for platform apps.

A platform app's merged-Config key is `applications/platform/my-app`. The external-service controllers derive the app part from the URL, so `/v1/applications/platform/my-app/external-services/svc1/...` yields `appPart = "platform/my-app"`, and `context.getConfig().selectDeployment("platform/my-app")` misses — the key carries the `applications/` prefix. `ExternalServiceCredentialsController#resolveApplication` therefore takes the dynamic branch (`staticApp = false`) and `verifyAccess` falls through to `accessService.lookupPermissions` on a platform-bucket descriptor, which is exactly why this grant was needed for admins.

Consequence: a user holding a role listed in the app's preserved `user_roles` gets `PermissionDeniedException` → 403, where the same app in the static config took the `resolved.application.hasAccess(context.getUserRoles())` branch and worked. Migrating a config app that has external services to the platform bucket therefore breaks user-level sign-in for everyone except admins.

If platform apps are meant to be config-equivalent, the fix probably belongs in `resolveApplication` (treat a platform-bucket app as a static app) or in `verifyAccess` (add a `hasAccess(userRoles)` arm for it), rather than in `AccessService`.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
```diff
@@ -1073,6 +1266,83 @@ private Future<?> handleSettingsDelete() {
...
+     * PUT-upsert for {@code APPLICATION}/{@code TOOL_SET} in the {@code platform} bucket. Delegates
+     * to {@link ApplicationService#putApplication} / {@link ToolSetService#putToolSet} — they own
+     * the atomic compute-based write, function lifecycle bookkeeping, admin-managed-field
+     * inheritance, and secret encryption (unchanged; both services already derive the encryption
+     * {@code BucketInfo} from the resource descriptor's own bucket, so {@code platform} needs no
+     * special-casing there). After the write, the entity is re-fetched with secrets decrypted so
+     * {@link MergedConfigStore#applyEntityWrite} receives a fully-decrypted entity, matching what
+     * the rebuild-time materialization loop produces.
+     */
+    private Future<?> handleAppOrToolSetPut() {
```
astsiapanay, line 1292: The name gate on this write surface is laxer than the toolset key validation the merged Config applies:

- `ENTITY_NAME_PATTERN` (L86) — `^[A-Za-z0-9._%:-]+$`
- `ConfigPostProcessor.isValidToolSetKey` → `RESOURCE_KEY_PATTERN` — `^[A-Za-z0-9-_]+$`

So `PUT /v1/toolsets/platform/my.toolset` returns 200, the blob is written, `applyEntityWrite` inserts it, and GET works. At the next full rebuild — a periodic file-config reload firing `requestRebuild()`, a pod restart, or the `requestRebuild()` fallback after a failed replica event — `processToolSets` drops the entry with only a warn log, because the trailing segment `my.toolset` fails `RESOURCE_KEY_PATTERN`.

After that the toolset is 404 on GET and absent from `/v1/toolsets` and `/v1/deployments`, while the blob still exists and re-PUTting keeps returning 200. Same for `:` and `%`.

Applications aren't affected (`processApplications` doesn't validate keys), so tightening the write-side check for TOOL_SET — or reusing `RESOURCE_KEY_PATTERN` on this path — would keep the two surfaces consistent.

astsiapanay (commented):
Reviewed the platform-bucket materialization end to end. Four issues below — all correctness/security, nothing stylistic. The design itself (delegating writes to `ApplicationService`/`ToolSetService`, canonical-id keying, redacting on GET) reads well; these are gaps at the edges.

The first one is the one I'd hold the merge for: `MANAGED_TYPES` drives both `rebuild()` and the cross-replica event path, but only `rebuild()` is bucket-scoped, so public-bucket applications/toolsets get materialized into the merged `Config` on every pod except the writer. Worth noting that no single-pod integration test can observe it — `onResourceEvent` short-circuits on `senderPodId.equals(thisPodId)`.

Two smaller notes not filed as findings:

- The comment at `MergedConfigStore` L1102-1106 says `onSkip` is never invoked for APPLICATION/TOOL_SET, but `processApplications`/`processToolSets` do call it via `skipOnDuplicate` on a duplicate deployment id.
- `redactSecretFields` hard-codes the JSON field names, duplicating `ResourceAuthSettings.withoutSecrets()`, whose javadoc calls itself "the single place read responses strip secrets" — a future secret field now has to be cleared in both.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Fixed in 7fff6eef. `onResourceEvent` now scopes the replica fast path to the managed (platform) bucket the same way `rebuild()` does — after the type filter it resolves `locationStrategy.resolveBucket(type, PLATFORM_SCOPE)` and drops the event unless it equals `descriptor.getBucketName()`:

```java
ResourceTypes managedType = (ResourceTypes) descriptor.getType();
String managedBucket = locationStrategy.resolveBucket(managedType, EntityLocationStrategy.PLATFORM_SCOPE);
if (managedBucket == null || !managedBucket.equals(descriptor.getBucketName())) {
    return;
}
```

So a `public`-bucket app/toolset publication event is now ignored on peer pods instead of being applied into the merged `Config`. It applies to every managed type (all platform-scoped), so it also hardens the pre-existing ones. Added `MergedConfigStoreReplicaUpdateTest` cases: public app/toolset events are dropped (no `taskExecutor`/`resourceService` interaction), platform app/toolset events are still dispatched.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
KirylKurnosenka: Fixed in 7fff6eef. The write path now rejects a `TOOL_SET` PUT **or DELETE** whose name segment fails `ConfigPostProcessor.isValidToolSetKey` — the exact validator `processToolSets` uses — so the write-side and rebuild-side checks are identical by construction rather than coincidentally equal:

```java
if (resourceType() == ResourceTypes.TOOL_SET
        && (method == HttpMethod.PUT || method == HttpMethod.DELETE)
        && !ConfigPostProcessor.isValidToolSetKey(path == null ? "" : path)) {
    context.respond(HttpStatus.BAD_REQUEST, ...);
}
```

Applied to both verbs to match the adjacent `ENTITY_NAME_PATTERN` gate. `PUT /v1/toolsets/platform/my.toolset` now returns 400 instead of creating a serve-until-next-rebuild orphan; applications are unaffected (not key-validated by `processToolSets`). Added `PlatformAppToolsetApiTest.testInvalidToolSetNameRejectedOnWrite`.

I used `400` to stay consistent with the adjacent `ENTITY_NAME_PATTERN` rejection rather than the `422` in your note — glad to switch to 422 if you'd prefer them aligned that way instead.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Worth closing, but I'd argue it's pre-existing `/reload` behavior rather than something this PR introduces for the migration case.

`ConfigController.handle` serializes the whole merged `Config`, and `Application` external-service `auth_settings` / `ToolSet` `auth_settings` fields are **not** `@JsonProperty(WRITE_ONLY)` (unlike `Upstream.key` / `Key.key`, which is exactly what keeps `/reload` safe for those). So config-file-defined apps/toolsets already expose these secrets — in plaintext — through `/reload` today. Migrating such an app from the file to the `platform` bucket therefore doesn't change its `/reload` visibility: plaintext-in-file before, decrypted-blob after.

The one genuinely new case is a **natively-created `platform` app/toolset** (not possible before this PR): its resource GET redacts via the new `redact*` helpers, but `/reload` still serializes the secret — a redact-on-GET vs. reveal-on-reload inconsistency with no file-era equivalent. Admin-only, as you note.

Since the underlying issue (`/reload` dumping non-`WRITE_ONLY` secret-bearing resource fields) predates this change and also affects file-defined apps, I'd prefer to address it as a separate hardening that redacts the `/reload` response for all cases, rather than scope it into this PR. Tracked as #1787. Happy to pull it in here instead if you'd rather not merge with the inconsistency open.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
KirylKurnosenka: Fixed in 6260e2df, in both external-service controllers.

Root cause: `appPart` (from `/v1/applications/platform/my-app/…`) carries no `applications/` type prefix, whereas the materialized Config key is `applications/platform/my-app`, so `selectDeployment(appPart)` missed and resolution fell through to the dynamic/folder-rules branch.

`resolveApplication` (credentials) and `resolveApp` (management) now resolve the app from the **in-memory Config** — verbatim first, then by canonical id (`applications/` + appPart) — so a platform-bucket app hits its materialized entry and resolves as a config-equivalent **static** app; `verifyAccess` / `canManageInline` then use `hasAccess(userRoles)`. The materialized entry already carries decrypted secrets (the write path re-fetches decrypted and calls `applyEntityWrite`), so there's no blob read on the request path.

That makes the `AccessService.isPlatformBucketAppOrToolSet` grant unnecessary — it only unblocked admins via `lookupPermissions` — so I removed it. Verified nothing else depends on it (`AccessServiceTest` and the config-admin read/write suites still pass).

Forward-compatible with short-name addressing: once that lands, `selectDeployment(shortName)` resolves the platform app directly via the verbatim branch.

Added `PlatformAppToolsetApiTest` cases: a non-admin user in the app's `user_roles` can reach a platform app's external service, and a user without the role gets 403.

ai-dial-actions:
<h1>Dependency Review</h1>
None

KirylKurnosenka:
Re the two smaller notes in the review body:

- **`onSkip` comment (`MergedConfigStore` L1102-1106)** — fixed in 156b5123. You're right: `processApplications`/`processToolSets` route a duplicate deployment-id skip through `onSkip` via `skipOnDuplicate`, and since these types are now in `MANAGED_TYPES` it's recorded via `recordSkippedByMapKey`. Code behavior was already correct; corrected the comment.
- **`redactSecretFields` duplicating `ResourceAuthSettings.withoutSecrets()`** — valid; the "single place" javadoc no longer holds. Deferring to a follow-up so we can pick the right approach (delegating JSON redaction through `withoutSecrets()` so it stays the literal single source, vs. a shared field-name source) rather than trading one hard-coded copy for another.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1771, opened 2026-07-30
fix: honor configured MCP endpoint path instead of defaulting to /mcp #1768

astsiapanay:
## Summary

Part of #1768. `HttpClientStreamableHttpTransport.Builder` defaults `.endpoint` to `"/mcp"`, and since it's resolved as an absolute-path reference, `URI#resolve` **replaces** (not appends to) the base URI's path. `ToolSetToolsController.fetchTools()` and `SecuredResourceService.validateApiKey()` both passed the whole configured endpoint as the transport's `baseUri` and never called `.endpoint(...)`, so any configured MCP endpoint whose path wasn't exactly `/mcp` was silently requested at `/mcp` instead.

- Added `McpClientUtils.transportBuilder(String endpoint)`, which splits the endpoint into origin + path/query and always sets `.endpoint(...)` explicitly, so the configured endpoint is what actually gets requested.
- Updated both call sites (`ToolSetToolsController`, `SecuredResourceService`) to use it.

## Not included in this PR

#1768 also describes a second, independent problem: `McpHttpClientBuilderService`'s `HttpClient.Builder` wrapper silently no-ops `followRedirects()` and `version()`, so 307/308 redirects (e.g. trailing-slash-normalizing servers like `markitdown-mcp`) still aren't followed, and the SDK's HTTP/1.1 preference is still discarded. That's a separate fix (touching the shared client builder's redirect policy) and is left for a follow-up — leaving #1768 open rather than closing it here.

## Test plan

- [x] Added `ToolSetApiTest.testGetAllTools_ExplicitPathEndpointIsHonored` — configures a toolset endpoint with an explicit non-`/mcp` path (`/call`) against a mock server with a strict fallback (404 for anything unmapped), proving the request lands on the configured path rather than `/mcp`.
- [x] `./gradlew :server:test --tests "com.epam.aidial.core.server.ToolSetApiTest"` and `--tests "com.epam.aidial.core.server.service.SecuredResourceServiceTest"`
- [x] `./gradlew :server:test` (full module): 1559 tests, 0 failures, 0 errors
- [x] `./gradlew checkstyleMain checkstyleTest`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1775, opened 2026-07-31
fix: follow 307/308 MCP redirects for SDK-driven toolset paths #1768

### LINKED ISSUE #1768: Toolset tool discovery ignores the configured MCP endpoint path and doesn't follow 307/308 redirects (#1667 fixed only the MCP proxy)
### Name and Version

aidial-core 0.47.0-dev.14 (`main` @ `7ce1780`)

### Description

#1667 / #1702 fixed the `/mcp` → `/mcp/` trailing-slash redirect problem for the **MCP proxy** path only — commit 92a8b45 touches nothing but `McpProxyController`. The **toolset tool-discovery** path has the same defect and is still broken, plus a second, independent bug that no amount of redirect-following would fix.

`ToolSetToolsController.fetchTools()` and `SecuredResourceService.validateApiKey()` pass the whole configured endpoint URL as the transport's **`baseUri`** and never call `.endpoint(...)`:

- [`ToolSetToolsController.fetchTools()`](server/src/main/java/com/epam/aidial/core/server/controller/ToolSetToolsController.java#L193-L199)
- [`SecuredResourceService.validateApiKey()`](server/src/main/java/com/epam/aidial/core/server/service/SecuredResourceService.java#L90-L96)

```java
HttpClientStreamableHttpTransport transport = HttpClientStreamableHttpTransport
        .builder(upstream.getEndpoint())   // full URL used as baseUri; .endpoint(...) never set
        .clientBuilder(mcpHttpClientBuilderService.httpClientBuilder())
        .jsonMapper(McpClientUtils.MCP_JSON_MAPPER)
        .httpRequestCustomizer((builder, method, endpoint, body, transportContext) ->
                customizeRequest(builder, deployment))
        .build();
```

In `io.modelcontextprotocol.sdk:mcp-core:1.1.2` the builder defaults to `private static final String DEFAULT_ENDPOINT = "/mcp"`, and every request resolves as `Utils.resolveUri(this.baseUri, this.endpoint)`:

```java
public static URI resolveUri(URI baseUrl, String endpointUrl) {
    if (!Utils.hasText(endpointUrl)) {
        return baseUrl;
    }
    URI endpointUri = URI.create(endpointUrl);
    if (endpointUri.isAbsolute() && !isUnderBaseUri(baseUrl, endpointUri)) {
        throw new IllegalArgumentException("Absolute endpoint URL does not match the base URL.");
    }
    else {
        return baseUrl.resolve(endpointUri);
    }
}
```

`"/mcp"` is root-absolute, so `baseUrl.resolve("/mcp")` **discards the configured path entirely**. Two consequences:

1. **The configured endpoint path is ignored.** A toolset configured at `https://host/api/mcp`, `https://host/v1/mcp` or `https://host/mcp/delivery_central` is unreachable from tool discovery — Core always requests `https://host/mcp`, no redirect involved. Only endpoints that happen to be exactly `/mcp` work.
2. **Trailing-slash frameworks fail.** For a configured `https://host/mcp/`, Core requests `https://host/mcp`, the server answers `307` → `/mcp/`, and the redirect is not followed, so the response falls through every handled branch (2xx, 400, 401, 403, 404) in `HttpClientStreamableHttpTransport.sendMessage()` into `new RuntimeException("Failed to send message: " + responseEvent)`.

Redirects are never followed because [`McpHttpClientBuilderService`](server/src/main/java/com/epam/aidial/core/server/service/McpHttpClientBuilderService.java#L28-L30) builds the shared client without `followRedirects`, so it keeps the JDK default `Redirect.NEVER`:

```java
this.httpClient = HttpClient.newBuilder()
        .connectTimeout(Duration.ofMillis(settings.getConnectTimeout()))
        .build();
```

Worse, its `HttpClient.Builder` wrapper (added in #1754 / #1755 to share one client) makes [`followRedirects()`](server/src/main/java/com/epam/aidial/core/server/service/McpHttpClientBuilderService.java#L88-L92) and `version()` **silent no-ops**, so any caller — including the SDK's own default builder, which sets `version(HTTP_1_1)` — is ignored:

```java
@Override
public HttpClient.Builder followRedirects(HttpClient.Redirect policy) {
    return this;
}
```

1. Deploy any MCP server whose framework performs trailing-slash redirects (Starlette/FastMCP and similar). A ready-made one: image `docker.io/mcp/markitdown:latest`, command `markitdown-mcp --http --host 0.0.0.0 --port 3001`. It serves streamable HTTP at `/mcp/` and answers `POST /mcp` with `307 Temporary Redirect` → `/mcp/`.
2. Register it as a toolset with the correct endpoint, including the trailing slash:
   ```json
   "markitdown-mcp": {
     "name": "markitdown-mcp",
     "endpoint": "https://<mcp-host>/mcp/",
     "forward_auth_token": false
   }
   ```
3. Call `GET /v1/toolset/markitdown-mcp/tools`.
4. Core responds `502 Bad Gateway` — `Failed to fetch tools from MCP server`.

Reproducing consequence (1) needs no redirect at all: point a toolset at an MCP server serving `/v1/mcp` and observe Core request `/mcp` and fail.

Tool discovery honours the configured endpoint path verbatim (including a trailing slash and any prefix such as `/api/mcp`), follows `307`/`308` when the upstream normalises the path, and returns the tool list.

### What do you see instead?

`502 Bad Gateway`, and in the logs:

```
Failed to fetch tools from MCP server for toolset: markitdown-mcp
java.lang.RuntimeException: Client failed to initialize by explicit API call
	at io.modelcontextprotocol.client.LifecycleInitializer.lambda$withInitialization$2(LifecycleInitializer.java:288)
	...
	at io.modelcontextprotocol.client.McpSyncClient.initialize(McpSyncClient.java:190)
	at com.epam.aidial.core.server.controller.ToolSetToolsController.fetchTools(ToolSetToolsController.java:206)
	at com.epam.aidial.core.server.controller.ToolSetToolsController.lambda$handle$0(ToolSetToolsController.java:148)
Caused by: java.lang.RuntimeException: Failed to send message: DummyEvent[responseInfo=jdk.internal.net.http.ResponseInfoImpl@4b157bbe]
	at io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport.lambda$sendMessage$30(HttpClientStreamableHttpTransport.java:635)
```

Ingress access log for the upstream, same host and same minute. Core (Java 21.0.11) requests `/mcp` and gets `307`; an MCP SDK client that sets `.endpoint("/mcp/")` requests `/mcp/` and completes the handshake against the very same server:

```
09:33:40  "POST /mcp  HTTP/2"   307   Java-http-client/21.0.11   <- dial-core, fails
09:33:44  "GET  /mcp  HTTP/2"   307   Java-http-client/21.0.11   <- dial-core, fails
09:33:57  "POST /mcp/ HTTP/1.1" 200   Java-http-client/21.0.12   initialize OK
09:33:57  "GET  /mcp/ HTTP/1.1" 200   Java-http-client/21.0.12   SSE stream OK
09:33:57  "POST /mcp/ HTTP/1.1" 202   Java-http-client/21.0.12   notification OK
09:33:57  "POST /mcp/ HTTP/1.1" 200   Java-http-client/21.0.12   tools/list OK
```

A manual `initialize` against `/mcp/` from inside the `dial-core` container returns `200` with `{"serverInfo":{"name":"markitdown","version":"1.8.1"}}`, confirming the upstream is healthy and reachable and that the trailing slash is the only difference.

The `HTTP/2` on Core's requests is itself a symptom of the no-op builder: the SDK's default `clientBuilder` sets `version(HTTP_1_1)`, which the wrapper discards.

### Additional information

**Suggested fix**

Split the configured URL into origin + path and pass the path through `.endpoint(...)`, rather than handing the full URL to `builder(...)`. `ai-dial-admin-deployment-manager-backend` already does exactly this in its `McpClientFactory`, which is why its health check against the same upstream succeeds:

```java
var builder = HttpClientStreamableHttpTransport.builder(baseUrl);
if (StringUtils.isNotBlank(endpointPath)) {
    builder = builder.endpoint(endpointPath);
}
```

Additionally:

- Stop `McpHttpClientBuilderService.HttpClientBuilder` from silently swallowing `followRedirects()` and `version()`. Either honour them or fail loudly — a no-op setter that discards a security- and correctness-relevant policy is a trap for every future caller.
- Set `Redirect.NORMAL` on the shared client so `307`/`308` path normalisation is handled for the SDK-driven paths too, matching what #1702 did for the proxy. Note `Redirect.NORMAL` already refuses HTTPS → HTTP downgrades; the same-origin concern from #1667 (not leaking `Authorization` cross-origin) still deserves a check, since auth headers are attached via `httpRequestCustomizer`.
- Apply the same fix to `SecuredResourceService.validateApiKey()`, which shares the pattern — toolset sign-in against a non-`/mcp` or trailing-slash upstream is broken the same way.

**Impact**

All MCP toolsets whose endpoint path is not exactly `/mcp` are unusable for tool discovery and toolset sign-in, regardless of configuration. On the environment where this was found, 7 of the configured toolset endpoints are affected (4 × `/mcp/`, 2 × `/mcp/delivery_central`, 1 × `/v1/mcp`). The failure mode is opaque — a `502` and a `Failed to send message: DummyEvent[...]` that names neither the status code nor the URL actually requested; logging the resolved request URI and upstream status would have made this self-diagnosing.

astsiapanay:
Toolset tool discovery (`GET /v1/toolset/{name}/tools`) and toolset sign-in fail against any MCP server that responds with a 307/308 redirect (e.g. Starlette/FastMCP trailing-slash normalization), because the shared `java.net.http.HttpClient` these SDK-driven paths use is left at the JDK default `Redirect.NEVER`. [#1771](https://github.com/epam/ai-dial-core/pull/1771) already fixed a related-but-separate bug (the configured endpoint path being discarded) but explicitly left this redirect-following problem as follow-up - this PR closes that remaining gap.

### Applicable issues

- fixes #1768

### Description of changes

- Added `RedirectSafeHttpClient`, a small `HttpClient` decorator that manually follows only **same-origin** 307/308 redirects, mirroring the same-origin guard `McpProxyController` already uses for the MCP proxy path (#1667/#1702). This is deliberately *not* just enabling the JDK's built-in `Redirect.NORMAL`: the JDK only strips the standard `Authorization`/`Cookie`/`Origin`/`Referer`/`Host` headers on a cross-origin redirect, not this codebase's own custom MCP auth headers (`Proxy.HEADER_API_KEY`, and the admin-configured `apiKeyHeader` used for API-key toolset credentials) - so a cross-origin redirect from a compromised/misbehaving MCP upstream could otherwise leak those.
- `McpHttpClientBuilderService` now pins the shared client to HTTP/1.1 (matching the MCP SDK's own default preference, which was otherwise silently lost) and wraps it in `RedirectSafeHttpClient`. Its `HttpClient.Builder` wrapper's `followRedirects()`/`version()` now throw `UnsupportedOperationException` instead of silently no-op-ing, since silently discarding either would be a correctness/security trap for a future caller.
- Extracted `isSameOrigin`/`resolvePort`/`MAX_MCP_REDIRECTS` out of `McpProxyController` into `McpClientUtils` so both the Vert.x proxy path and the new SDK-driven path share identical origin-checking logic instead of duplicating it.
- Added `RedirectSafeHttpClientTest` (including a two-server test proving a cross-origin redirect target receives **zero** requests - the sensitive header never leaves the original origin) and an end-to-end `testGetAllTools_Follows307Redirect` in `ToolSetApiTest` reproducing the reported bug.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/mcp/McpHttpClientBuilder.java
```diff
@@ -88,12 +98,14 @@ public HttpClient.Builder executor(Executor executor) {
 
         @Override
         public HttpClient.Builder followRedirects(HttpClient.Redirect policy) {
-            return this;
+            throw new UnsupportedOperationException(
+                    "the shared HttpClient's redirect policy is fixed and cannot be overridden per caller");
```
DmytroZaichenkoDev: `version()` / `followRedirects()` throwing unconditionally is a tripwire. Nothing calls them today, but the SDK already calls `.version(HTTP_1_1)` on its own default builder - if a future release moves that into build(), every tools/list and sign-in 500s, and the mocked-client tests won't catch it. Suggest throwing only when the value actually differs: 
                                                                           
      if (policy != httpClient.followRedirects()) {                                                                                                                                                                                      
          throw new UnsupportedOperationException(                                                                                                                                                                                       
                  "the shared HttpClient's redirect policy is fixed and cannot be overridden per caller");                                                                                                                               
      }                                                                                                                                                                                                                                  
      return this;

server/src/main/java/com/epam/aidial/core/server/mcp/McpHttpClientBuilder.java
```diff
@@ -88,12 +98,14 @@ public HttpClient.Builder executor(Executor executor) {
...
         }
 
         @Override
         public HttpClient.Builder version(HttpClient.Version version) {
-            return this;
+            throw new UnsupportedOperationException(
+                    "the shared HttpClient's HTTP version is fixed and cannot be overridden per caller");
```
DmytroZaichenkoDev: `version()` throwing unconditionally is a tripwire. Nothing calls them today, but the SDK already calls `.version(HTTP_1_1)` on its own default builder - if a future release moves that into build(), every tools/list and sign-in 500s, and the mocked-client tests won't catch it. Suggest throwing only when the value actually differs: 
                                                                           
      if (version != httpClient.version()) {                                                                                                                                                                                             
          throw new UnsupportedOperationException(                                                                                                                                                                                       
                  "the shared HttpClient's HTTP version is fixed and cannot be overridden per caller");                                                                                                                                  
      }                                                                                                                                                                                                                                  
      return this;

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/mcp/McpHttpClientBuilder.java
astsiapanay: Good catch, thanks — fixed in 1f6aeb45. `followRedirects()` now only throws when the requested policy actually differs from what the shared client already has (`policy != httpClient.followRedirects()`), so it stays a no-op for a matching call. Added `followRedirectsAcceptsTheAlreadyConfiguredPolicyAsNoOp` to cover exactly the scenario you described.

server/src/main/java/com/epam/aidial/core/server/mcp/McpHttpClientBuilder.java
astsiapanay: Same fix applied here — fixed in 1f6aeb45. `version()` now only throws on an actual mismatch (`version != httpClient.version()`), with `versionAcceptsTheAlreadyConfiguredVersionAsNoOp` covering the matching-call case.

DmytroZaichenkoDev (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [30906996861](https://github.com/epam/ai-dial-ci/actions/runs/30906996861)

| [deploy-review](https://chat-ai-dial-core-pr-1775.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30906996861/job/91984294191) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30906996861/job/91986787050) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1776, opened 2026-07-31
feat: support overrideName for apps and interceptors interfaces

khshanovskyi:
### Applicable issues

- feat #1618

### Description of changes

- Moved `overrideName` up from `Model` into the shared `Deployment` base class, so applications and interceptors can also declare it (previously only models could).
- Added `Deployment.getTargetName()` as the single source of truth for the override fallback (`overrideName` if set, otherwise the deployment's own name), and reused it in both `Deployment.resolveRequestUri` (new interfaces flow) and `ChatCompletionInterceptorController.buildUri` (interceptor chat-completions routing) to avoid divergent duplicate logic.
- Wired `overrideName` through `EnhanceDeploymentRequestFn` (replacing the old model-only `EnhanceModelRequestFn`) so the outgoing `model` field and `X-DIAL-OVERRIDE-NAME` header are rewritten for any deployment type, not just models.
- Documented the new field in `applications.md`, `interceptors.md`, `models.md`, and the OpenAPI spec.

### Checklist

- [x] All tests pass (MessagesApiTest 9/9, token parser tests, controller routing tests, checkstyle clean)


server/src/main/java/com/epam/aidial/core/server/util/ProxyUtil.java
```diff
@@ -87,11 +86,12 @@ public static void copyHeaders(MultiMap from, MultiMap to, MultiMap excludeHeade
     }
 
     public static void setOverrideNameHeader(MultiMap headers, Deployment deployment) {
-        if (deployment instanceof Model model) {
-            String overrideName = model.getOverrideName();
-            if (overrideName != null) {
-                headers.set(Proxy.HEADER_OVERRIDE_NAME, overrideName);
-            }
+        if (deployment == null) {
```
khshanovskyi, line 89: @astsiapanay, I have some doubts if its okay to apply it for whole deployment, maybe restrict it with Model, Application and Interceptor?

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/util/ProxyUtil.java
astsiapanay: I would leave logic as is. It's a helper method.

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1780, opened 2026-08-03
fix: Login Pop-Up Broken (No Events Reaching Channel) #1778

### LINKED ISSUE #1778: Login Pop-Up Broken (No Events Reaching Channel)
### Name and Version

0.45.0

The login pop-up window for Toolset has stopped working. Initial frontend/client-side investigation reveals that while it is successful to subscribe to the channel via core, no subsequent events are reaching the channel. This blocks users from completing the authentication flow.

The login pop-up opens, authentication events are successfully published to the channel, and the user is logged in.

### What do you see instead?

The pop-up fails to function. Subscribing to the channel with core works, but no events reach the channel afterward.

### Additional information

_No response_

astsiapanay:
Fixes the client-channel Redis pub/sub subscription silently dying after a period of inactivity, which left the toolset sign-in pop-up subscribed with no events ever reaching it.

### Applicable issues

- fixes #1778

### Description of changes

- Added `RedisTopicWatchdog`, a shared helper used by `RpcRequestTopic` and `RpcResponseTopic` that periodically publishes a canary message to the topic and forces a resubscribe if it's never delivered back to the same listener - recovering from Redis silently dropping an idle subscription connection (e.g. NAT/LB idle timeout) without either side noticing.
- Also attaches a `StatusListener` to log Redisson's own subscribe/unsubscribe events for diagnostics.
- The watchdog check avoids `synchronized` (which would pin virtual-thread carrier threads during the blocking Redis publish call) in favor of a lock-free `AtomicBoolean` guard.
- Added new setting `clientChannel.watchdog.period` (default 10000ms) controlling the watchdog interval, documented in the README.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/clientchannel/RedisTopicWatchdog.java
```diff
@@ -0,0 +1,109 @@
...
+        // and a synchronized block here would pin the carrier thread for virtual threads (JEP 444)
+        if (!watchdogRunning.compareAndSet(false, true)) {
+            return;
+        }
+        try {
+            String missedNonce = pendingCanary.getAndSet(null);
+            if (missedNonce != null) {
+                log.warn("{} watchdog canary {} was not received back. Resubscribing listener {}",
+                        topicName, missedNonce, listenerId);
+                topic.removeListener(listenerId);
```
DmytroZaichenkoDev: This won't reach Redis — the StatusListener at line 60 blocks it.                                                                                                                                                                      
                                                                                                                                                                                                                                         
`removeListenerAsync` only sends UNSUBSCRIBE when `!entry.hasListeners(channelName)`, and the StatusListener sits in that same per-channel queue, so it's never empty. No UNSUBSCRIBE means the channel stays in `name2PubSubConnection`, so line 99 takes `subscribeNoTimeout`'s fast path and sends no SUBSCRIBE either.                                                                                                                              
                                                                                                                                                                                                                                         Result: the listener is swapped in memory on the same connection, zero bytes to Redis.                                                                                                                                                    
                                                                                                                                                                                                                                         
Option: `removeAllListeners()` unsubscribes unconditionally — that plus re-registering both listeners should work. (Redisson 3.27.0.)

server/src/main/java/com/epam/aidial/core/server/service/clientchannel/RedisTopicWatchdog.java
```diff
@@ -0,0 +1,109 @@
...
+                listenerId = registerListener();
+            }
+
+            String nonce = UUID.randomUUID().toString();
+            pendingCanary.set(nonce);
+            topic.publish(canaryFactory.apply(WATCHDOG_CHANNEL_ID, nonce));
```
DmytroZaichenkoDev: If the repair above throws (`removeListener` blocks and can time out on a dead connection), this publish is skipped. Then `pendingCanary` stays null, so the next watchdog run finds nothing to check and detection only works every  second run. Putting the publish in a `finally` fixes it.

server/src/main/java/com/epam/aidial/core/server/service/clientchannel/RedisTopicWatchdog.java
astsiapanay: Confirmed against the Redisson 3.27.0 source — `removeListenerAsync` only unsubscribes when `!entry.hasListeners(channelName)`, and the StatusListener kept that non-empty, so `subscribeNoTimeout` then took the existing-`name2PubSubConnection`-entry fast path with no actual SUBSCRIBE. Switched to `removeAllListeners()` + re-registering both listeners in 7ebafd41.

server/src/main/java/com/epam/aidial/core/server/service/clientchannel/RedisTopicWatchdog.java
astsiapanay: Fixed — moved the fresh canary publish into a `finally` so it always runs even if the resubscribe attempt above throws, in 7ebafd41.

ai-dial-actions:
<h1>Dependency Review</h1>
None

DmytroZaichenkoDev (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [30895758267](https://github.com/epam/ai-dial-ci/actions/runs/30895758267)

| [deploy-review](https://chat-ai-dial-core-pr-1780.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30895758267/job/91948131490) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/30895758267/job/91950305631) |

PR #1792, opened 2026-08-05
feat: Enhance token usage statistics #1753

### LINKED ISSUE #1753: Enhance token usage statistics
### Name and Version

core 0.47

At the moment DIAL Core aggregates token usage spent by LLMs for DIAL Applications and writes stats to the same field `tokenUsage` that confuses users. There is a lack of a field which could report aggregated token usage stats by their descendants in the call tree. At the same time DIAL applications may report token usage stats too.

Introduce a new top level field in the chat completion response `statistics.usage_per_model`:

**openapi-generator/src/main/resources/schemas/Statistics.yaml**
```YAML
type: object
properties:
  usage_per_model:
    type: array
    items:
      $ref: '#/components/schemas/UsagePerModel'
  discarded_messages:
    type: array
    items:
      type: integer
```
**openapi-generator/src/main/resources/schemas/UsagePerModel.yaml**
```YAML
type: object
properties:
  index:
    type: integer
  model:
    type: string
  prompt_tokens:
    type: integer
  completion_tokens:
    type: integer
  total_tokens:
    type: integer
  prompt_tokens_details:
    $ref: '#/components/schemas/PromptTokensDetails'
  completion_tokens_details:
    $ref: '#/components/schemas/CompletionTokensDetails'
```

The new field `usage_per_model` should contain aggregated statistics of token usage to be provided by all descendant deployments from the call tree.

The field should be written to analytics logs(See GfLogStore) and Chat Completion response.

- DIAL Core shouldn't put aggregated info to the field `token_usage` in analytics log any longer. Use `usage_per_model` instead of.
- DIAL Core should provide the field `statistics.usage_per_model` to a response both for streaming and non-streaming mode(last chunk).
- Any deployment(Model or Application) may report token usage in chat completion response. So DIAL Core should capture stats for any deployment if the deployment returns non empty token usage object and update the stats `usage_per_model ` for ascendents.
- DIAL Core should override the field value if it's already returned by application in response.
- DIAL Core should recursively update the stats `usage_per_model` from the current node to all ascendents (till the root).

_No response_

astsiapanay:
Adds a new `statistics.usage_per_model` field to Chat Completion, Anthropic Messages, and Responses API responses, giving a per-model token usage breakdown of a call tree's *descendant* DIAL deployments, instead of conflating it into `token_usage` in the analytics log. Core's value always overrides whatever a deployment already supplied.

### Applicable issues

- fixes #1753

### Description of changes

- `TokenStatsTracker` maintains a per-deployment-name usage map per trace span, but only ever merges a *descendant's* report into it as it walks up to each ancestor - a deployment never adds itself to its own stored map, and its own outward-facing response is never supplemented with its own entry either. `usage_per_model` reflects only what DIAL-tracked descendants below the reporting deployment contributed; it never duplicates the reporting deployment's own top-level `usage`. A bare Model call, or an Application whose upstream isn't itself a nested DIAL deployment, therefore has no `usage_per_model` field at all - there's nothing beyond its own top-level `usage` to add. The same model called more than once anywhere in the tree still collapses into one entry at each shared ancestor.
- A span's own `tokenUsage` is *set* to whatever it self-reports (Model or Application), never merged with what its descendants reported. The one exception is `aggCost`, which keeps accumulating into every ancestor (and into a self-reporting span's own total) so subtree cost/billing visibility survives regardless of self-report order.
- `TokenUsageParser`'s backward byte-scan for a deployment's own `usage` object now excludes nested `statistics.usage_per_model[].usage` ranges, so it can no longer misparse a deployment's own usage once nested usage blocks routinely appear (an unavoidable consequence of the new schema shape).
- Any deployment (Model or Application) with non-empty usage in its own response is captured; the analytics log's `token_usage` for non-Model deployments now reflects that deployment's own usage instead of its subtree's conflated aggregate.
- `statistics.usage_per_model` is injected into every qualifying response that has a non-empty descendant breakdown to report (chat completions, Anthropic Messages, Responses API), streaming and non-streaming, always overriding any value a deployment already supplied. `/embeddings` and legacy `/completions` are explicitly out of scope and unaffected.
- Added `usage_per_model` to the `GfLogStore` analytics log entry.
- Updated the `UsagePerModel` OpenAPI schema to `{index, model, usage}` (nested shape) and added `statistics` to the non-streaming `CreateChatCompletionResponse` schema; regenerated `docs/open_api_core.yaml`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

openapi-generator/src/main/resources/schemas/UsagePerModel.yaml
```diff
@@ -4,13 +4,5 @@ properties:
...
-    type: integer
-  completion_tokens:
-    type: integer
-  total_tokens:
-    type: integer
-  prompt_tokens_details:
-    $ref: '#/components/schemas/PromptTokensDetails'
-  completion_tokens_details:
-    $ref: '#/components/schemas/CompletionTokensDetails'
+  usage:
```
adubovik: It add a **new field**, we should rather reuse it:

```yaml
type: object
allOf:
  - $ref: "#/components/schemas/Usage"
  - type: object
    properties:
      index:
        type: integer
      model:
        type: string
```

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
```diff
@@ -211,6 +253,51 @@ protected TokenUsage parseTokenUsage(Buffer responseBody) {
...
 
+    /**
+     * Rewrites {@code body} to carry {@code statistics.usage_per_model} when there's something to
+     * report; returns {@code body} unchanged (byte-identical) otherwise, including on any parse
+     * surprise - injection is best-effort, never a reason to corrupt or drop a response. Shared by
+     * every non-streaming response path (chat completions, Anthropic Messages, Responses API).
+     */
+    protected Buffer maybeInjectUsagePerModel(Buffer body) {
+        List<UsagePerModel> usagePerModel = context.getUsagePerModel();
+        if (usagePerModel == null || usagePerModel.isEmpty()) {
```
adubovik: This means that an adapter can return a fake `usage_per_model` and we pass it to the user as-is. We should rather remove it altogether.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/token/UsagePerModel.java
```diff
@@ -0,0 +1,25 @@
...
+ * accumulated and {@link Usage} for why the nested breakdown isn't just a {@link TokenUsage}.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UsagePerModel {
+    private Integer index;
+    private String model;
+    private Usage usage;
```
adubovik: It supposed to be inlined

server/src/main/java/com/epam/aidial/core/server/token/TokenUsageParser.java
```diff
@@ -68,6 +84,111 @@ private int findUsage(Buffer body) {
         return -1;
     }
 
+    /**
+     * Locates every {@code "statistics": { ... }} value in the buffer (there may be more than one
+     * across concatenated SSE chunks) and returns their {@code [start, end)} byte ranges, {@code start}
+     * being the value's opening {@code {} and {@code end} the index right after its matching {@code }}.
+     * Bounded to the size of the statistics blocks themselves, not the whole body.
+     */
+    private List<int[]> findStatisticsRanges(Buffer body) {
```
adubovik: You don't need this method, since the "usage" field isn't expected to occur in `statistics` anyway.

docs/open_api_core.yaml
```diff
@@ -15703,16 +15705,8 @@ components:
...
-          type: integer
-        completion_tokens:
-          type: integer
-        total_tokens:
-          type: integer
-        prompt_tokens_details:
-          $ref: "#/components/schemas/PromptTokensDetails"
-        completion_tokens_details:
-          $ref: "#/components/schemas/CompletionTokensDetails"
+        usage:
```
adubovik: It should be inlined.

server/src/test/java/com/epam/aidial/core/server/MessagesApiTest.java
```diff
@@ -51,12 +52,26 @@ public void testNonStreamingRoundTripAndUsage() throws IOException, InterruptedE
                     "api-key", "proxyKey1", "anthropic-version", "2023-06-01");
 
             assertEquals(200, response.status());
-            assertEquals(NON_STREAM_RESPONSE, response.body());
             // exact ingress path is forwarded (base_url + request.uri())
             assertEquals(MESSAGES_PATH, captured.get().getPath());
             // Anthropic-specific headers pass through to the adapter
             assertEquals("2023-06-01", captured.get().getHeader("anthropic-version"));
 
+            // body content is unchanged; Core additionally injects statistics.usage_per_model -
```
adubovik: I don't think we need to support this for the Messages API.
`statistics` - is an extension of the Chat Completions API specifically.

server/src/test/java/com/epam/aidial/core/server/ResponsesApiTest.java
```diff
@@ -55,6 +56,38 @@ public void testResponsesApi() throws IOException, InterruptedException {
         }
     }
 
+    @Test
+    public void testResponsesApiStreaming_InjectsUsagePerModel() throws IOException {
```
adubovik: I don't think we need to support this for the Responses API.
`statistics` - is an extension of the Chat Completions API specifically.

server/src/main/java/com/epam/aidial/core/server/token/TokenStatsTracker.java
```diff
@@ -107,37 +124,69 @@ void addSpan(ProxyContext context) {
...
+            return result;
+        }
     }
 
     @Data
+    @JsonIgnoreProperties(ignoreUnknown = true)
     public static class TokenStats {
         TokenUsage tokenUsage;
         String parentSpanId;
+        Map<String, TokenUsage> usagePerModel = new LinkedHashMap<>();
```
adubovik: It should be a list, not a map. Two callsID into the same deployment id should result in **two separate** entries in the `usage_per_model` list.

adubovik (changes requested):

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
astsiapanay: No, the method injects the usage provided by Core. The Core removes any usage provided by an adapi

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
astsiapanay: It doesn't work for non-streaming responses. needs to fix

server/src/main/java/com/epam/aidial/core/server/token/TokenStatsTracker.java
astsiapanay: confirmed. we shouldn't merge usage stats.

openapi-generator/src/main/resources/schemas/UsagePerModel.yaml
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/token/UsagePerModel.java
astsiapanay: fixed

docs/open_api_core.yaml
astsiapanay: fixed

server/src/test/java/com/epam/aidial/core/server/MessagesApiTest.java
astsiapanay: fixed

server/src/test/java/com/epam/aidial/core/server/ResponsesApiTest.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/token/TokenStatsTracker.java
astsiapanay: fixed

server/src/main/java/com/epam/aidial/core/server/token/TokenUsageParser.java
astsiapanay: fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

adubovik (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [31153281111](https://github.com/epam/ai-dial-ci/actions/runs/31153281111)

| [deploy-review](https://chat-ai-dial-core-pr-1792.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31153281111/job/92787245258) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31153281111/job/92788753342) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #786, opened 2025-04-10
feat: Allow sequential calls of schema-rich applications, enabling access to their containing files during completion calls

sergey-zinchenko:
### Description of changes

This PR enables schema-rich applications to be called sequentially by enabling read access to `dial:file` marked resources during completion calls to the schema-rich applications.

### Checklist

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1793, opened 2026-08-05
fix: chained agent cannot reach its own sub-deployments #1790

### LINKED ISSUE #1790: Agent cannot call its own sub-agent when it is invoked by another agent (403)
### Name and Version

core 0.47.0-dev.25

1. Create agent **C** (for example a knowledge-base agent) and keep it private — do not publish it.
2. Create agent **B** and add agent **C** to it.
3. Create agent **A** (an orchestrator) and add agent **B** to it.
4. Talk to agent **A** and ask something that makes it delegate to agent **B**.

All three agents belong to the same user.

Agent A calls agent B, agent B calls agent C, and the answer comes back to the user.

If the user can use an agent directly, that agent should keep working when it is reached through another agent.

### What do you see instead?

The request fails with `403 Forbidden` as soon as agent A delegates to agent B:

```
Access denied to the deployment applications/<bucket>/C
```

Agent B is never able to reach agent C.

### Additional information

- The problem appears only when an agent is invoked **by another agent**. Calling agent B directly from the chat works, and so does a single orchestrator with its agents attached directly to it.
- Publishing agent C (making it public) avoids the problem, so the failure is limited to private, personally owned sub-agents.
- Ownership is not the issue — the user owns all three agents. A sub-agent that was *shared* to the user by someone else is reachable in the same position; only the user's own sub-agent is refused.
- Files behave correctly in exactly this situation: an agent invoked by another agent can still read its own attached files. Only sub-agents are refused. The equivalent behaviour for files was added in #786, and sub-agents appear to have been left out of it.
- This blocks any orchestrator setup more than two levels deep, which is becoming a common way to build agents (see epam/ai-dial-quickapps-backend#426).

DmytroZaichenkoDev:
A schema-rich application could not reach the sub-agents and toolsets it declares once it was invoked by another agent rather than directly by the user. The access rule that exists for exactly this situation covered an application's declared files but never its declared deployments.

### Applicable issues

- fixes #1790

### Description of changes

- `AccessService.getOwnResourcesAccessForChainedSchemaRichApplication` now also consults `ApplicationSchemaService.getDeployments()`, not just `getFiles()`. The grant stays `READ_ONLY` and stays scoped to resources the application actually declares in the initiator's bucket, so nothing is widened beyond what the files path already allowed.
- Why the rule is the right place: on the second hop the request runs under a per-request key, so `BucketBuilder.buildUserBucket()` resolves to `Keys/<deployment>/` and the user's ownership of the sub-deployment becomes invisible. This rule is the one that uses `buildInitiatorBucket()` instead, which is why it already worked for files. It was added for chained schema-rich applications in #786 and deployments were simply left out of it.
- The per-hop key propagation itself is correct and unchanged: each hop populates the next hop's key with its own declared deployments, so the chain is unbounded in depth once the authorization step stops failing.
- Both `APPLICATION` and `TOOL_SET` are covered, mirroring what `getDeployments()` has always returned — a private declared toolset failed at depth two for the same reason a private sub-agent did.
- Collecting declared resources re-validates the application against its schema, so the rule now returns early when no requested resource lives in the initiator's bucket, and asks only for the resource kinds actually being looked up. This makes the common file lookup cheaper than before rather than more expensive.

### Tests

- Five unit tests in `AccessServiceTest`: declared sub-agent granted, declared toolset granted, undeclared deployment refused, deployment in another bucket refused, files and deployments together.
- New integration test `CustomApplicationApiTest.testChainedSchemaRichApplicationCanReachOwnDeployment` drives the real flow — a private sub-agent, a schema-rich parent declaring it, and a chat completion to the parent under a per-request key carrying only the parent. It returns 403 without the fix and 200 with it.
- This path had no integration coverage before: `CollectDeploymentsFnTest` stubs `hasReadAccess` to `true`, and the existing per-request-key tests inject an already-populated grant, so the authorization step in between was untested from both directions.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
@@ -320,7 +320,7 @@ public class AccessService {
         }
 
         // Collecting declared resources re-validates the application against its schema, so ask only for the kinds actually requested.
-        List<ResourceDescriptor> declaredResources = new ArrayList<>();
+        Set<ResourceDescriptor> declaredResources = new HashSet<>();
         if (ownResources.stream().anyMatch(resource -> resource.getType() == ResourceTypes.FILE)) {
             declaredResources.addAll(applicationSchemaService.getFiles(application));
         }
--- a/server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
@@ -18,6 +18,7 @@ import com.epam.aidial.core.server.util.ResourceDescriptorFactory;
 import com.epam.aidial.core.storage.data.MetadataBase;
 import com.epam.aidial.core.storage.data.ResourceFolderMetadata;
 import com.epam.aidial.core.storage.resource.ResourceDescriptor;
+import com.epam.aidial.core.storage.resource.ResourceType;
 import com.epam.aidial.core.storage.resource.ResourceTypes;
 import com.epam.aidial.core.storage.resource.ResourceUtil;
 import com.google.common.annotations.VisibleForTesting;
@@ -53,6 +54,8 @@ public class AccessService {
     private final List<String> createCodeAppRoles;
 
     private final ApplicationSchemaService applicationSchemaService;
+
+    private static final Set<ResourceType> DEPLOYMENT_TYPES = Set.of(ResourceTypes.APPLICATION, ResourceTypes.TOOL_SET);
 
     private final List<PermissionRule> permissionRules = List.of(
             AccessService::getOwnResourcesAccess,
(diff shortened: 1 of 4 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/security/AccessService.java
```diff
@@ -305,19 +308,33 @@ private static Map<ResourceDescriptor, Set<ResourceAccessType>> getOwnResourcesA
...
+        String location = BucketBuilder.buildInitiatorBucket(context);
+        Set<ResourceDescriptor> ownResources = resources.stream()
+                .filter(resource -> resource.getBucketLocation().equals(location))
+                .collect(Collectors.toSet());
+        if (ownResources.isEmpty()) {
+            return Map.of();
+        }
+
+        // Collecting declared resources re-validates the application against its schema, so ask only for the kinds actually requested.
+        List<ResourceDescriptor> declaredResources = new ArrayList<>();
```
astsiapanay: HashSet would be more effective here

DmytroZaichenkoDev: Good catch — Fixed

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1794, opened 2026-08-06
feat: configurable claims and operation_duration_ms in analytics logs

Pasichniuk:
### Applicable issues

- _

### Description of changes

Analytics log entries carried a fixed set of claims — `user_id`, `roles`, `user_display_name` — and no latency figure, so end-to-end duration was only recoverable by correlating log timestamps. This PR makes the collected claims configurable and adds request duration as an always-present field.

### Configurable claims

**`analytics.claimsAllowlist`** (default empty) lists claims to add to the `claims` object of each analytics log entry, taken verbatim from the identity provider's token or user info response. Each entry is a dot-separated path, emitted under that exact path as its key:

```jsonc
"analytics": {
  "collectClaims": true,
  "claimsAllowlist": ["email", "resource_access.roles"]
}
```
```json
"claims": {"user_id": "...", "roles": [...], "email": "jane.doe@example.com", "resource_access.roles": ["a", "b"]}
```

- `["*"]` collects every top-level claim. Malformed paths (`"."`, `"a..b"`, `".email"`) are rejected at startup with a warning — `"."` would otherwise resolve to the root of the payload and dump every claim without the explicit `"*"` opt-in.
- One list covers several providers that name the same thing differently (`email`, `upn`, `preferred_username`); missing claims are simply omitted, as for API key requests, which carry no JWT claims.
- Independent of `analytics.collectClaims`: either setting alone emits the `claims` object, and each contributes only its own members. An allowlist entry named `user_id`, `roles` or `user_display_name` is skipped only when that member was already written from the fixed set, so the object can never carry a duplicate key.
- Values keep their JSON type; anything whose JSON exceeds 4 KB is emitted as a truncated string ending in `>>`, matching how oversized headers and bodies are already reported.
- `user.id` is unchanged, so existing user-level joins and dashboards stay continuous. Interceptor and application hops log the same claims, and async job completions match live requests — `BackgroundJobRecord` now carries the claim payload the way it already carries raw request headers: stored whole, filtered on write.

> [!NOTE]
> Nothing new is *collected* — `ExtractedClaims.userClaims` has always held the whole payload and already travels with per-request key data. The setting only decides what reaches the log line, and the values are written unhashed, so treat it as personal data in logs. Where `identityProviders.*.loggingKey` points at the email claim and `obfuscateUserEmail` is enabled (the default), `user.id` already identifies the same user as a salted hash.                                                            
### Operation duration                                                                                                                                                           
**`operation_duration_ms`** on every entry: wall-clock ms from request receipt to the response body being complete, never negative, matching the total in the existing `Timing:` info log. Paths that respond without ever producing a body fall back to emias no response body to measure against, so it runs to the moment the entry is emitted, which includes the poller's detection lag — up to `backgroundJob.maxPollIntervalMs` once the backoff has grown.                                                      
Four controllers — MCP proxy, deployment features, and both route paths — never set a response-body timestamp and would have reported a wider span; with those added, all twelve call sites measure the same point.

> [!WARNING]
> A background job that never reaches a terminal state expires, and the core writes a synthetic entry with `response.status` `500` whose `operation_duration_ms` is the whole
time it waited before giving up — `backgroundJob.jobTtlMs` (24h by default)dJob.scanIntervalMs` tick (10 min), not a real response latency. Filter those rows out of latency aggregates.

### Other

- New `AnalyticsSettings` record replaces `GfLogStore`'s four positional parameters, two of them adjacent booleans that could swap silently.
- Claim member names are JSON-escaped, since they now come from configuration and from the token payload rather than being hardcoded. The shared helper that writes user_id/user_display_name was made consistent — no output change, since those literals need no escaping.

### Checklist

- [x] Title of the pull request follows [Conventional Commits specificationcommits.org/en/v1.0.0/)

By submitting this pull request, I confirm that my contribution is made und 2.0 license.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
```diff
@@ -26,6 +27,8 @@
  *     </ul>
  * </p>
  */
+// tolerate fields written by a newer core version, so per-request keys survive a rolling upgrade
+@JsonIgnoreProperties(ignoreUnknown = true)
```
siarhei-fedziukovich: why we need it in scope of this PR?

server/src/main/java/com/epam/aidial/core/server/security/ExtractedClaims.java
```diff
@@ -1,13 +1,16 @@
...
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.List;
 
+// tolerate members written by a newer core version, so per-request keys survive a rolling upgrade
+@JsonIgnoreProperties(ignoreUnknown = true)
 public record ExtractedClaims(String userId, List<String> userRoles, String userHash,
-                              ObjectNode userClaims, String project, String userDisplayName) {
+                              ObjectNode userClaims, String project, String userDisplayName, String userEmail) {
```
siarhei-fedziukovich: does it mean that core collect only predefined set of claims? and here you just extend predefined set of claims with new `userEmail` entry? if so than let's better support whitelist (with ability to declare `*` as all claims) for extraclaims (per provider). and operators/clients will be able to collect as much claims as needed

server/src/main/java/com/epam/aidial/core/server/log/AnalyticsLogContext.java
```diff
@@ -123,9 +129,36 @@ public static AnalyticsLogContext from(BackgroundJobRecord record, ResponsesApiC
                 .responseStatusCode(result == null ? 500 : 200) // 500 when expired
                 .responseBody(result == null ? null : result.body())
                 .tokenUsage(result == null ? null : result.usage())
+                .operationDurationMs(operationDurationMs(record, result))
                 .build();
     }
 
+    /**
+     * The job is finalized once the poller notices it reached a terminal state, which may lag the actual completion by
```
siarhei-fedziukovich: do we really want to parse upstream/response completionTs? i think we measure core request->response latency/duration. not coreRequest->upstreamResponse

server/src/main/java/com/epam/aidial/core/server/security/ExtractedClaims.java
Pasichniuk: Good catch, and no — nothing about claim collection is predefined. `IdentityProvider.extractUserClaims` is `MAPPER.valueToTree(map)`, so `ExtractedClaims.userClaims` already holds the entire JWT payload / userinfo body, and it already travels with the per-request API key data. What was predefined is only the set the analytics log *writes*.

So the whitelist belongs on the emit side, and that is what it is now:

```jsonc
"analytics": {
  "collectClaims": true,
  "claimsAllowlist": ["email", "resource_access.roles"]
}
```

- dot-separated paths, emitted under that exact path as the key
- `["*"]` collects every top-level claim
- values keep their JSON type, so arrays/objects/numbers work
- independent of `collectClaims`; entries colliding with `user_id`/`roles`/`user_display_name` are skipped so the object can never carry a duplicate key
- absent claims are omitted, so one list can cover several IdPs that name the same thing differently (`email`, `upn`, `preferred_username`)

`userEmailPath`, `collectEmail` and `ExtractedClaims.userEmail` are all gone.

server/src/main/java/com/epam/aidial/core/server/data/ApiKeyData.java
Pasichniuk: Fair — removed, from `ExtractedClaims` too. It was there because this PR was adding a field to `ExtractedClaims`, which is serialized into `ApiKeyData` in Redis, so during a rolling deploy an old pod would reject the record. But the annotation only protects *future* additions (old pods do not have it), so it did nothing for this PR anyway. With the claim allowlist reading from the existing `userClaims` node, no field is added at all and the concern is moot.

server/src/main/java/com/epam/aidial/core/server/log/AnalyticsLogContext.java
Pasichniuk: Agreed — dropped. `operation_duration_ms` is now core request-receipt → log emit on every path, background jobs included; `ResponsesApiClient` no longer parses `completed_at`.

For the record, the reason it was there: the poller finalizes a job only when it next observes a terminal state, and with backoff up to `maxPollIntervalMs` (5 min default) that can inflate an async job's duration by minutes. Consistency of the metric wins over that, so the lag is documented on the field instead of corrected for.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

siarhei-fedziukovich (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/ProxyContext.java
```diff
@@ -221,6 +222,23 @@ public boolean isSecuredApiKey() {
...
+    public ObjectNode getUserClaims() {
+        return extractedClaims == null ? null : extractedClaims.userClaims();
+    }
+
+    /**
+     * How long the core worked on the request: from accepting it to having the full response body. Paths that never
+     * produce a response body, e.g. a deployment without an endpoint, fall back to the current time. Both bounds are
+     * wall-clock, so a backwards clock adjustment in between must not produce a negative duration.
+     */
+    public long getOperationDurationMs() {
```
astsiapanay: get means the result remaining the same over multiple calls
The better name is calculate

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1796, opened 2026-08-06
fix: rewrite multi-segment deployment id in upstream path #1795

### LINKED ISSUE #1795: Deployment path-segment rewrite mangles multi-segment (platform-bucket) deployment ids
### Name and Version

ai-dial-core (development branch, current HEAD)

1. Configure a model that is addressed by its platform-bucket canonical id, e.g. stored/derivable under the key `models/platform/gemini-3.1-pro-preview`, with no interceptors:
```json
{
  "name": "gemini-3.1-pro-preview",
  "interceptors": [],
  "interfaces": {
    "openaiChatCompletions": {
      "base_url": "http://dial-vertexai.dial.svc.cluster.local.:80"
    }
  }
}
```
2. Call `POST /openai/deployments/models/platform/gemini-3.1-pro-preview/chat/completions?api-version=2025-01-01-preview`
3. Observe the URL Core forwards to the upstream `base_url`.

```
http://dial-vertexai.dial.svc.cluster.local/openai/deployments/gemini-3.1-pro-preview/chat/completions?api-version=2025-01-01-preview
```
The multi-segment ingress id collapses to the deployment's own name (or `overrideName`), same as it does for a single-segment id today.

### What do you see instead?

```
http://dial-vertexai.dial.svc.cluster.local/openai/deployments/gemini-3.1-pro-preview/platform/gemini-3.1-pro-preview/chat/completions?api-version=2025-01-01-preview
```
The `/platform/gemini-3.1-pro-preview` remainder of the original id is left dangling in the path, producing an invalid upstream URL.

### Additional information

`Config.resolve` (`config/.../Config.java`) and `RouteTemplate.POST_DEPLOYMENT` (`^/+openai/deployments/(?<id>.+?)/(completions|chat/completions|embeddings)$`) both treat the deployment id as an arbitrary, potentially multi-segment string — platform-bucket entities are addressed as `models/platform/{name}`. Lookup and routing handle this correctly.

The path-rewrite logic that swaps the ingress id segment for the deployment's target/override name assumes the id is exactly one path segment, though:

- **Primary / default path** (no interceptors — the common case): `Deployment.DEPLOYMENT_SEGMENT` in `config/src/main/java/com/epam/aidial/core/config/Deployment.java`: `/deployments/[^/]+/`, used by `rewriteDeploymentPathSegment`, called from `resolveRequestUri`, called from `BaseDeploymentPostController.createProxyRequest(InterfaceType)` — the terminal call issuing the actual outgoing request for every chat-completions call without interceptors.
- **Secondary occurrence** (only when interceptors are configured): `ChatCompletionInterceptorController.DEPLOYMENT_PATH` in `server/.../ChatCompletionInterceptorController.java`: `(/openai/deployments/)([^/]+)(/.*)`, same bug, different call site (`handleInterceptor` → interceptor chain).

For a multi-segment id, `[^/]+` only consumes the first segment (`models`), so only that segment gets replaced and the remaining segments of the original id are left dangling.

Not affected: Responses API and Anthropic Messages flows — those resolve the deployment from the request body, not the URL path, so they never touch a `/deployments/{id}/` segment.

KirylKurnosenka:
Fixes the deployment-id path rewrite so it collapses a full multi-segment canonical deployment id (e.g. `models/platform/{name}` for platform-bucket entities) to the target deployment name, instead of only replacing its first path segment.

### Applicable issues

- fixes #1795

### Description of changes

- `config/src/main/java/com/epam/aidial/core/config/Deployment.java`: widened `DEPLOYMENT_SEGMENT` to capture the ingress deployment id non-greedily up to the trailing `completions`/`chat/completions`/`embeddings` action literal (with a `(?=$|\?)` lookahead, since the ingress URI may carry a query string), instead of stopping at the first `/`. This is the primary code path — reached via `resolveRequestUri` from `BaseDeploymentPostController.createProxyRequest(InterfaceType)` — used for every chat-completions request that has no interceptors configured.
- `server/src/main/java/com/epam/aidial/core/server/controller/ChatCompletionInterceptorController.java`: applied the identical fix to `DEPLOYMENT_PATH`, since it has the same bug on the interceptor-chain code path.
- Added regression tests in `DeploymentTest.java` and `ChatCompletionInterceptorControllerTest.java` covering a multi-segment ingress id, asserting it collapses to a single target-name segment in the rewritten URL. Existing single-segment test cases are unaffected.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1808, opened 2026-08-10
feat: add bulk user limits API #1799

### LINKED ISSUE #1799: Add a bulk API for user deployment limits and rolling usage
### Name and Version

DIAL Core 0.47 (development branch, commit `31b38d72`)

Users need a single view of their token limits and current usage across deployments for daily, weekly, and monthly windows.

DIAL Core currently exposes:

```http
GET /v1/deployments/{deployment_name}/limits
```

This endpoint returns `LimitStats` for one known deployment. A client that needs to display a user usage dashboard must:

1. Retrieve the list of deployments.
2. Send a separate limits request for every deployment.
3. Merge the responses on the client side.

This produces an N+1 request pattern and does not provide a way to efficiently discover deployments for which the caller already has stored usage.

DIAL Core already maintains caller-scoped rate-limit aggregates in `ResourceTypes.LIMIT`:

- Token and request usage is stored per deployment.
- Token usage has rolling minute, day, week, and month buckets.
- Cost usage is stored globally for the caller across all models.

These records contain rolling aggregates rather than raw request history, but they are sufficient for reporting usage for the requested windows.

Add an authenticated bulk endpoint, for example:

```http
GET /v1/user/limits
```

The endpoint should return the effective limits and current rolling usage for the deployments available to the authenticated caller.

Example response:

```json
{
  "deployments": [
    {
      "id": "gpt-4o",
      "tokenStats": {
        "day": {
          "total": 100000,
          "used": 42000
        },
        "week": {
          "total": 500000,
          "used": 180000
        },
        "month": {
          "total": 2000000,
          "used": 640000
        }
      }
    }
  ],
  "costStats": {
    "day": {
      "total": 100.0,
      "used": 14.5
    },
    "week": {
      "total": 500.0,
      "used": 82.3
    },
    "month": {
      "total": 2000.0,
      "used": 310.7
    }
  }
}
```

The exact response schema may reuse the existing `LimitStats`, `ItemLimitStats`, and `CostItemLimitStats` types.

#### Acceptance criteria

- The response is scoped to the authenticated initiator: a JWT user or an API-key project. The endpoint must not accept an arbitrary user identifier.
- Effective limits are resolved using the same role and default-role logic as the existing per-deployment endpoint.
- Day, week, and month usage follows the existing rolling-window semantics: 24 hours, 7 days, and 30 days.
- Each deployment entry contains both the effective limit and used token count for every supported window.
- Deployments with a finite configured limit but no usage are returned with `used: 0`.
- Deployments with persisted usage are discoverable without issuing one request per deployment.
- Only deployments currently accessible to the caller are returned; revoked or inaccessible deployment names must not be exposed.
- User-level cost limits are returned once at the top level because cost usage is currently aggregated across all models, rather than per deployment.
- Existing `RateLimiter` state is used as the source of truth. Analytics logs and transient per-trace `TokenStatsTracker` records must not be required.
- The endpoint and response schemas are documented in OpenAPI.
- Unit and API tests cover role resolution, empty usage, multiple deployments, access filtering, and rolling-window values.

Application usage that exists only in transient per-trace statistics is outside the persisted rate-limit history. Supporting such applications would require persistent aggregation before they can be included in this API.

- Continue using `GET /v1/deployments` followed by `GET /v1/deployments/{deployment_name}/limits` for every deployment. This causes an N+1 request pattern and cannot efficiently identify only deployments with stored usage.
- Add user-specific usage fields to `GET /v1/deployments`. This would couple deployment discovery with rate-limit storage and make the general listing endpoint more expensive.
- Calculate usage from analytics or prompt logs. These logs may be external, disabled, or unavailable for secured requests, while Core already stores the rolling aggregates required for this feature.

oleksii-donets:
### Applicable issues

- fixes #1799

### Description of changes

Adds two endpoints that report the authenticated caller's limits and rolling usage across deployments in one
request, replacing the N+1 pattern of `/v1/deployments` followed by a per-deployment `/limits` call.

- **`GET /v1/user/limits`** — every deployment the caller can access, used or not. One response labels an
  entire model picker, so switching model needs no refetch.
- **`GET /v1/user/usage`** — the same response with the all-zero entries dropped, leaving what was actually
  used in the trailing 30 days. Same field names, same figures; a client changes nothing but the set it
  iterates.

> **This description was rewritten after review.** Earlier revisions of this PR described a read-only
> endpoint returning `deployments` as a *list*. Both have changed: the response is keyed by deployment id,
> and per-deployment cost attribution adds one write to the completion path. See **Cost** below.

```json
{
  "deployments": {
    "chat-gpt-35-turbo": {
      "minuteTokenStats": { "total": 100000,   "used": 1200 },
      "dayTokenStats":    { "total": 10000000, "used": 42000 },
      "weekTokenStats":   { "total": 9223372036854775807, "used": 180000 },
      "monthTokenStats":  { "total": 9223372036854775807, "used": 640000 },
      "hourRequestStats": { "total": 9223372036854775807, "used": 3 },
      "dayRequestStats":  { "total": 9223372036854775807, "used": 12 },
      "minuteCostStats":  { "total": 9223372036854775807, "used": 0.02 },
      "dayCostStats":     { "total": 9223372036854775807, "used": 1.85 },
      "weekCostStats":    { "total": 9223372036854775807, "used": 6.2 },
      "monthCostStats":   { "total": 9223372036854775807, "used": 21.4 }
    },
    "gpt-4": { "...": "accessible but never used: real limits, every used at zero" }
  },
  "minuteCostStats": { "total": 10.0,   "used": 0.4 },
  "dayCostStats":    { "total": 100.0,  "used": 14.5 },
  "weekCostStats":   { "total": 500.0,  "used": 82.3 },
  "monthCostStats":  { "total": 2000.0, "used": 310.7 }
}
```

Each entry is a full `LimitStats` — the type `/v1/deployments/{deployment_name}/limits` already returns — so
`ItemLimitStats` and `CostItemLimitStats` are reused unchanged and clients share one parser and one rendering
path across all three endpoints. Keyed by id rather than listed: ids are unique, lookup of the selected model
needs no scan, and there is no sort order for a client to depend on.

Both are initiator-scoped like the existing endpoint — the caller is a JWT user or an API-key project, and
there is no user parameter. Neither takes a filter; a `?deployments=a,b` parameter is designed but
deliberately deferred, and its absence changes nothing specified here.

#### Models only

`collectTokenUsage` guards the whole `increase` call on `instanceof Model`, and a non-model deployment's
self-reported usage goes to a `deployment_cost_stats/` document deleted when the span ends, so `{app}/tokens`
is never written. `ModelCostCalculator.calculate` returns `null` for any non-`Model`, so no cost is attributed
either. Listing applications would therefore advertise a configured `total` against a `used` that can never
move — a limit that cannot fire. The sub-model tokens and spend an application drives **are** reported,
against the model that incurred them.

`{app}/requests` records *can* exist, since `limit()` is called for non-models on the MCP, Responses and
Anthropic paths. Such a name is excluded for free: the key set comes from `config.getModels()`, so a record
found in storage with no matching accessible model is never reported.

The field is named `deployments`, not `models`, so that applications and toolsets can widen what is returned
without a contract change once their usage is recorded.

#### Cost: budget once, spend per deployment

The two halves of cost are scoped differently. **The budget is global** — `Role.costLimit` has no deployment
key and `getCostLimitByUser` takes no deployment argument — so it is reported once at the top level, and a
per-entry `*CostStats.total` carries the unlimited sentinel, which is the accurate reading since only the
global budget constrains spend on any one deployment.

**Spend can be attributed.** `RateLimiter.increase` already computes the exact cost of each request; only the
storage path discarded the attribution. This adds a second, deployment-scoped record at `{deployment}/costs`
alongside the existing global one — `CostRateLimit` reused unchanged, one extra `computeResource` per *priced*
request, on a distinct key. **This is the one new write in this PR.** Enforcement keeps reading the global
document and is untouched. The figure cannot be derived after the fact: `TokenRateLimit` keeps only total
tokens, discarding the prompt/completion split the price formula needs, and pricing is hot-reloaded, so
recomputing history at today's rates would restate past spend.

#### How usage is resolved

The key set comes from config — `config.getModels().values()` filtered by `hasAccess` — so a deployment the
caller can no longer access can never be reported, and one never used is completed from config with every
`used` at zero at no storage cost. Which records to *read* is decided by a single recursive listing of the
caller's `limits/` folder, rather than by asking storage for each expected key: an installation with hundreds
of models no longer pays a lookup per model, nor a blob probe per model that has no counters at all. Listed
paths are matched against the expected paths rather than parsed, which keeps multi-segment deployment ids
working.

Order matters, and not obviously. Redis is read first and the age cutoff is applied **only** to cache misses:
`redisPut` re-scores the sync queue on every write, so a deployment used more often than `syncDelay` keeps
deferring its own flush, and the first write only creates an empty stub blob for listing. A hot record
therefore sits in Redis behind an arbitrarily stale blob object, and judging it by that timestamp would drop
the caller's *most active* deployment. A record last written before the widest window opened is left unread —
`RateWindow.MONTH` keeps 30 one-day intervals, so nothing older can project to a non-zero figure.

Every window in a response is projected against a single timestamp, so no two figures can disagree.
Membership in the usage response is decided on those projected sums, not on any timestamp, which is what makes
the set exactly "used in the last 30 days".

Storage changes are confined to `ResourceService.load`: a cutoff-aware overload that skips the blob fetch for
a stale miss, and a `blobFallbackConcurrency` bound so one bulk read cannot fan out without limit. Existing
callers keep their behaviour via the two-argument overload.

### Limitations clients must handle

Documented in the OpenAPI description, because a UI author would otherwise get them wrong:

1. **`total >= 2^53` means unlimited.** `Long.MAX_VALUE` exceeds JavaScript's `Number.MAX_SAFE_INTEGER`, so a
   browser receives it already rounded and must compare against the threshold, rendering `used` alone rather
   than a `used / total` ratio.
2. **Every window is trailing, not calendar-aligned.** `day` is the last 24 hours, `month` the last 30 days —
   never "since midnight". `used` therefore decreases on its own as activity ages out, with no reset to point
   at. Labels must read "last 24 hours", not "today".
3. **`*CostStats` means something different on the legacy endpoint.**
   `/v1/deployments/{deployment_name}/limits` returns the caller's *global* cost figures inside its
   per-deployment response. The shape is identical, so the scope must be re-read, not just the parser.
4. **Per-deployment spend does not reconcile to the global figure.** Attribution starts at rollout and does
   not back-fill, and a model without `pricing` never contributes while still consuming tokens. A UI must not
   present the breakdown as a decomposition of the global figure, nor compute "other" as the remainder.
5. **Absence from the usage response means zero, not unknown.** A client wanting the whole accessible set
   should call `/v1/user/limits` rather than reconstruct it. A deployment used and then lost access to is
   absent from both, even though its spend still counted toward the global figures.

Not to be confused with `limits` in the deployment listing, which is `TokenLimitsData` — context-window sizes,
unrelated to rate limits. The OpenAPI description says so explicitly.

### Review fixes

The later commits fix findings from an adversarial re-review; each is worth a reviewer's attention:

- **A lost update in the bulk blob fallback.** The fallback wrote the blob value back into Redis marked
  synced, which both replaces the body and removes the key from the sync queue — so racing a `computeResource`
  increment discarded the newer value *and* its pending flush. For rate-limit counters that silently
  under-counts usage, letting a caller past their limit. Latent on the `listResources` path; routing counters
  through it is what made it reachable. Now takes the lock, opt-in per call so read-only sweeps such as the
  config rebuild keep their lock-free fan-out.
- **`IllegalArgumentException` no longer maps to 401.** Three unrelated faults raise it below the controller —
  an over-long record path, a counter document that no longer parses, and unparsable blob metadata, since
  `NumberFormatException` *is* an `IllegalArgumentException`. Answering 401 to any of them signs the user out
  over a server-side problem. The caller is now resolved before dispatch, so 401 covers exactly the
  unresolvable initiator and everything else is 500 by construction.
- **The pagination marker is decoded before reuse.** `getFolderMetadata` returns it percent-encoded for HTTP
  clients while the blob provider takes it verbatim, so any key holding an escapable character resumed *past*
  the true marker — a space encodes to `%20`, which sorts after it — and the records in between were reported
  as unused.
- A name that cannot form a resource path is skipped and an unreadable record leaves its own window at zero,
  so neither takes down the report for every other deployment. Cost attribution failure is recovered
  separately, so it cannot be logged as a failure to record the spend that enforces.

### Found while implementing — not fixed here

Both are pre-existing and change what installations enforce, so they do not belong in this PR. Happy to raise
them separately.

1. **Application token limits are silently inert.** A role can configure `limits: { "my-app": { "day": 10000 } }`
   and DIAL will never enforce it, since nothing increments `{my-app}/tokens`. Either record application usage
   or reject such config at load time.
2. **Partial-window merge annihilation.** `Limit`/`CostLimit` initialise every field to `Long.MAX_VALUE` and
   the merge uses `Math.max`, so an *unspecified* window is indistinguishable from an explicit grant of
   unlimited. Role A `{day: 100}` + role B `{week: 500}` resolves to unlimited on **both** windows — granting
   the second role silently removes the first's cap. Pinned by a characterization test so this endpoint's
   output cannot shift silently if the merge is fixed later.

### Checklist

- [X] Tests pass on CI — the Linux `test / code_checks` job (`./gradlew check`, full suite plus checkstyle)
      is green, as are `build`, `style_checks` and `validate_openai_spec_generation`.

`docker_build` is red on a Trivy HIGH finding, `CVE-2026-54399` in `httpcore5-5.4.2.jar`. It is **not from this
PR**: `apache_httpclient5_version=5.4.3` is identical on `development`, the vulnerable jar is transitive, and
the same pipeline passed on this branch on 10 Aug — the CVE has been published since, and GitHub reports it
against the default branch. It needs a `httpcore5` bump to 5.4.3, which belongs in its own change.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
```diff
@@ -57,6 +62,81 @@ public Future<?> getLimits(String deploymentId) {
...
+        return respondWithUserStats(true);
+    }
+
+    private Future<?> respondWithUserStats(boolean dropEmpty) {
+        // Resolved here rather than inferred from the failure below: an unresolvable caller is the only 401,
+        // and it is the only condition this can tell apart. Deeper in, IllegalArgumentException means
+        // something else entirely - a corrupt counter document, or unparsable blob metadata - and answering
+        // 401 to those would sign the user out over a server-side data problem.
+        try {
+            BucketBuilder.buildInitiatorBucket(context);
```
astsiapanay: That never happens. Please remove the entire block

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
```diff
@@ -57,6 +62,81 @@ public Future<?> getLimits(String deploymentId) {
...
+        } catch (IllegalArgumentException e) {
+            context.respond(HttpStatus.UNAUTHORIZED, e.getMessage());
+            log.warn("LimitController. Can't resolve the request initiator", e);
+            return Future.succeededFuture();
+        }
+
+        // no executor hop: listAccessibleModels only streams over the in-memory config, and
+        // getUserStats submits the blocking part itself
+        proxy.getRateLimiter().getUserStats(context, listAccessibleModels(), dropEmpty)
```
astsiapanay: Don't inline the method call. Make it a separate variable

server/src/main/java/com/epam/aidial/core/server/data/UserLimitStats.java
```diff
@@ -0,0 +1,23 @@
...
+ * Limits and rolling usage for the deployments of the authenticated caller, keyed by deployment id.
+ *
+ * <p>The top-level cost stats are the caller's money budget and the spend against it: {@code Role.costLimit}
+ * has no deployment key and usage accumulates in a single {@code limits/costs} document, so the budget cannot
+ * be scoped to a deployment. A per-deployment {@code *CostStats} is that deployment's attributed spend against
+ * no cap, so its {@code total} carries the unlimited sentinel.
+ */
+@Data
+public class UserLimitStats {
+    private Map<String, LimitStats> deployments = new LinkedHashMap<>();
```
astsiapanay, line 24: Add javadoc what's a key and what's a value

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -36,6 +47,7 @@ public class RateLimiter {
     private static final Limit DEFAULT_LIMIT = new Limit();
     private static final CostLimit DEFAULT_COST_LIMIT = new CostLimit();
     private static final String DEFAULT_USER_ROLE = "default";
+    private static final int LIST_PAGE_SIZE = 1000;
```
astsiapanay: DEFAULT_PAGE_SIZE?

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -57,9 +69,26 @@ public Future<Void> increase(
                     usage.setAggCost(cost);
                 }
 
-                String costsPath = getPathToCosts();
-                ResourceDescriptor costResourceDescription = getResourceDescription(bucket, costsPath);
-                costFuture = taskExecutor.submit(() -> updateCostLimit(costResourceDescription, cost));
+                ResourceDescriptor costResourceDescription = getResourceDescription(bucket, getPathToCosts());
```
astsiapanay: Wrong variable name. `userCostDescriptor`

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -57,9 +69,26 @@ public Future<Void> increase(
...
+                // the global document is what enforces; the deployment-scoped one only attributes the same
+                // figure, so that a bulk report can break spend down without re-deriving it from stored
+                // tokens - which is impossible anyway, since only total tokens are kept and pricing reloads
+                ResourceDescriptor deploymentCostDescription =
```
astsiapanay: Wrong variable name

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -57,9 +69,26 @@ public Future<Void> increase(
...
+                        getResourceDescription(bucket, getPathToDeploymentCosts(roleBasedEntity.getName()));
+                // recovered, not composed: attribution is reporting-only and never back-fills anyway, so a
+                // failure there must not be reported as a failure to record the spend that enforces. Letting
+                // it fail the pair would leave an operator unable to tell the two apart in the logs
+                Future<Void> attribution = taskExecutor
+                        .submit(() -> updateCostLimit(deploymentCostDescription, cost))
+                        .recover(error -> {
+                            log.warn("Failed to attribute cost to deployment {}. The enforced total is unaffected",
```
astsiapanay: Why do we need a recovery here?

astsiapanay: It must be an error

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+            if (resourceService == null) {
+                return Future.succeededFuture();
+            }
+            return taskExecutor.submit(() -> collectUserStats(context, deployments, dropEmpty));
+        } catch (Throwable e) {
+            return Future.failedFuture(e);
+        }
+    }
+
+    private UserLimitStats collectUserStats(
```
astsiapanay: As we discussed. We have a list of deployments we want to get stats for.
We have a list of files in blob storage. We merge two lists in order to get a single one: all deployments which have non-zero usage. That could be an optimal way to building the list.

I think we could have two methods instead of one.

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
```diff
@@ -328,27 +333,55 @@ public List<Pair<ResourceItemMetadata, String>> listResources(ResourceDescriptor
...
+     *
+     * @param lockOnFallback take the resource lock around each fallback read. Required for a resource mutated
+     *                       through {@link #computeResource}: the fallback writes the blob value back into
+     *                       Redis, marking the key synced and dropping it from the sync queue, so racing a
+     *                       concurrent update would overwrite the newer value and discard its pending flush.
+     *                       Left off for read-only sweeps, where the lock would add two Redis round-trips per
+     *                       missed item to paths such as the config rebuild.
+     */
+    @SneakyThrows
+    public void load(List<ResourceItemMetadata> items, List<Pair<ResourceItemMetadata, String>> result,
```
astsiapanay: Please rollback the changes

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        do {
+            ResourceFolderMetadata page = resourceService.getFolderMetadata(folder, nextToken, LIST_PAGE_SIZE, true);
+            if (page == null) {
+                // a root folder is never reported as missing, so this only guards against a provider
+                // returning nothing mid-pagination - keep what was already collected
+                break;
+            }
+            for (MetadataBase item : page.getItems()) {
+                if (item instanceof ResourceItemMetadata metadata
+                        && wanted.contains(metadata.getDescriptor().getAbsoluteFilePath())) {
```
astsiapanay: Let's filter out blobs which are updated later than 30 days too.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+            if (limit == null) {
+                log.warn("Limit is not found for {}", name);
+                continue;
+            }
+            // DEFAULT_COST_LIMIT leaves every cost window at the unlimited sentinel: an entry reports the
+            // deployment's attributed spend, and only the global budget can cap it
+            LimitStats limitStats = create(limit, DEFAULT_COST_LIMIT);
+            Map<String, Counter> deploymentCounters = new HashMap<>();
+            try {
+                deploymentCounters.put(recordPath(bucketLocation, getPathToTokens(name)),
```
astsiapanay: Please don't inline the method calls as arguments

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        }
+    }
+
+    private UserLimitStats collectUserStats(
+            ProxyContext context, List<? extends RoleBasedEntity> deployments, boolean dropEmpty) {
+        String bucketLocation = BucketBuilder.buildInitiatorBucket(context);
+        // one instant for the age cutoff and the projection alike, so no window can disagree with another
+        long timestamp = System.currentTimeMillis();
+
+        Map<String, LimitStats> collected = new TreeMap<>();
```
astsiapanay: Let's assign meaningful name to the var

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        Map<String, Counter> counters = new HashMap<>();
```
astsiapanay: the same as the previous

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+    }
+
+    private String recordPath(String bucketLocation, String path) {
+        return getResourceDescription(bucketLocation, path).getAbsoluteFilePath();
+    }
+
+    private record Counter(LimitStats stats, RecordType type) {
+    }
+
+    private enum RecordType {
```
astsiapanay: LimitType?

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+                || stats.getDayCostStats().getUsed().signum() > 0
+                || stats.getWeekCostStats().getUsed().signum() > 0
+                || stats.getMonthCostStats().getUsed().signum() > 0;
+    }
+
+    private String recordPath(String bucketLocation, String path) {
+        return getResourceDescription(bucketLocation, path).getAbsoluteFilePath();
+    }
+
+    private record Counter(LimitStats stats, RecordType type) {
```
astsiapanay: Wrong record's name. It does't reflect the meaning of what it really does

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +156,199 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+                || stats.getMonthTokenStats().getUsed() > 0
+                || stats.getHourRequestStats().getUsed() > 0
+                || stats.getDayRequestStats().getUsed() > 0
+                || stats.getMinuteCostStats().getUsed().signum() > 0
+                || stats.getDayCostStats().getUsed().signum() > 0
+                || stats.getWeekCostStats().getUsed().signum() > 0
+                || stats.getMonthCostStats().getUsed().signum() > 0;
+    }
+
+    private String recordPath(String bucketLocation, String path) {
```
astsiapanay: `getLimitAbsolutePath`

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
astsiapanay: This method is used for bulk loads. Let's don't introduce unnecessary changes which could affect the performance.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: The listing's timestamp can't tell us that. `redisPut` re-scores the sync queue on every
write (`set.add(time() + syncDelay, key)`) and `sync()` only picks up keys that are already
due, so a record written more often than `syncDelay` (120s by default) never flushes — its
blob object keeps the timestamp of the empty stub `putResource` wrote when the record was
first created, while Redis holds the live value. `ResourceItemMetadata.getUpdatedAt()` here
therefore means "last time this deployment's traffic paused for two minutes", not "last time
it was used", so filtering on it would report zero usage for the caller's *busiest*
deployments — silently, and worst for steady-traffic project keys.

That's why the cutoff sits behind the Redis lookup rather than in front of it: a cached
record is never judged by the blob timestamp, and only a cache miss older than the widest
window (`RateWindow.MONTH`, 30 one-day intervals) is skipped. I can move it out of
`ResourceService.load` per your other comment, but it has to stay on the miss branch to be
correct.

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
oleksii-donets: Done. An unresolvable initiator now falls to the 500 handler.

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/data/UserLimitStats.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — `deploymentCostDescriptor`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — recovery removed, both writes composed. Callers already log a failed `increase` and carry on, so it surfaces without failing the response.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — `statsByDeployment`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — `targetsByRecordPath`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — `StatsTarget`: the stats a record is collected into, plus which record it is.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
oleksii-donets: Done. The semaphore, its setting, the default and the docs for it are gone, and `load(items, result)` behaves exactly as on `development`. The cutoff and the locked fallback moved to a separate `loadRecentlyUpdated`.

One line stayed: `else if (redisResult.exists())`. A Redis tombstone carries no body, so the original line NPEs on it — `testLoadOmitsMissingAndDeleted` covers it. Happy to pull it into its own PR if you'd rather.

oleksii-donets: Agreed — `load(items, result)` is back to its `development` behaviour, and the counter read now goes through its own `loadRecentlyUpdated`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: The listing's timestamp can't tell us that: `redisPut` re-scores the sync queue on every write, so a record written more often than `syncDelay` never flushes and its blob keeps the timestamp of the stub created on the first write, while Redis holds the live value. Filtering here would report zero usage for the caller's busiest deployments.

So the cutoff stays behind the Redis lookup, where it only judges a cache miss — moved out of `load` into `loadRecentlyUpdated` per your other comment.

ai-dial-actions:
<h1>Dependency Review</h1>
<tr><td><a href="https://github.com/apache/httpcomponents-core"> maven/org.apache.httpcomponents.core5:httpcore5 </a></td><td>5.4.3</td>
      <td>:green_circle: 7.2</td><td><details><summary>Details</summary><table><tr><th>Check</th><th>Score</th><th>Reason</th></tr><tr><td>Maintained</td><td>:green_circle: 10</td><td>27 commit(s) and 0 issue activity found in the last 90 days -- score normalized to 10</td></tr><tr><td>Code-Review</td><td>:green_circle: 3</td><td>Found 9/28 approved changesets -- score normalized to 3</td></tr><tr><td>Dangerous-Workflow</td><td>:green_circle: 10</td><td>no dangerous workflow patterns detected</td></tr><tr><td>Packaging</td><td>:warning: -1</td><td>packaging workflow not detected</td></tr><tr><td>Token-Permissions</td><td>:green_circle: 10</td><td>GitHub workflow tokens follow principle of least privilege</td></tr><tr><td>CII-Best-Practices</td><td>:warning: 0</td><td>no effort to earn an OpenSSF best practices badge detected</td></tr><tr><td>Security-Policy</td><td>:green_circle: 10</td><td>security policy file detected</td></tr><tr><td>Binary-Artifacts</td><td>:green_circle: 10</td><td>no binaries found in the repo</td></tr><tr><td>Signed-Releases</td><td>:warning: -1</td><td>no releases found</td></tr><tr><td>License</td><td>:green_circle: 10</td><td>license file detected</td></tr><tr><td>Branch-Protection</td><td>:warning: 0</td><td>branch protection not enabled on development/release branches</td></tr><tr><td>Pinned-Dependencies</td><td>:warning: 0</td><td>dependency not pinned by hash detected -- score normalized to 0</td></tr><tr><td>Fuzzing</td><td>:green_circle: 10</td><td>project is fuzzed</td></tr><tr><td>SAST</td><td>:green_circle: 10</td><td>SAST tool is run on all commits</td></tr></table></details></td></tr>
<tr><td><a href="https://github.com/apache/httpcomponents-core"> maven/org.apache.httpcomponents.core5:httpcore5-h2 </a></td><td>5.4.3</td>
      <td>:green_circle: 7.2</td><td><details><summary>Details</summary><table><tr><th>Check</th><th>Score</th><th>Reason</th></tr><tr><td>Maintained</td><td>:green_circle: 10</td><td>27 commit(s) and 0 issue activity found in the last 90 days -- score normalized to 10</td></tr><tr><td>Code-Review</td><td>:green_circle: 3</td><td>Found 9/28 approved changesets -- score normalized to 3</td></tr><tr><td>Dangerous-Workflow</td><td>:green_circle: 10</td><td>no dangerous workflow patterns detected</td></tr><tr><td>Packaging</td><td>:warning: -1</td><td>packaging workflow not detected</td></tr><tr><td>Token-Permissions</td><td>:green_circle: 10</td><td>GitHub workflow tokens follow principle of least privilege</td></tr><tr><td>CII-Best-Practices</td><td>:warning: 0</td><td>no effort to earn an OpenSSF best practices badge detected</td></tr><tr><td>Security-Policy</td><td>:green_circle: 10</td><td>security policy file detected</td></tr><tr><td>Binary-Artifacts</td><td>:green_circle: 10</td><td>no binaries found in the repo</td></tr><tr><td>Signed-Releases</td><td>:warning: -1</td><td>no releases found</td></tr><tr><td>License</td><td>:green_circle: 10</td><td>license file detected</td></tr><tr><td>Branch-Protection</td><td>:warning: 0</td><td>branch protection not enabled on development/release branches</td></tr><tr><td>Pinned-Dependencies</td><td>:warning: 0</td><td>dependency not pinned by hash detected -- score normalized to 0</td></tr><tr><td>Fuzzing</td><td>:green_circle: 10</td><td>project is fuzzed</td></tr><tr><td>SAST</td><td>:green_circle: 10</td><td>SAST tool is run on all commits</td></tr></table></details></td></tr>
<ul><li>settings.gradle</li></ul>

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
```diff
@@ -328,8 +328,34 @@ public List<Pair<ResourceItemMetadata, String>> listResources(ResourceDescriptor
...
+     *
+     * <p>A cache miss whose {@link ResourceItemMetadata#getUpdatedAt()} predates {@code updatedAfter} is
+     * skipped instead of being fetched, so a caller reading rolling counters does not pay for bodies that
+     * cannot hold anything within its window. The timestamp comes from the listing that produced
+     * {@code items}, so this costs no extra request. It is meaningful for a miss only: a resource written
+     * more often than {@link Settings#syncDelay} keeps deferring its own flush, so it stays in Redis behind
+     * an arbitrarily stale blob object - which is why Redis is consulted first and the cutoff never judges a
+     * cached resource.
+     */
+    public void loadRecentlyUpdated(List<ResourceItemMetadata> items,
```
astsiapanay: Let's remove the method. That's ok to have a stale data on reading stats

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
```diff
@@ -384,27 +411,33 @@ private List<Pair<ResourceItemMetadata, String>> runReadBatch(int start, int end
...
                 result.add(pair);
             }
+            // else: cached as absent, so omitted - its blob object can still be listed until the delete
+            // syncs, and a tombstone carries no body to decode
         }
         log.debug("Number of missing resources in Redis cache: {}", missed.size());
         List<Future<Pair<ResourceItemMetadata, String>>> futures = new ArrayList<>();
         for (ResourceItemMetadata metadata : missed) {
+            Long updatedAt = metadata.getUpdatedAt();
+            if (updatedAt != null && updatedAt < blobCutoff) {
```
astsiapanay: Move filtering to RateLimiter

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +147,205 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        // record is a sibling of theirs, so a deployment named "costs" lands on "costs/costs" and cannot
+        // collide with it
+        CostLimit userCostLimit = getCostLimitByUser(context);
+        LimitStats costStats = create(DEFAULT_LIMIT, userCostLimit);
+        String userCostsPath = getPathToCosts();
+        String userCostsRecord = getLimitAbsolutePath(bucketLocation, userCostsPath);
+        targetsByRecordPath.put(userCostsRecord, new StatsTarget(costStats, LimitType.COSTS));
+
+        Set<String> recordPaths = targetsByRecordPath.keySet();
+        for (Pair<ResourceItemMetadata, String> loaded : readCounters(bucketLocation, recordPaths, timestamp)) {
```
astsiapanay: Don't inline the method call. Let's assign a result to a separate variable

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +147,205 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+                String requestsPath = getPathToRequests(name);
+                String costsPath = getPathToDeploymentCosts(name);
+                String tokensRecord = getLimitAbsolutePath(bucketLocation, tokensPath);
+                String requestsRecord = getLimitAbsolutePath(bucketLocation, requestsPath);
+                String costsRecord = getLimitAbsolutePath(bucketLocation, costsPath);
+                deploymentTargets.put(tokensRecord, new StatsTarget(limitStats, LimitType.TOKENS));
+                deploymentTargets.put(requestsRecord, new StatsTarget(limitStats, LimitType.REQUESTS));
+                deploymentTargets.put(costsRecord, new StatsTarget(limitStats, LimitType.COSTS));
+            } catch (RuntimeException e) {
+                // A name that cannot form a resource path: a brace or a quote fails URL decoding, an
```
astsiapanay: Let's remove the catch block. The exception never happens

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +147,205 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        }
+    }
+
+    private UserLimitStats collectUserStats(
+            ProxyContext context, List<? extends RoleBasedEntity> deployments, boolean dropEmpty) {
+        String bucketLocation = BucketBuilder.buildInitiatorBucket(context);
+        // one instant for the age cutoff and the projection alike, so no window can disagree with another
+        long timestamp = System.currentTimeMillis();
+
+        Map<String, LimitStats> statsByDeployment = new TreeMap<>();
```
astsiapanay: Do we really need to return the result order by name?
Is the requirement mentioned anywhere in the design?

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +147,205 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        Set<String> recordPaths = targetsByRecordPath.keySet();
+        for (Pair<ResourceItemMetadata, String> loaded : readCounters(bucketLocation, recordPaths, timestamp)) {
+            String path = loaded.getKey().getDescriptor().getAbsoluteFilePath();
+            StatsTarget target = targetsByRecordPath.get(path);
+            try {
+                target.type().collect(loaded.getValue(), target.stats(), timestamp);
+            } catch (RuntimeException e) {
+                // A document that no longer parses leaves that one window at zero rather than failing the
+                // whole report for every other deployment
+                log.warn("Ignoring unreadable limit record {}", path, e);
```
astsiapanay: I don't think it's ok to skip invalid stats item. We want to throw an exception

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +147,205 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+                if (item instanceof ResourceItemMetadata metadata
+                        && wanted.contains(metadata.getDescriptor().getAbsoluteFilePath())) {
+                    items.add(metadata);
+                }
+            }
+            // decoded before being fed back: the token is percent-encoded for HTTP clients, while the blob
+            // provider takes the marker verbatim. Re-sending it encoded resumes past the true marker for any
+            // key holding an escapable character - a space encodes to %20, which sorts after it - silently
+            // skipping the records in between
+            nextToken = UrlUtil.decodePath(page.getNextToken());
```
astsiapanay: Why do we decode the next token?

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
oleksii-donets: Done.

storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java
oleksii-donets: Done — the cutoff now filters the listing in `RateLimiter.readCounters`, and `ResourceService` is back to its `development` shape apart from the tombstone branch.

One consequence to be aware of: in front of the Redis lookup the cutoff judges every listed record, not only cache misses. A record written more often than `syncDelay` never flushes, so its blob object keeps the timestamp of the stub written at creation while Redis holds the live value — such a record now reports zero. Correct for anything quieter than the sync delay.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Not required anywhere — #1799 says nothing about ordering, and nothing in the codebase sorts a response map by key. Switched to `LinkedHashMap`, so entries come back in the order the deployments were passed (config order) instead of by name.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — the catch is gone, so the `IllegalArgumentException` from `ProxyUtil.convertToObject` propagates and the controller answers 500. The test was inverted to assert that.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Because `getFolderMetadata` is asymmetric: it hands the incoming token to the provider verbatim (`ResourceService:283` → `BlobStorage:241` `afterMarker`), but percent-encodes the outgoing one (`ResourceService:295`, for HTTP clients that get it as a query parameter). A server-side loop has no URL-decoding step, so it must decode itself.

Verified against the filesystem provider — five records, page size 2, resuming with the marker as returned vs decoded:

```
rawMarker       = .../bucket/gpt 4/requests
encodedMarker   = .../bucket/gpt%204/requests
resume(raw)     = [gpt 4/tokens, gpt 5/tokens, zebra/tokens]
resume(encoded) = [zebra/tokens]           <- two records silently dropped
resume(decoded) = [gpt 4/tokens, gpt 5/tokens, zebra/tokens]
```

`%` (0x25) sorts after a space (0x20), so the encoded marker resumes past keys that should still be listed, and those deployments report zero usage with no error. Deployment names may contain spaces — only `/ { } "` are rejected.

Note `ResourceService.listResources` (`:321` → `:308`) feeds the token back without decoding and has the same latent issue; happy to fix it in a separate PR since you asked to keep this one out of `ResourceService`.

ai-dial-actions:
<h1>Dependency Review</h1>
<tr><td><a href="https://github.com/apache/httpcomponents-core"> maven/org.apache.httpcomponents.core5:httpcore5 </a></td><td>5.4.3</td>
      <td>:green_circle: 7.2</td><td><details><summary>Details</summary><table><tr><th>Check</th><th>Score</th><th>Reason</th></tr><tr><td>Maintained</td><td>:green_circle: 10</td><td>27 commit(s) and 0 issue activity found in the last 90 days -- score normalized to 10</td></tr><tr><td>Code-Review</td><td>:green_circle: 3</td><td>Found 9/28 approved changesets -- score normalized to 3</td></tr><tr><td>Dangerous-Workflow</td><td>:green_circle: 10</td><td>no dangerous workflow patterns detected</td></tr><tr><td>Packaging</td><td>:warning: -1</td><td>packaging workflow not detected</td></tr><tr><td>Token-Permissions</td><td>:green_circle: 10</td><td>GitHub workflow tokens follow principle of least privilege</td></tr><tr><td>CII-Best-Practices</td><td>:warning: 0</td><td>no effort to earn an OpenSSF best practices badge detected</td></tr><tr><td>Security-Policy</td><td>:green_circle: 10</td><td>security policy file detected</td></tr><tr><td>Binary-Artifacts</td><td>:green_circle: 10</td><td>no binaries found in the repo</td></tr><tr><td>Signed-Releases</td><td>:warning: -1</td><td>no releases found</td></tr><tr><td>License</td><td>:green_circle: 10</td><td>license file detected</td></tr><tr><td>Branch-Protection</td><td>:warning: 0</td><td>branch protection not enabled on development/release branches</td></tr><tr><td>Pinned-Dependencies</td><td>:warning: 0</td><td>dependency not pinned by hash detected -- score normalized to 0</td></tr><tr><td>Fuzzing</td><td>:green_circle: 10</td><td>project is fuzzed</td></tr><tr><td>SAST</td><td>:green_circle: 10</td><td>SAST tool is run on all commits</td></tr></table></details></td></tr>
<tr><td><a href="https://github.com/apache/httpcomponents-core"> maven/org.apache.httpcomponents.core5:httpcore5-h2 </a></td><td>5.4.3</td>
      <td>:green_circle: 7.2</td><td><details><summary>Details</summary><table><tr><th>Check</th><th>Score</th><th>Reason</th></tr><tr><td>Maintained</td><td>:green_circle: 10</td><td>27 commit(s) and 0 issue activity found in the last 90 days -- score normalized to 10</td></tr><tr><td>Code-Review</td><td>:green_circle: 3</td><td>Found 9/28 approved changesets -- score normalized to 3</td></tr><tr><td>Dangerous-Workflow</td><td>:green_circle: 10</td><td>no dangerous workflow patterns detected</td></tr><tr><td>Packaging</td><td>:warning: -1</td><td>packaging workflow not detected</td></tr><tr><td>Token-Permissions</td><td>:green_circle: 10</td><td>GitHub workflow tokens follow principle of least privilege</td></tr><tr><td>CII-Best-Practices</td><td>:warning: 0</td><td>no effort to earn an OpenSSF best practices badge detected</td></tr><tr><td>Security-Policy</td><td>:green_circle: 10</td><td>security policy file detected</td></tr><tr><td>Binary-Artifacts</td><td>:green_circle: 10</td><td>no binaries found in the repo</td></tr><tr><td>Signed-Releases</td><td>:warning: -1</td><td>no releases found</td></tr><tr><td>License</td><td>:green_circle: 10</td><td>license file detected</td></tr><tr><td>Branch-Protection</td><td>:warning: 0</td><td>branch protection not enabled on development/release branches</td></tr><tr><td>Pinned-Dependencies</td><td>:warning: 0</td><td>dependency not pinned by hash detected -- score normalized to 0</td></tr><tr><td>Fuzzing</td><td>:green_circle: 10</td><td>project is fuzzed</td></tr><tr><td>SAST</td><td>:green_circle: 10</td><td>SAST tool is run on all commits</td></tr></table></details></td></tr>
<ul><li>settings.gradle</li></ul>

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +146,196 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+    }
+
+    /**
+     * Lists the caller's {@code limits/} folder and reads the bodies of the records that belong to the
+     * response, ignoring any other name the listing turns up. A record last written before the widest
+     * window opened is left unread: {@link RateWindow#MONTH} keeps 30 one-day intervals, so nothing older
+     * can still project to a non-zero figure. The age comes from the listing entry, so skipping one costs
+     * nothing extra.
+     */
+    private List<Pair<ResourceItemMetadata, String>> readCounters(
```
astsiapanay: The method's name is wrong. It doesn't read counters.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +146,196 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        }
+    }
+
+    private UserLimitStats collectUserStats(
+            ProxyContext context, List<? extends RoleBasedEntity> deployments, boolean dropEmpty) {
+        String bucketLocation = BucketBuilder.buildInitiatorBucket(context);
+        // one instant for the age cutoff and the projection alike, so no window can disagree with another
+        long timestamp = System.currentTimeMillis();
+
+        Map<String, LimitStats> statsByDeployment = new LinkedHashMap<>();
```
astsiapanay: We don't need the map any longer

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +146,196 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        Map<String, StatsTarget> targetsByRecordPath = new HashMap<>();
+        for (RoleBasedEntity deployment : deployments) {
+            String name = deployment.getName();
+            Limit limit = getLimitByUser(context, deployment);
+            if (limit == null) {
```
astsiapanay: limit can't be null

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
astsiapanay: getDeploymentLimits

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
```diff
@@ -57,6 +61,71 @@ public Future<?> getLimits(String deploymentId) {
...
+    @ApiOperation(
+            method = "GET",
+            path = "/v1/user/limits",
+            operationId = "getUserLimits",
+            tags = {"Limits"},
+            responses = {
+                    @ApiResponse(code = 200, description = "Success", body = @ApiSchema(implementation = UserLimitStats.class)),
+                    @ApiResponse(code = 401),
+                    @ApiResponse(code = 500),
+                    @ApiResponse(code = 503)
```
astsiapanay: I don't think the endpoint may return 503

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
```diff
@@ -57,6 +61,71 @@ public Future<?> getLimits(String deploymentId) {
...
+    @ApiOperation(
+            method = "GET",
+            path = "/v1/user/usage",
+            operationId = "getUserUsage",
+            tags = {"Limits"},
+            responses = {
+                    @ApiResponse(code = 200, description = "Success", body = @ApiSchema(implementation = UserLimitStats.class)),
+                    @ApiResponse(code = 401),
+                    @ApiResponse(code = 500),
+                    @ApiResponse(code = 503)
```
astsiapanay: the same thing

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
astsiapanay: Can you add a test to prove that works in different providers like AWS, Azure?

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — `loadLimitRecords`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done — the entries go straight into `UserLimitStats.deployments`, which is already an initialized map, so the local copy is gone.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
oleksii-donets: Done — and the code that produced it is gone too: the `stats == null` branch in `respondWithUserStats` and the `resourceService == null` guard in `getUserStats`. `ResourceService` is always constructed at startup, so the endpoints now answer 200 or 500 only. Also dropped from `open_api_core.yaml`, and the test that asserted the guard is removed.

server/src/main/java/com/epam/aidial/core/server/controller/LimitController.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: The decode is gone — the hand-rolled pagination is replaced by the existing `ResourceService.listResources`, which `DeploymentService` and `MergedConfigStore` already use, so this PR no longer touches the next token at all and there is nothing new to prove per provider.

The marker is now fed back exactly as that method has always fed it, i.e. this is already existing code and its behaviour is unchanged by this PR. If the problem turns out to reproduce on AWS or Azure, the fix belongs in `listResources` in a separate PR, where it covers all of its callers rather than this endpoint alone.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -57,9 +62,17 @@ public Future<Void> increase(
                     usage.setAggCost(cost);
                 }
 
-                String costsPath = getPathToCosts();
-                ResourceDescriptor costResourceDescription = getResourceDescription(bucket, costsPath);
-                costFuture = taskExecutor.submit(() -> updateCostLimit(costResourceDescription, cost));
+                String userCostsPath = getPathToCosts();
```
astsiapanay: Move the block to a separate method `updateCostLimits`. Moreover you don't need to run two asynchronous tasks to update both costs. You can put two methods into a single asynchronous task.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +140,167 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        String userCostsPath = getPathToCosts();
+        String userCostsRecord = getLimitAbsolutePath(bucketLocation, userCostsPath);
+        targetsByRecordPath.put(userCostsRecord, new StatsTarget(costStats, LimitType.COSTS));
+
+        Set<String> recordPaths = targetsByRecordPath.keySet();
+        List<Pair<ResourceItemMetadata, String>> records = loadLimitRecords(bucketLocation, recordPaths, timestamp);
+        for (Pair<ResourceItemMetadata, String> loaded : records) {
+            String path = loaded.getKey().getDescriptor().getAbsoluteFilePath();
+            StatsTarget target = targetsByRecordPath.get(path);
+            target.type().collect(loaded.getValue(), target.stats(), timestamp);
```
astsiapanay: I don't think we need to pass the timestamp since it's used just here.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +140,167 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+        for (RoleBasedEntity deployment : deployments) {
+            String name = deployment.getName();
+            Limit limit = getLimitByUser(context, deployment);
+            // DEFAULT_COST_LIMIT leaves every cost window at the unlimited sentinel: an entry reports the
+            // deployment's attributed spend, and only the global budget can cap it
+            LimitStats limitStats = create(limit, DEFAULT_COST_LIMIT);
+            String tokensPath = getPathToTokens(name);
+            String requestsPath = getPathToRequests(name);
+            String costsPath = getPathToDeploymentCosts(name);
+            String tokensRecord = getLimitAbsolutePath(bucketLocation, tokensPath);
```
astsiapanay: The variable name is confusing.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -127,10 +140,167 @@ private LimitStats getLimitStats(ProxyContext context, Limit limit, String name)
...
+
+        UserLimitStats userLimitStats = new UserLimitStats();
+        Map<String, LimitStats> statsByDeployment = userLimitStats.getDeployments();
+        Map<String, StatsTarget> targetsByRecordPath = new HashMap<>();
+        for (RoleBasedEntity deployment : deployments) {
+            String name = deployment.getName();
+            Limit limit = getLimitByUser(context, deployment);
+            // DEFAULT_COST_LIMIT leaves every cost window at the unlimited sentinel: an entry reports the
+            // deployment's attributed spend, and only the global budget can cap it
+            LimitStats limitStats = create(limit, DEFAULT_COST_LIMIT);
```
astsiapanay, line 169: Let's refactor the block in this way:

```
LimitStats limitStats = create(limit, DEFAULT_COST_LIMIT);
            String tokensPath = getLimitAbsolutePath(bucketLocation, getPathToTokens(name));
            String requestsPath = getLimitAbsolutePath(bucketLocation, getPathToRequests(name));
            String costsPath = getLimitAbsolutePath(bucketLocation, getPathToDeploymentCosts(name));
            targetsByRecordPath.put(tokensPath, new StatsTarget(limitStats, LimitType.TOKENS));
            targetsByRecordPath.put(requestsPath, new StatsTarget(limitStats, LimitType.REQUESTS));
            targetsByRecordPath.put(costsPath, new StatsTarget(limitStats, LimitType.COSTS));
            statsByDeployment.put(name, limitStats);
```

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
astsiapanay: What's if the target is not found?

oleksii-donets: It can't be. recordPaths handed to loadLimitRecords is targetsByRecordPath.keySet() — the same map's live key set, not a copy — and the listing filter admits an entry only when wanted.contains(metadata.getDescriptor().getAbsoluteFilePath()). ResourceService.listResources (:300) then pairs each loaded body with the very ResourceItemMetadata instance that passed that filter (load(items, result), :320), so the path the loop looks up is the string already matched against the map. load can only drop entries — a missing or tombstoned body — never introduce a path that isn't a key.

A deployment with no stored record therefore never reaches the loop at all: its entry keeps the zeros it was built with from config, which is the correct reading.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
oleksii-donets: Done.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [32137281364](https://github.com/epam/ai-dial-ci/actions/runs/32137281364)

| [deploy-review](https://chat-ai-dial-core-pr-1808.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32137281364/job/95711455801) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32137281364/job/95715209944) |

PR #1809, opened 2026-08-10
fix: accept MIME types with parameters in MCP resource endpoint

karatkep:
## Summary

- `McpResourceController` was rejecting `text/html;profile=mcp-app` returned by the [MCP SDK](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/app.ts) (`RESOURCE_MIME_TYPE = "text/html;profile=mcp-app"`) because the MIME allowlist check used exact string matching.
- Strip MIME parameters (everything after `;`) before checking the allowlist. Only the base type is written into the `Content-Type` response header — upstream-controlled parameters never reach HTTP headers, preventing header injection.

## Test plan

- [x] `sendResourceResponse_mcpAppMimeType_stripsParamsAndReturnsHtml` — `text/html;profile=mcp-app` is accepted, `Content-Type: text/html` is set
- [x] `sendResourceResponse_disallowedMimeTypeWithParams_returns502` — `application/octet-stream;foo=bar` is still rejected
- [x] `sendResourceResponse_mimeTypeWithCrlf_returns502` — existing CRLF injection test still passes (no `;` in payload, so base type is the full string, which is not in the allowlist)
- [x] All existing `McpResourceControllerTest` tests pass


ai-dial-actions:
<h1>Dependency Review</h1>
None

DmytroZaichenkoDev (commented):
Reviewed the MIME-parameter stripping. The header-injection guard holds: any CR/LF either lands in the stripped parameter section (never emitted) or stays in the base type, which then fails the allowlist -> 502. Three notes inline.

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
```diff
@@ -221,7 +225,7 @@ void sendResourceResponse(McpSchema.ReadResourceResult result) {
         }
         String text = textContents.text();
         context.getResponse()
-                .putHeader(HttpHeaders.CONTENT_TYPE, mimeType)
+                .putHeader(HttpHeaders.CONTENT_TYPE, baseMimeType)
```
DmytroZaichenkoDev: `charset` is silently dropped. `text/html;charset=utf-8` is common from MCP servers and is newly *accepted* by this change, but it is now served as bare `text/html`. The body is written as UTF-8 by `end(String)`, yet the response declares no charset and the widget runs in a sandboxed null-origin iframe with no parent document to inherit encoding from, so browsers fall back to the locale default (typically windows-1252) -> mojibake for Cyrillic/CJK/emoji widget content. Suggest preserving a whitelisted `charset` param, or appending `;charset=utf-8` for `text/*` and `application/json`.

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
```diff
@@ -211,7 +211,11 @@ void sendResourceResponse(McpSchema.ReadResourceResult result) {
             context.respond(HttpStatus.BAD_GATEWAY, "Resource missing mimeType");
             return;
         }
-        if (!ALLOWED_MIME_TYPES.contains(mimeType)) {
+        // Strip parameters (e.g. ";profile=mcp-app") before allowlist check to support
+        // vendor-extended MIME types like "text/html;profile=mcp-app" from MCP SDK.
+        // Use only the base type in the response header to prevent header injection via params.
+        String baseMimeType = mimeType.contains(";") ? mimeType.substring(0, mimeType.indexOf(';')).strip() : mimeType;
```
DmytroZaichenkoDev: The allowlist lookup stays case-sensitive, but MIME types are case-insensitive (RFC 2045 5.1), so `Text/HTML;profile=mcp-app` still 502s. Since this fix exists to tolerate real-world upstream formatting, `Locale.ROOT` lowercasing of `baseMimeType` before the lookup belongs here too.

DmytroZaichenkoDev: Trimming is inconsistent: `strip()` only runs on the parameter-bearing branch, so `"text/html ;profile=x"` succeeds while `"text/html "` 502s. Stripping `mimeType` up front, before the split, makes it uniform. The `contains(";")` guard is also redundant with `indexOf(';')` -- a single `int idx = mimeType.indexOf(';')` covers both.

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
karatkep: Fixed in the latest commit. The charset parameter is now forwarded when its value is on the allowlist (utf-8, us-ascii, iso-8859-1 — covers ~100% of real-world MCP servers). Disallowed charsets are silently dropped and the base type is returned without a charset parameter. Additionally, to avoid input-size amplification from a maliciously long params string, the implementation now iterates over the small fixed allowlist (3 entries) using endsWith/contains on a semicolon-prefixed params string — no split() involved.

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
karatkep: Fixed. The full mimeType string is now lowercased upfront via `.strip().toLowerCase(Locale.ROOT)` before any splitting — both the base type and all parameter names/values are normalised in one step. `Text/HTML;profile=mcp-app` and `TEXT/HTML;CHARSET=UTF-8` are handled correctly.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
karatkep: Fixed. `strip()` and `toLowerCase(Locale.ROOT)` are now applied to the full raw mimeType before the `indexOf(';')` split, so whitespace is handled uniformly regardless of whether params are present. The redundant `contains(";") + indexOf` pair is also gone — a single `indexOf` covers both branches.

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
```diff
@@ -211,20 +218,47 @@ void sendResourceResponse(McpSchema.ReadResourceResult result) {
...
+            return Optional.empty();
+        }
+        if (semicolonIdx >= 0) {
+            // Prepend ";" so every param token has a uniform prefix — avoids split() and
+            // bounds the work to O(|ALLOWED_CHARSETS|) regardless of how many params the
+            // upstream sends (prevents ReDoS / large-input amplification via split).
+            String params = ";" + stripped.substring(semicolonIdx + 1);
+            for (String charset : ALLOWED_CHARSETS) {
+                String target = ";charset=" + charset;
+                if (params.endsWith(target) || params.contains(target + ";")) {
```
DmytroZaichenkoDev: The matching requires the param to abut `;` with no whitespace, but RFC 9110 allows OWS around `;` and the quoted form is legal too, so these all silently drop the charset:

```
text/html; charset=utf-8                   -> text/html   (most common wire form)
text/html; charset=utf-8; profile=mcp-app  -> text/html
text/html ; charset=utf-8                  -> text/html
text/html;charset="utf-8"                  -> text/html
text/html;charset=utf-8 ;profile=x         -> text/html
```

Express/FastAPI/Starlette emit the space form, so the mojibake case is still live for most real responses. The tests only cover the no-space form, which is why it slipped through.

A split-based scan is both whitespace/quote-correct and still bounded (`split(";")` on a literal delimiter is linear — no backtracking, so the ReDoS note above doesn't really apply; the 3-iteration `contains` loop actually scans more than one split would):

```java
for (String param : stripped.substring(semicolonIdx + 1).split(";")) {
    String p = param.strip();
    if (p.startsWith("charset=")) {
        String cs = p.substring(8).strip().replace("\"", "");
        return Optional.of(ALLOWED_CHARSETS.contains(cs) ? baseMimeType + ";charset=" + cs : baseMimeType);
    }
}
```

First-match-wins also removes the nondeterminism when a malformed value repeats the param (`;charset=utf-8;charset=iso-8859-1`) — `Set.of` iteration order is unspecified.

DmytroZaichenkoDev (commented):
Thanks — normalisation and the charset allowlist look good, and the header value is now assembled purely from constants, so injection is structurally impossible. One gap left on the charset matching.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/McpResourceController.java
karatkep: Addressed in the latest commit: switched to `split(";")` + `strip()` per token — handles OWS around ; (RFC 9110 §8.3), enclosing double quotes (RFC 9110 §5.6.6 quoted-string), and first-match-wins for duplicate charset `params.split(";")` on a literal delimiter has no backtracking, so no DoS concern.

ai-dial-actions:
<h1>Dependency Review</h1>
None

DmytroZaichenkoDev (approved):

PR #1813, opened 2026-08-10
feat: key blob-sourced config entities by short name, schemas by $id #1783

### LINKED ISSUE #1783: Slice B+C — Short-name resolution via direct keying + schema $id index
# Slice B+C — Short-name resolution via direct keying + schema `$id` index

Part of #1781.

## Goal

Make every materialized `platform`-bucket config entity **short-name addressed**,
inbound and outbound. Add the schema-lookup mechanism that lets blob-stored schemas
resolve by their JSON-Schema `$id`.

- **Inbound**: `gpt-4` resolves to the blob entity for models/applications/toolsets/
  interceptors/roles.
- **Outbound**: listings/responses emit `gpt-4`, never a canonical path.

### Part B — models / applications / toolsets / interceptors / roles

`Config`'s maps for these five types are keyed **directly by short name**. File-sourced
and blob-sourced entries for the same entity share one map key, so a blob write
naturally overwrites the file-sourced entry via ordinary map `put` — no separate "blob
shadows file" removal step, no outbound renaming pass, no `resolve()`-style accessor
layer. Every consumer (`RateLimiter`, `ShareService`, `ModelController`,
`DeploymentController`, `DeploymentPostController`, `CollectResponseAttachmentsFn`,
`BlobEntityValidator`, `Config.selectDeployment`) does a plain `Map.get`/`containsKey`
against the short name.

**Limitation:** canonical ids (`models/platform/gpt-4`) are not a map key for these
five types, so they don't resolve inbound — only the short name does.

Two mechanisms this requires:

- **Single-entity GET reads directly from blob storage** by descriptor
  (`ConfigResourceController.handleSingleGetFromBlob`) for models/apps/toolsets/
  interceptors/roles and both schema kinds, since a blob-managed entity and a
  file-only entity can share the identical short-name key and the in-memory `Config`
  map alone can't tell which is which. Only `PROJECT_KEY`/`ROUTE` still use the
  map-based GET.
- **Cross-type deployment-id uniqueness**, since a model and an application can now
  collide on the same short name. Guarded by `ConfigPostProcessor.isDeploymentIdTaken`
  (full-rebuild / batch-apply path) and `ConfigResourceController.rejectDuplicateDeploymentId`
  (single-entity PUT path).

### Part C — schema `$id` index

`applicationTypeSchemas`/`catalogSchemas` are keyed **directly by `$id`**. The blob
address for a schema remains the admin/client-chosen path segment, independent of its
`$id`.

Schema `$id` uniqueness is enforced at two surfaces sharing one implementation, so they
can't drift apart:

- Single-entity `PUT` (`ConfigResourceController.buildSchemaEventMetadata`) — `409
  Conflict` if a different blob already claims the `$id`.
- Batch `/v1/admin/apply` (real-apply and `precheck` dry-run) and
  `/v1/admin/validate` — all route through the same static helpers in
  `AdminApplyController` (`schemaValidationError`/`schemaConflictMessage`/
  `readOldSchemaId`), with `AdminValidateController` delegating to them directly, so
  apply and validate can never disagree on a given batch.

## Depends on

Slice A for the applications/toolsets portion (models/interceptors/roles and the
schema `$id` handling are independent of A).

KirylKurnosenka:
Makes every materialized `platform`-bucket config entity short-name addressed, inbound and outbound, by keying `Config`'s maps for models/applications/toolsets/interceptors/roles directly by short name, and keys the app-type/catalog schema maps directly by their JSON-Schema `$id` so blob-stored schemas resolve by `$id` with no separate alias index.

### Applicable issues

- fixes #1783

### Description of changes

- `Config.java`: `models`/`applications`/`toolsets`/`interceptors`/`roles` maps are keyed directly by short name (no canonical-id keys, no `resolve()` derivation layer, no accessor methods); `selectDeployment` is a plain sequential `Map.get`. `applicationTypeSchemas`/`catalogSchemas` are keyed directly by `$id` (no alias index).
- `ConfigPostProcessor.java`: `entity.setName(mapKey)` directly, since the map key already is the short name; added the cross-type deployment-id uniqueness check (`isDeploymentIdTaken`) needed now that a model and an application can collide on the same short name.
- `MergedConfigStore.java`: blob entries for the five short-name-keyed types are inserted at the short-name key, so a blob write naturally supersedes the file entry via ordinary `Map.put`; no separate shadow-removal step. Schema entries are inserted keyed by `$id`.
- `ConfigResourceController.java`: single-entity GET for models/apps/toolsets/interceptors/roles/both schema kinds now reads directly from blob storage by descriptor (`handleSingleGetFromBlob`) instead of the `Config` map, since the map alone can no longer distinguish a blob-managed entity from a file-only one sharing the same key; added the single-entity-PUT counterpart of the deployment-id and schema-`$id` uniqueness checks (`rejectDuplicateDeploymentId`, `buildSchemaEventMetadata`, `409 Conflict` on collision).
- `AdminApplyController.java`: extracted `schemaValidationError`/`schemaConflictMessage`/`readOldSchemaId` so `applySchema` (real-apply) and `validateOnly` (apply precheck / dry-run) share one schema-`$id`-uniqueness implementation; `AdminValidateController` delegates to the same static helpers so validate and apply can never disagree on a batch.
- Call sites (`RateLimiter`, `ShareService`, `ModelController`, `DeploymentController`, `DeploymentPostController`, `CollectResponseAttachmentsFn`, `BlobEntityValidator`) do plain direct map access against the short-name/`$id`-keyed maps.
- Tests: `ConfigTest`/`ConfigPostProcessorTest`/`MergedConfigStoreTest` updated for direct-keying behavior and the new deployment-id/`$id` uniqueness checks; `AdminApplyApiTest`/`AdminValidateApiTest` extended with apply/validate parity coverage for schema `$id` collisions (precheck=true and precheck=false).

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/Config.java
```diff
@@ -56,43 +56,97 @@ public class Config {
 
     private List<String> globalInterceptors = List.of();
 
+    /**
+     * $id → canonical-id index for {@code platform}-bucket schema entities, built at rebuild
+     * time from blob bodies. Bridges $id-keyed file entries and canonical-id-keyed blob entries
+     * in {@link #applicationTypeSchemas}, since a schema's $id is not derivable from its path.
+     */
+    @JsonIgnore
+    private Map<String, String> schemaAliasesById = Map.of();
```
astsiapanay: applicationSchemaAliasesById

config/src/main/java/com/epam/aidial/core/config/Config.java
```diff
@@ -56,43 +56,97 @@ public class Config {
...
+        return resolveSchema(catalogSchemas, catalogSchemaAliasesById, schemaId);
+    }
+
+    /**
+     * Resolves a schema by its $id: verbatim lookup first (canonical-id callers, and file entries
+     * already keyed by $id), then falls back through the $id → canonical-id alias index for a
+     * migrated blob entry. A schema's $id is not derivable from its path, so unlike {@link
+     * #resolve}, the alias index must be maintained explicitly (see {@code MergedConfigStore}).
+     */
+    private static String resolveSchema(Map<String, String> schemas, Map<String, String> aliasesById, URI schemaId) {
```
astsiapanay: @Nullable

config/src/main/java/com/epam/aidial/core/config/Config.java
```diff
@@ -56,43 +56,97 @@ public class Config {
 
     private List<String> globalInterceptors = List.of();
 
+    /**
+     * $id → canonical-id index for {@code platform}-bucket schema entities, built at rebuild
```
astsiapanay: The javadoc is not clear what key is and what value is.

server/src/main/java/com/epam/aidial/core/server/service/ShareService.java
```diff
@@ -292,11 +293,11 @@ private void updateLimits(ProxyContext context, ResourceType resourceType, Share
 
     private ShareResourceLimit getLimit(ProxyContext context, ResourceType resourceType) {
         List<String> userRoles = context.getUserRoles();
-        Map<String, Role> roles = context.getConfig().getRoles();
+        Config config = context.getConfig();
         ShareResourceLimit defaultLimit = DEFAULT_LIMITS.get(resourceType);
         ShareResourceLimit limit = null;
         for (String userRole : userRoles) {
-            ShareResourceLimit candidate = Optional.ofNullable(roles.get(userRole)).map(Role::getShare).map(limits -> limits.get(resourceType.name())).orElse(null);
+            ShareResourceLimit candidate = Optional.ofNullable(config.getRole(userRole)).map(Role::getShare).map(limits -> limits.get(resourceType.name())).orElse(null);
```
astsiapanay: why do we need these changes?

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -949,28 +961,105 @@ private static Object peekEntity(Config config, ResourceTypes type, String canon
...
+        schemas.put(canonicalId, body);
+        try {
+            recordSchemaAlias(schemas, aliasesById, canonicalId, ProxyUtil.BLOB_MAPPER.readTree(body));
+        } catch (JsonProcessingException e) {
+            log.warn("Failed to parse schema body for $id alias index: {} ({})", canonicalId, e.getMessage());
+        }
+    }
+
+    /**
+     * Records the {@code $id → canonicalId} alias read from a schema's body, and removes the
```
astsiapanay: Who will read a such long javadoc?
The method has 5 lines the doc is more than 10 long lines?

server/src/main/java/com/epam/aidial/core/server/service/ShareService.java
KirylKurnosenka: Blob-stored `Role` entities are now short-name addressed via canonical id (e.g. `roles/platform/<name>`), same as models/applications/toolsets/interceptors in this PR. A raw `roles.get(userRole)` map lookup would miss a role migrated to the blob store, since its map key is the canonical id, not the short name. `Config.getRole(id)` goes through the new `resolve()` helper (short-name-first, canonical-id fallback), so this swap is part of the same call-site sweep as the other accessors — not unrelated churn.

ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/Config.java
KirylKurnosenka: Renamed to `applicationSchemaAliasesById` (field/getter/setter, and all call sites).

config/src/main/java/com/epam/aidial/core/config/Config.java
KirylKurnosenka: `javax.annotation.Nullable` isn't on the `config` module's classpath (only pulled in transitively by `server`), and adding a direct dependency just for this annotation felt disproportionate. Documented nullability via `@return ... or {@code null} if ...` javadoc on `getCustomApplicationSchema`/`getCatalogSchema`/`resolveSchema` instead.

config/src/main/java/com/epam/aidial/core/config/Config.java
KirylKurnosenka: Reworded to explicitly state key = schema $id, value = canonical id, and added a concrete example.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Trimmed both this method's javadoc and removeEntityInPlace's down to their essential contract.

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -416,6 +441,24 @@ private static boolean skipOnDuplicate(String name, ResourceTypes type, Set<Stri
         return true;
     }
 
+    public static boolean isDeploymentIdTaken(Config config, ResourceTypes type, String shortName) {
```
astsiapanay: What does the method do?

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -136,31 +138,54 @@ static void validateSingleModel(Config config, String canonicalId,
...
+            role.setName(mapKey);
+        }
+    }
+
+    /**
+     * Targeted per-type helper for {@link MergedConfigStore} partial-update path. Sets
+     * {@code application.name} from the map key. No cross-ref validation — applications have
+     * no outbound refs checked here.
+     */
+    static void validateSingleApplication(Config config, String mapKey) {
```
astsiapanay: The method name doesn't reflect the logic

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -391,7 +397,30 @@ void applyReplicaEvent(ResourceDescriptor descriptor, ResourceEvent.Action actio
...
+
+    /**
+     * Handles a schema-type replica delete, using the {@code $id} carried in {@code eventMetadata}
+     * when available. Falls back to a full rebuild when the metadata is absent — the blob body is
+     * gone at delete time so the map key cannot be derived any other way.
+     */
+    private void applyReplicaSchemaDelete(ResourceTypes type, @Nullable Map<String, String> eventMetadata) {
+        String schemaId = eventMetadata != null ? eventMetadata.get("$id") : null;
+        if (schemaId == null) {
+            requestRebuild();
```
astsiapanay, line 423: Why do we rebuilt here?

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -603,10 +632,10 @@ private void cancelPendingRebuild() {
      * resurrected when the new interceptor satisfies their references. For other types no transitive
      * effect runs.
      */
-    public Config applyEntityWrite(ResourceTypes type, String canonicalId, Object decryptedEntity) {
+    public Config applyEntityWrite(ResourceTypes type, String mapKey, Object decryptedEntity) {
```
astsiapanay, line 641: Why do you replace canonical ID with mapKey?

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -1237,56 +1295,130 @@ public static String canonicalId(ResourceDescriptor descriptor) {
...
+    /**
+     * The Config map key for non-schema types.
+     * <ul>
+     *   <li>Models/interceptors/roles/applications/toolsets → short name (last path segment)</li>
+     *   <li>Every other non-schema type → canonical id</li>
+     * </ul>
+     * Throws {@link IllegalArgumentException} if called for a schema type — use
+     * {@link #mapKeyForSchema(JsonNode)} for {@code APP_TYPE_SCHEMA} / {@code CATALOG_SCHEMA}.
+     */
+    public static String mapKeyFor(ResourceDescriptor descriptor) {
```
astsiapanay: What's a map key? Why do we need the method at all?

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
```diff
@@ -347,6 +380,82 @@ static ValidationResult validateOnly(AdminManifest entry, Config scratch, boolea
         return new ValidationResult(id, ValidationStatus.VALID, null);
     }
 
+    /**
+     * Shared by {@link #validateOnly} (precheck) and the real-apply {@code applyX} methods:
+     * non-null iff {@code parsed}'s short name is already claimed by a different
+     * model/application/interceptor/toolset in {@code scratch}. See
+     * {@link ConfigPostProcessor#isDeploymentIdTaken}.
+     */
+    private static String duplicateDeploymentIdMessage(Config scratch, ResourceTypes type, ParsedName parsed) {
```
astsiapanay: The method name `duplicateDeploymentIdMessage` is weird the same as logic

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
```diff
@@ -347,6 +380,82 @@ static ValidationResult validateOnly(AdminManifest entry, Config scratch, boolea
...
+        }
+        return null;
+    }
+
+    /**
+     * Shared by {@code validateOnly} (precheck) and {@code applySchema} (real-apply): validates a
+     * Schema/CatalogSchema spec and, if it's well-formed, checks its {@code $id} for an in-place
+     * change or a collision against a different blob via {@link #schemaConflictMessage}.
+     */
+    private static String schemaValidationError(AdminManifest entry, ParsedName parsed, Config scratch,
```
astsiapanay: Why the method name is a noun?

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
astsiapanay: hasDeploymentDuplicate?

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
astsiapanay: please add a warning message

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
astsiapanay: Perhaps it's wort documenting what a mapKey is to class description.

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
KirylKurnosenka: Renamed

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
KirylKurnosenka: Renamed

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Added comment regarding `mapKey` definition to the top of the class. The method is needed to resolve `mapKey` for different resourceTypes.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Since now we starting to store different entities under different key in in-memory config: the short name for models/interceptors/roles/applications/toolsets, the `$id` for schema types, and the canonical id for all other types (keys, routes)

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Since there are no other ways (except event metadata) to identify schema $id which should be removed from in-memory config, when schema $id is missing in event metadata we need to request rebuild to restore correct state. Added `log.warn` to log such situation

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
KirylKurnosenka: Renamed, added javadoc and refactored

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
KirylKurnosenka: Refactored

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
```diff
@@ -550,22 +699,24 @@ static void mutateScratch(Config scratch, AdminManifest entry) {
                     scratch.getToolsets().put(entry.name(), toolSet);
```
astsiapanay: **Correctness (CONFIRMED):** `mutateScratch` keys newly-applied batch entries by `entry.name()` (the raw manifest resource name, e.g. `interceptors/platform/myinterceptor`) instead of the short name. Since this PR moved Model/Interceptor/Role/Application/ToolSet maps to short-name keying everywhere else, this call site was missed (only Schema/CatalogSchema branches key by `$id`).

Failure scenario: a batch defines an Interceptor `myinterceptor` followed by a Model referencing `interceptors: ["myinterceptor"]`. `mutateScratch` stores the interceptor under the wrong key, so `ConfigPostProcessor.validateCrossReferences`'s `scratch.getInterceptors().containsKey("myinterceptor")` returns false and the Model is spuriously rejected — even though the interceptor is defined earlier in the same batch. The same mis-keying also defeats `isDeploymentIdTakenByAnotherDeploymentType`'s same-batch duplicate-id detection.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
```diff
@@ -1530,16 +1647,32 @@ private Future<?> handleDelete() {
...
                 throw new HttpException(HttpStatus.NOT_FOUND,
                         "Resource not found: " + descriptor.getUrl());
             }
             if (deletedSecret != null) {
                 apiKeyStore.removeKey(deletedSecret);
             }
-            mergedConfigStore.applyEntityDelete(typeOf(descriptor), MergedConfigStore.canonicalId(descriptor));
+            String deleteMapKey = deletedSchemaId != null
+                    ? deletedSchemaId
+                    : MergedConfigStore.resolveMapKeyFor(descriptor);
```
astsiapanay: **Correctness (CONFIRMED):** the schema DELETE handler falls back to `MergedConfigStore.resolveMapKeyFor(descriptor)` when `$id` extraction from the pre-delete blob fails — but `resolveMapKeyFor` explicitly throws `IllegalArgumentException` for `APP_TYPE_SCHEMA`/`CATALOG_SCHEMA` ("Use mapKeyForSchema(JsonNode) for schema types"), contradicting the comment above it that says "applyEntityDelete will be skipped."

Failure scenario: `DELETE /v1/admin/config/schemas/platform/{name}` targets a resource whose stored JSON is malformed/unparseable. `extractSchemaId` throws, `deletedSchemaId` stays `null`. `resourceService.deleteResource(...)` already ran and succeeded — the blob is physically gone. Then `resolveMapKeyFor(descriptor)` throws, propagating to `handleWriteError` and returning a 500 to the client even though the delete succeeded, while `applyEntityDelete` never runs, leaving a stale entry in the in-memory `Config` map until the next rebuild.

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
```diff
@@ -347,6 +380,82 @@ static ValidationResult validateOnly(AdminManifest entry, Config scratch, boolea
...
+        if (schemaId == null || schemaId.isBlank()) {
+            return kindLabel + " spec must contain a non-blank $id field";
+        }
+        ResourceDescriptor descriptor = ResourceDescriptorFactory.fromDecoded(
+                type, parsed.bucket(), parsed.location(), parsed.name());
+        String oldSchemaId = readOldSchemaId(resourceService, descriptor);
+        return validateSchemaId(scratch, type, schemaId, oldSchemaId);
+    }
+
+    private static String readOldSchemaId(ResourceService resourceService, ResourceDescriptor descriptor) {
```
astsiapanay, line 423: **Correctness (CONFIRMED):** `readOldSchemaId` swallows *all* exceptions (`catch (Exception ignored)`), including transient storage failures from `resourceService.getResource` — not just a "not found" case — and returns `null`. The caller then treats `null` as "this is a create," silently skipping the `$id`-immutability check.

Failure scenario: an apply/validate batch updates an existing schema (current `$id` "urn:a") with a new spec whose `$id` is "urn:b". If `resourceService.getResource` throws due to a transient Redis/blob error, this method swallows it and returns `null`, indistinguishable from "no prior schema." `validateSchemaId` then treats it as a create and skips the immutability check, silently accepting the `$id` rename. Note `ConfigResourceController.buildSchemaEventMetadata` handles the equivalent case by failing *closed* instead — the two write paths (bulk apply vs. single-entity PUT) can now disagree on the same request.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -1302,25 +1440,18 @@ private static void warnIfReplaced(ResourceTypes type, String canonicalId, Objec
         }
     }
 
-    private static void removeAddedEntity(ResourceTypes type, String canonicalId,
-                                          Map<String, Model> models, Map<String, Interceptor> interceptors,
-                                          Map<String, Role> roles, Map<String, Key> keys,
-                                          LinkedHashMap<String, Route> routes, Map<String, String> schemas,
-                                          Map<String, String> catalogSchemas,
-                                          Map<String, Application> applications, Map<String, ToolSet> toolsets) {
+    private static void removeAddedEntity(Config config, ResourceTypes type, String mapKey) {
```
astsiapanay: **Correctness (CONFIRMED):** `removeAddedEntity` unconditionally does `config.getModels().remove(mapKey)` (etc.) on decrypt-failure rollback, rather than restoring whatever entity previously occupied that key. Since blob and file entries for these five types now share one short-name key, the `put()` in `addBlobEntity` can shadow a valid file-sourced entry — and the `put()`'s return value (the prior entry) is only passed to `warnIfReplaced` for logging, never saved for rollback.

Failure scenario: a file-configured model `gpt-4` coexists with a blob-sourced model also named `gpt-4` whose secret field fails KMS decryption during rebuild. `addBlobEntity`'s `put` overwrites the file entry; `decryptManagedEntity` then throws; `removeAddedEntity` deletes the `gpt-4` key entirely instead of restoring the original valid file-sourced model — so a working file-defined model disappears from the running config until the next full rebuild.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
```diff
@@ -1237,56 +1301,130 @@ public static String canonicalId(ResourceDescriptor descriptor) {
...
+    }
+
+    /**
+     * Resolves the Config map key for a blob entry during a full rebuild. For schema types, returns
+     * the {@code $id} extracted from the blob body, or {@code null} (and logs a warning) when the
+     * body has no textual {@code $id}. For all other types, delegates to
+     * {@link #resolveMapKeyFor(ResourceDescriptor)}.
+     */
+    @Nullable
+    private static String resolveRebuildMapKey(ResourceDescriptor descriptor, JsonNode node, String canonicalId) {
```
astsiapanay: **Correctness (CONFIRMED):** `resolveRebuildMapKey` silently drops schema blobs missing a textual `$id` via a bare `continue` in the rebuild loop, without calling `recordInvalid` — unlike every other failure branch in the same loop (parse failure, decrypt failure), which does record the entity as invalid.

Failure scenario: a blob-stored `APP_TYPE_SCHEMA`/`CATALOG_SCHEMA` lacking a string `$id` (e.g. pre-migration data) is silently excluded from `config.getApplicationTypeSchemas()`/`getCatalogSchemas()` on every rebuild, with only a WARN log line. `AdminHealthConfigController`'s `/v1/admin/health/config` endpoint builds its `status`/`skipped` response entirely from `getInvalidEntities()`, so it reports `"ok"` even though a schema silently failed to load — hiding the problem from operators in a way that a parse or decrypt failure on the same blob would not.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
```diff
@@ -1473,18 +1544,60 @@ private Future<?> handlePut() {
...
             }));
         }).onSuccess(meta -> context.putHeader(HttpHeaders.ETAG, meta.getEtag())
-                .respond(HttpStatus.OK, createNameEnvelope(name)))
+                .respond(HttpStatus.OK, createNameEnvelope(path)))
                 .onFailure(this::handleWriteError);
 
         return Future.succeededFuture();
     }
 
+    private Map<String, String> buildSchemaEventMetadata(ResourceTypes schemaType, JsonNode requestNode,
```
astsiapanay, line 1562: **Reuse:** the schema `$id` immutability/uniqueness rule is independently reimplemented here (`buildSchemaEventMetadata`) instead of sharing `AdminApplyController.validateSchemaId`/`readOldSchemaId`. Both implement the same two rules (immutability + near-identical error messages) but with different exception-handling semantics — this one fails closed on a read/parse error, `AdminApplyController.readOldSchemaId` fails open (see the comment on that method).

This divergence is exactly why the bulk-apply and single-entity-PUT paths can already disagree on the same request (see the `readOldSchemaId` finding). Extracting one shared static helper (e.g. `(Config scratch, ResourceTypes schemaType, String newSchemaId, String oldSchemaId) -> Optional<String>`) and calling it from both controllers would prevent future changes to the rule from silently reintroducing this kind of drift.

docs/open_api_core.yaml
```diff
@@ -16249,6 +16249,8 @@ components:
           type: integer
         url:
           type: string
+        metadata:
```
astsiapanay, line 16252: **Conventions:** `docs/CODE_STYLE.md` states: "`docs/open_api_core.yaml` is generated and auto-committed by CI (`./gradlew replaceSpec` in `.github/workflows/pr.yml`). Never hand-edit it — change the controller annotations."

This diff hand-adds a `metadata: $ref: "#/components/schemas/MapStringString"` entry to the `ResourceEvent` schema here, matching the new `metadata` field added to `ResourceEvent.java`. When `replaceSpec` next regenerates the spec, it may produce a differently-shaped entry (field order, nullability, description) and diverge from this hand-written stanza, causing spurious diff noise or silent drift in future PRs. Consider running `./gradlew replaceSpec` instead of hand-editing this file.

astsiapanay:
**Correctness (CONFIRMED)** — not attributable to a single diff line since `RateLimiter.java` itself is untouched by this PR, but flagging as a real regression it introduces:

`RateLimiter.getLimit()`/`getLimitByUser()` (`server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java`, ~line 578) look up `role.getLimits().get(name)` where `name = roleBasedEntity.getName()`. This PR changes what `getName()` returns for API-managed (blob-sourced) models/interceptors/roles/applications/toolsets — from the canonical id (e.g. `"models/platform/gpt-4"`) to the short name (e.g. `"gpt-4"`). The prior behavior was an explicit, named contract (the removed "OQ-23" comment in `MergedConfigStore.java` stated canonical IDs surface in "rate-limit role-limit lookups for API-managed deployments"), and a test (`MergedConfigStoreApiTest.testBlobModelSurfacesAfterReload`) previously asserted the canonical-id form; it's now rewritten to assert the short name instead.

Failure scenario: an operator configured `Role.limits` keyed by the old canonical id for a blob-managed model. After this PR, the lookup key is the short name, so the configured limit silently stops matching and the deployment falls back to `DEFAULT_LIMIT` with no error or warning (`Role.limits` cross-refs are warning-only and no validation was added for this migration). Worth confirming whether this transition is documented/handled for existing configs.

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
KirylKurnosenka: Fixed — `mutateScratch` now keys via `parseName(entry).name()` (the short name) for Interceptor/Role/Model, matching the short-name-keyed maps everywhere else. Also gated the Application/ToolSet branches on the platform bucket, since public-bucket apps/toolsets aren't materialized into `MergedConfigStore` and shouldn't be pushed into scratch. Pushed in 87d1434b.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
KirylKurnosenka: Fixed — the DELETE handler now branches explicitly on `isSchema`: if `deletedSchemaId` is null (blob missing/corrupt), we skip the in-memory `applyEntityDelete` call entirely instead of falling back to the throwing `resolveMapKeyFor`. The blob delete has already succeeded at that point; any stale in-memory entry is cleaned up on the next full rebuild. Non-schema types still resolve their map key via `resolveMapKeyFor` as before. Pushed in 87d1434b.

server/src/main/java/com/epam/aidial/core/server/controller/AdminApplyController.java
KirylKurnosenka: Fixed — `readOldSchemaId` now only treats a genuinely missing resource (`getResource` returning null) as "no prior $id". Any exception from the read or the JSON parse now throws `HttpException(500)` instead of being swallowed, so it fails closed like `ConfigResourceController.buildSchemaEventMetadata` does — the two write paths no longer disagree on this. Pushed in 87d1434b.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Fixed — `addBlobEntity` now returns a `BlobPutResult(added, previous)` capturing whatever the `put()` shadowed, and `removeAddedEntity` restores that `previous` value (or removes the key if there was none) instead of unconditionally deleting on rollback. A file-sourced entry shadowed by a blob entry that then fails decryption is now correctly restored rather than lost. Pushed in 87d1434b.

server/src/main/java/com/epam/aidial/core/server/config/MergedConfigStore.java
KirylKurnosenka: Fixed — `resolveRebuildMapKey` now calls `recordInvalid` itself for the schema-$id-missing case (the only way it can return null), so the entity shows up in `getInvalidEntities()`/the health-config endpoint like every other skip reason, instead of only a WARN log line. Moved the `recordInvalid` call out of the `rebuild()` loop and into the resolver so the failure reason is tied to its actual cause rather than assumed by the caller. Pushed in 87d1434b.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
KirylKurnosenka: Extracted into `MergedConfigStore.validateSchemaId(schemaMap, isUpdate, newSchemaId, oldSchemaId)` (+ `getSchemaMapOf` for the map lookup), shared by both `AdminApplyController` and `ConfigResourceController.buildSchemaEventMetadata` now. Combined with the `readOldSchemaId` fail-closed fix above, both write paths now apply the identical immutability/uniqueness rule. Pushed in 87d1434b.

docs/open_api_core.yaml
KirylKurnosenka: Updated by `./gradlew replaceSpec` in `.github/workflows/pr.yml`

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA 9f3944a5d39a4dcf989a5ff00a193b6539d94371.</blockquote>
None

KirylKurnosenka:
> **Correctness (CONFIRMED)** — not attributable to a single diff line since `RateLimiter.java` itself is untouched by this PR, but flagging as a real regression it introduces:
> 
> `RateLimiter.getLimit()`/`getLimitByUser()` (`server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java`, ~line 578) look up `role.getLimits().get(name)` where `name = roleBasedEntity.getName()`. This PR changes what `getName()` returns for API-managed (blob-sourced) models/interceptors/roles/applications/toolsets — from the canonical id (e.g. `"models/platform/gpt-4"`) to the short name (e.g. `"gpt-4"`). The prior behavior was an explicit, named contract (the removed "OQ-23" comment in `MergedConfigStore.java` stated canonical IDs surface in "rate-limit role-limit lookups for API-managed deployments"), and a test (`MergedConfigStoreApiTest.testBlobModelSurfacesAfterReload`) previously asserted the canonical-id form; it's now rewritten to assert the short name instead.
> 
> Failure scenario: an operator configured `Role.limits` keyed by the old canonical id for a blob-managed model. After this PR, the lookup key is the short name, so the configured limit silently stops matching and the deployment falls back to `DEFAULT_LIMIT` with no error or warning (`Role.limits` cross-refs are warning-only and no validation was added for this migration). Worth confirming whether this transition is documented/handled for existing configs.



> **Correctness (CONFIRMED)** — not attributable to a single diff line since `RateLimiter.java` itself is untouched by this PR, but flagging as a real regression it introduces:
> 
> `RateLimiter.getLimit()`/`getLimitByUser()` (`server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java`, ~line 578) look up `role.getLimits().get(name)` where `name = roleBasedEntity.getName()`. This PR changes what `getName()` returns for API-managed (blob-sourced) models/interceptors/roles/applications/toolsets — from the canonical id (e.g. `"models/platform/gpt-4"`) to the short name (e.g. `"gpt-4"`). The prior behavior was an explicit, named contract (the removed "OQ-23" comment in `MergedConfigStore.java` stated canonical IDs surface in "rate-limit role-limit lookups for API-managed deployments"), and a test (`MergedConfigStoreApiTest.testBlobModelSurfacesAfterReload`) previously asserted the canonical-id form; it's now rewritten to assert the short name instead.
> 
> Failure scenario: an operator configured `Role.limits` keyed by the old canonical id for a blob-managed model. After this PR, the lookup key is the short name, so the configured limit silently stops matching and the deployment falls back to `DEFAULT_LIMIT` with no error or warning (`Role.limits` cross-refs are warning-only and no validation was added for this migration). Worth confirming whether this transition is documented/handled for existing configs.

Since the admin management API is not yet stable, for `Role.limits` keyed by the old canonical id it's known issue. Operators should reconfigure them to use short name instead of canonical id.

KirylKurnosenka:
/deploy-review

>GitHub actions run: [32833368699](https://github.com/epam/ai-dial-ci/actions/runs/32833368699)

| [deploy-review](https://chat-ai-dial-core-pr-1813.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32833368699/job/97756773515) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/32833368699/job/97759735531) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1818, opened 2026-08-11
feat: add DIAL_NATIVE auth type and per-provider offline client config

DmytroZaichenkoDev:
Part 1/4 of the offline-delegation split (replaces #1815).

Introduces the `DIAL_NATIVE` authentication type for external services that target DIAL itself: the declaration carries no credential material (its validator rejects every OAuth/API-key field, records never expire or refresh), and each identity provider may configure an `offlineClient` used later to obtain a user's offline credentials. The type is declarable but inert — generic sign-in/sign-out refuse it; the endpoints that make it functional come in the follow-up PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1819, opened 2026-08-11
feat: add offline credentials endpoints

DmytroZaichenkoDev:
Part 2/4 of the offline-delegation split (replaces #1815). Based on #1818.

Adds `/v1/user/offline-credentials` (status), `/signin` and `/signout`: a user grants DIAL one platform-wide offline credential, stored in their own bucket under a reserved id no app-scoped endpoint can address. Sign-in exchanges the code via the provider's `offlineClient`, verifies through the ID token that the code belongs to the caller, and records the issuer for later refresh. Refresh responses without a refresh token keep the existing one (RFC 6749 §6). All operations audited; shared error mapping extracted into `ExternalServiceErrors`.

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Consumer;
 
 @Slf4j
 @RequiredArgsConstructor
@@ -48,6 +49,18 @@ public class ResourceCredentialsService {
                                        ResourceAuthSettings resourceAuthSettings,
                                        ResourceSignInRequest resourceSignInRequest,
                                        String userId) {
+        addResourceCredentials(credentialsDescriptor, resourceAuthSettings, resourceSignInRequest, userId, credentials -> { });
+    }
+
+    /**
+     * As above, but runs {@code verifier} on the issued credentials <b>before</b> anything is stored, so a caller
+     * can refuse them. The ID token is verification material and is cleared rather than persisted.
+     */
+    public void addResourceCredentials(CredentialsDescriptor credentialsDescriptor,
+                                       ResourceAuthSettings resourceAuthSettings,
+                                       ResourceSignInRequest resourceSignInRequest,
+                                       String userId,
+                                       Consumer<ResourceCredentials> verifier) {
         log.info("Adding resource credentials for resourceId={}, bucket={}, credentialsLevel={}",
                 credentialsDescriptor.getResourceId(), credentialsDescriptor.getBucketName(), resourceSignInRequest.getCredentialsLevel());
         ResourceCredentialsFactory factory = resourceCredentialsFactoryProvider.getFactory(resourceSignInRequest.getAuthenticationType());
@@ -58,6 +71,9 @@ public class ResourceCredentialsService {
             // Offline-usage consent is per-user; record it so the on-behalf-of retrieval path can gate on it.
             resourceCredentials.setOfflineUsageConsent(resourceSignInRequest.isOfflineUsageConsent());
         }
+
+        verifier.accept(resourceCredentials);
+        resourceCredentials.setIdToken(null);
 
         byte[] encryptedBody = encrypt(credentialsDescriptor, resourceCredentials);
         resourceService.putResourceBytes(credentialsDescriptor.toResourceDescriptor(), encryptedBody, EtagHeader.ANY);
@@ -188,6 +204,13 @@ public class ResourceCredentialsService {
                 throw new IllegalArgumentException("Can't delete other user's personal credentials");
             }
         }
+    }
+
+    /** Deletes one record addressed directly, for records that are not app-scoped. */
+    public boolean deleteCredentialsRecord(CredentialsDescriptor credentialsDescriptor) {
+        log.info("Deleting resource credentials for resourceId={}, bucket={}",
+                credentialsDescriptor.getResourceId(), credentialsDescriptor.getBucketName());
+        return resourceService.deleteResource(credentialsDescriptor.toResourceDescriptor(), EtagHeader.ANY);
     }
 
     @Nullable
@@ -339,7 +362,11 @@ public class ResourceCredentialsService {
         resourceCredentials.setExpiresInSeconds(newAccessTokenResponse.getExpiresIn());
         resourceCredentials.setUpdatedAt(timeProvider.getCurrentTime());
         resourceCredentials.setAccessToken(newAccessTokenResponse.getAccessToken());
-        resourceCredentials.setRefreshToken(newAccessTokenResponse.getRefreshToken());
+        // RFC 6749 §6: no refresh token in the response leaves the existing one in force. Overwriting it with
+        // null would end offline access on providers that do not rotate.
+        if (newAccessTokenResponse.getRefreshToken() != null) {
+            resourceCredentials.setRefreshToken(newAccessTokenResponse.getRefreshToken());
+        }
         log.debug("Finished updating expired token for Resource: {}", resourceId);
     }
 
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
@@ -213,11 +213,20 @@ public class ResourceCredentialsService {
         return resourceService.deleteResource(credentialsDescriptor.toResourceDescriptor(), EtagHeader.ANY);
     }
 
+    /**
+     * NONE has no credential by definition; DIAL_NATIVE has none <b>per service</b> — a record left from before a
+     * switch to either type must never be served, and APPLICATION-level purging cannot reach USER-level leftovers.
+     */
+    private static boolean hasNoServiceCredential(ResourceAuthSettings authSettings) {
+        AuthenticationType type = authSettings.getAuthenticationType();
+        return type == AuthenticationType.NONE || type == AuthenticationType.DIAL_NATIVE;
+    }
+
     @Nullable
     public ResourceCredentials getRefreshedResourceCredentials(CredentialsLocator credentialsLocator,
                                                                ResourceAuthSettings authSettings,
                                                                String userSub) {
-        if (authSettings.getAuthenticationType() == AuthenticationType.NONE) {
+        if (hasNoServiceCredential(authSettings)) {
             return null;
         }
 
@@ -263,7 +272,7 @@ public class ResourceCredentialsService {
     public ResourceCredentials getRefreshedUserCredentials(CredentialsLocator credentialsLocator,
                                                            ResourceAuthSettings authSettings,
                                                            String ownerUserId) {
-        if (authSettings.getAuthenticationType() == AuthenticationType.NONE) {
+        if (hasNoServiceCredential(authSettings)) {
             return null;
         }
 
--- a/server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
@@ -7,6 +7,7 @@ import com.epam.aidial.core.config.ResourceAuthSettings;
 import com.epam.aidial.core.credentials.data.credentials.CredentialsDescriptor;
 import com.epam.aidial.core.credentials.data.credentials.ResourceCredentials;
 import com.epam.aidial.core.credentials.data.credentials.ResourceSignInRequest;
+import com.epam.aidial.core.credentials.service.ResourceAuthSettingsService;
 import com.epam.aidial.core.credentials.service.ResourceCredentialsService;
 import com.epam.aidial.core.openapi.annotations.ApiOperation;
 import com.epam.aidial.core.openapi.annotations.ApiResponse;
@@ -43,14 +44,14 @@ public class OfflineCredentialsController {
     private final ProxyContext context;
     private final AsyncTaskExecutor taskExecutor;
     private final ResourceCredentialsService resourceCredentialsService;
+    private final ResourceAuthSettingsService resourceAuthSettingsService;
     private final AccessTokenValidator tokenValidator;
-    private final Proxy proxy;
 
     public OfflineCredentialsController(Proxy proxy, ProxyContext context) {
-        this.proxy = proxy;
         this.context = context;
         this.taskExecutor = proxy.getTaskExecutor();
         this.resourceCredentialsService = proxy.getResourceCredentialsService();
+        this.resourceAuthSettingsService = proxy.getResourceAuthSettingsService();
         this.tokenValidator = proxy.getTokenValidator();
     }
 
@@ -74,7 +75,7 @@ public class OfflineCredentialsController {
         }
         taskExecutor.submit(() -> {
             // Same test the app listing uses, so the two endpoints cannot disagree about one user.
-            boolean connected = proxy.getResourceAuthSettingsService().hasUnexpiredCredentials(descriptor());
+            boolean connected = resourceAuthSettingsService.hasUnexpiredCredentials(descriptor());
             // Resolved only when not connected: a connected caller does not need it, and a provider that cannot
             // be resolved (or carries no offline client) is reported as unavailable rather than as an error —
             // sign-in is where refusing loudly matters.
--- a/server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
@@ -1,0 +1,233 @@
+package com.epam.aidial.core.server.controller;
+
+import com.auth0.jwt.exceptions.JWTDecodeException;
+import com.epam.aidial.core.config.AuthenticationType;
+import com.epam.aidial.core.config.CredentialsLevel;
+import com.epam.aidial.core.config.ResourceAuthSettings;
+import com.epam.aidial.core.credentials.data.credentials.CredentialsDescriptor;
+import com.epam.aidial.core.credentials.data.credentials.ResourceCredentials;
+import com.epam.aidial.core.credentials.data.credentials.ResourceSignInRequest;
+import com.epam.aidial.core.credentials.service.ResourceCredentialsService;
+import com.epam.aidial.core.openapi.annotations.ApiOperation;
+import com.epam.aidial.core.openapi.annotations.ApiResponse;
+import com.epam.aidial.core.openapi.annotations.ApiSchema;
+import com.epam.aidial.core.openapi.annotations.OpenApiDescriptions;
+import com.epam.aidial.core.server.Proxy;
+import com.epam.aidial.core.server.ProxyContext;
+import com.epam.aidial.core.server.data.OfflineCredentialsSignInRequest;
+import com.epam.aidial.core.server.data.OfflineCredentialsStatus;
+import com.epam.aidial.core.server.log.ExternalServiceAuditLog;
+import com.epam.aidial.core.server.security.AccessTokenValidator;
+import com.epam.aidial.core.server.util.CredentialsDescriptorFactory;
+import com.epam.aidial.core.server.util.ProxyUtil;
+import com.epam.aidial.core.server.validation.ValidationUtil;
+import com.epam.aidial.core.server.vertx.AsyncTaskExecutor;
+import com.epam.aidial.core.storage.http.HttpException;
+import com.epam.aidial.core.storage.http.HttpStatus;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.http.HttpHeaders;
+
+/**
+ * The user's own offline credentials: one refresh token per user, platform-wide, never handed to any application.
+ *
+ * <p>Deliberately not on the app-scoped external-service endpoints: those reach their blobs through
+ * {@code parseExternalServiceScope}, whose required shape these credentials sit outside.
+ */
+@Slf4j
+public class OfflineCredentialsController {
+
+    private final ProxyContext context;
+    private final AsyncTaskExecutor taskExecutor;
+    private final ResourceCredentialsService resourceCredentialsService;
+    private final AccessTokenValidator tokenValidator;
+    private final Proxy proxy;
+
+    public OfflineCredentialsController(Proxy proxy, ProxyContext context) {
+        this.proxy = proxy;
+        this.context = context;
+        this.taskExecutor = proxy.getTaskExecutor();
+        this.resourceCredentialsService = proxy.getResourceCredentialsService();
+        this.tokenValidator = proxy.getTokenValidator();
+    }
+
+    /** Status plus, when not connected, the parameters chat needs to build the authorization URL. */
+    @ApiOperation(
+            method = "GET",
+            path = "/v1/user/offline-credentials",
+            operationId = "getOfflineCredentials",
+            tags = {"External Services"},
+            responses = {
+                    @ApiResponse(code = 200, description = OpenApiDescriptions.RESPONSE_SUCCESS,
+                            body = @ApiSchema(implementation = OfflineCredentialsStatus.class)),
+                    @ApiResponse(code = 400),
+                    @ApiResponse(code = 401),
+                    @ApiResponse(code = 500)
+            }
+    )
+    public Future<?> getStatus() {
+        if (rejectNonUserCaller()) {
+            return Future.succeededFuture();
+        }
+        taskExecutor.submit(() -> {
+            // Same test the app listing uses, so the two endpoints cannot disagree about one user.
+            boolean connected = proxy.getResourceAuthSettingsService().hasUnexpiredCredentials(descriptor());
+            // Resolved only when not connected: a connected caller does not need it, and a provider that cannot
+            // be resolved (or carries no offline client) is reported as unavailable rather than as an error —
+            // sign-in is where refusing loudly matters.
+            return OfflineCredentialsStatus.of(connected, connected ? null : offlineClientOrNull());
+        })
+                .onSuccess(status -> context.respond(HttpStatus.OK, status))
+                .onFailure(error -> respondError("Can't read offline credentials", error));
+        return Future.succeededFuture();
+    }
+
+    /** Exchanges the authorization code and stores the refresh token in the caller's own bucket. */
+    @ApiOperation(
+            method = "POST",
+            path = "/v1/user/offline-credentials/signin",
+            operationId = "offlineCredentialsSignIn",
+            requestBody = @ApiSchema(implementation = OfflineCredentialsSignInRequest.class),
+            tags = {"External Services"},
+            responses = {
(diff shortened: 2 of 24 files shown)


server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceErrors.java
```diff
@@ -0,0 +1,65 @@
...
+import com.epam.aidial.core.storage.http.HttpException;
+import com.epam.aidial.core.storage.http.HttpStatus;
+import jakarta.validation.ConstraintViolationException;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
+
+/** One error mapping for the external-service and offline-credentials endpoints, which each had their own. */
+@Slf4j
+@UtilityClass
+public class ExternalServiceErrors {
```
astsiapanay: Is the Utility class used in more than one place? Do we really need this?

astsiapanay: The class name is a bit weird

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
```diff
@@ -126,6 +126,47 @@ private Future<ExtractedClaims> extractClaimsFromUserInfo(String accessToken, Su
...
     }
 
+    /**
+     * The provider that issued the caller's token. A userinfo-only provider carries no {@code issuerPattern} and so
+     * can never match here, and cannot offer offline credentials.
+     *
+     * @throws IllegalArgumentException when no token is presented or no provider matches.
+     * @throws JWTDecodeException when the access token is opaque, so the issuer cannot be read from it.
+     */
+    public IdentityProvider resolveProvider(String authHeader) {
```
astsiapanay: Let's don't expose provider itself. The class should have package level access.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
```diff
@@ -0,0 +1,231 @@
...
+                context, action, ExternalServiceErrors.asRuntime(result.cause()));
+    }
+
+    private IdentityProvider provider() {
+        try {
+            return proxy.getTokenValidator().resolveProvider(context.getRequest().getHeader(HttpHeaders.AUTHORIZATION));
+        } catch (JWTDecodeException e) {
+            // An opaque access token carries no issuer, so with several providers configured there is no way to
+            // tell which one to obtain offline credentials from.
+            throw new HttpException(HttpStatus.SERVICE_UNAVAILABLE,
```
astsiapanay: Status code 400?

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
```diff
@@ -0,0 +1,231 @@
...
+                    "Offline credentials are not available for this identity provider");
+        }
+    }
+
+    private ResourceAuthSettings requireOfflineClient(IdentityProvider provider) {
+        ResourceAuthSettings offlineClient = provider.getOfflineClient();
+        if (offlineClient == null) {
+            throw new HttpException(HttpStatus.SERVICE_UNAVAILABLE,
```
astsiapanay: 400?

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceErrors.java
DmytroZaichenkoDev: yes, it's used by three controllers: ExternalServiceCredentialsController, ExternalServiceManagementController, OfflineCredentialsController

DmytroZaichenkoDev: fixed

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
DmytroZaichenkoDev: fixed

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
DmytroZaichenkoDev: fixed

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
DmytroZaichenkoDev: fixed

ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
```diff
@@ -190,6 +206,13 @@ private void validateDeleteOperation(ResourceCredentials existingCredentials,
...
+    /** Deletes one record addressed directly, for records that are not app-scoped. */
+    public boolean deleteCredentialsRecord(CredentialsDescriptor credentialsDescriptor) {
+        log.info("Deleting resource credentials for resourceId={}, bucket={}",
+                credentialsDescriptor.getResourceId(), credentialsDescriptor.getBucketName());
+        return resourceService.deleteResource(credentialsDescriptor.toResourceDescriptor(), EtagHeader.ANY);
+    }
+
     @Nullable
     public ResourceCredentials getRefreshedResourceCredentials(CredentialsLocator credentialsLocator,
                                                                ResourceAuthSettings authSettings,
```
astsiapanay, line 227: **[correctness — CONFIRMED] Stale credential still served after switching a service to DIAL_NATIVE**

Two lines below this signature (line 220, not shown in this hunk), `getRefreshedResourceCredentials` only skips returning a credential when the current auth type is `NONE`:
```java
if (authSettings.getAuthenticationType() == AuthenticationType.NONE) {
    return null;
}
```
`DIAL_NATIVE` was added as `NONE`'s sibling everywhere else in this PR (e.g. `TokenRefreshStrategyFactory`) but not here — the identical guard in `getRefreshedUserCredentials` (line ~266) has the same gap.

**Failure scenario:** Service `svc` is OAUTH-configured; a user signs in and a USER-level credential with `authenticationType=OAUTH` is stored. Admin later switches `svc`'s config to `DIAL_NATIVE` (an in-place PUT triggers no automatic purge). The same user's app calls the credential-retrieval endpoint; since `authSettings.getAuthenticationType()` is now `DIAL_NATIVE`, not `NONE`, this guard doesn't short-circuit, and the stale record is returned unrefreshed. `AuthorizationHeaderProvider` then branches on the *stored* record's own `authenticationType` field (still `OAUTH`, set at creation) and returns a working `Authorization: Bearer <old-token>` header for a service just declared credential-less.

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
```diff
@@ -33,19 +33,36 @@ private ExternalServiceAuditLog() {
...
+     * self-service, so actor and owner are the same person.
+     */
+    public static void offlineCredentials(ProxyContext context, String action, RuntimeException error) {
+        AUDIT.info("event=offline_credentials action={} outcome={} actor={} user_id={} trace_id={}{}",
+                sanitizeToken(action), outcomeOf(error), actorEvidence(context),
+                sanitizeToken(context.getUserId()), context.getTraceId(),
+                reasonOf(error));
+    }
+
+    private static String outcomeOf(RuntimeException error) {
```
astsiapanay, line 54: **[correctness — CONFIRMED] Offline sign-in rejections audited as ERROR, not DENIED**

`outcomeOf(RuntimeException)` maps only `PermissionDeniedException` to `"DENIED"`. `OfflineCredentialsController.verifyIssuedForCaller` rejects a cross-user sign-in by throwing `HttpException(FORBIDDEN, ...)` — an unrelated sibling `RuntimeException` subtype — so it falls through to the `default -> "ERROR"` case.

**Failure scenario:** someone submits an authorization code obtained under a different account to `POST /v1/user/offline-credentials/signin`. The request correctly returns HTTP 403, but the `DIAL_OBO_AUDIT` log line records `outcome=ERROR` instead of `outcome=DENIED` (the exception propagates unwrapped through `AsyncTaskExecutor` and `ExternalServiceErrorHandler.asRuntime`, which returns the same `HttpException` instance, unmapped). Any alert/SIEM rule filtering this audit stream for `outcome=DENIED` to catch credential-hijack attempts misses this case; the sibling OBO-retrieval audit path already asserts exact outcome values in its tests, but this new path's tests only assert `outcome != SUCCESS`.

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
```diff
@@ -126,6 +127,77 @@ private Future<ExtractedClaims> extractClaimsFromUserInfo(String accessToken, Su
...
+    }
+
+    /**
+     * The provider that issued the caller's token. A userinfo-only provider carries no {@code issuerPattern} and so
+     * can never match here, and cannot offer offline credentials.
+     *
+     * @throws IllegalArgumentException when no token is presented or no provider matches.
+     * @throws JWTDecodeException when the access token is opaque, so the issuer cannot be read from it.
+     */
+    IdentityProvider resolveProvider(String authHeader) {
```
astsiapanay, line 167: **[correctness — CONFIRMED] Offline credentials unusable for opaque-token identity providers**

`resolveProvider(authHeader)`, which every offline-credentials endpoint depends on, only resolves the caller's identity provider by locally JWT-decoding the access token. Unlike the pre-existing general request-authentication path (`extractClaims`), it has no fallback to a userinfo-endpoint lookup for opaque-token providers — even though such providers are a real, tested, supported configuration elsewhere in this class.

**Failure scenario:** two or more identity providers are configured; one is userinfo-endpoint-only (opaque access tokens, `hasUserinfoUrl()=true`) and has an `offlineClient` configured. A user authenticated against that provider calls `GET /v1/user/offline-credentials` or `POST .../signin` with their normal (opaque) access token: `resolveProvider` throws `JWTDecodeException`, which gets caught and turned into `available: false` / HTTP 400 "not available" — even though an `offlineClient` is genuinely configured for that provider. The same user's ordinary chat requests work fine via the userinfo fallback that `resolveProvider` lacks.

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
```diff
@@ -126,6 +127,77 @@ private Future<ExtractedClaims> extractClaimsFromUserInfo(String accessToken, Su
...
+    public ResourceAuthSettings resolveOfflineClient(String authHeader) {
+        return resolveProvider(authHeader).getOfflineClient();
+    }
+
+    /**
+     * As above, for a recorded issuer — the refresh path, where there is no caller token to match on.
+     *
+     * @throws IllegalArgumentException when no provider matches.
+     */
+    public ResourceAuthSettings resolveOfflineClientByIssuer(String issuer) {
```
astsiapanay, line 146: **[simplification — CONFIRMED] Offline-credential refresh path is dead code, never wired up**

`resolveOfflineClientByIssuer(String issuer)` and the `ResourceCredentials.issuer` field it exists to serve (set at sign-in by `OfflineCredentialsController.verifyIssuedForCaller`) have zero callers/readers anywhere in the repository — the "refresh with no caller token present" path this PR lays groundwork for is never actually invoked.

**Scope:** bounded today because nothing consumes offline credentials for outbound calls yet either — worth confirming this is deliberately deferred to a follow-up PR rather than a gap that ships silently incomplete once something does start consuming them.

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
```diff
@@ -126,6 +127,77 @@ private Future<ExtractedClaims> extractClaimsFromUserInfo(String accessToken, Su
...
+
+    /**
+     * The provider that issued the caller's token. A userinfo-only provider carries no {@code issuerPattern} and so
+     * can never match here, and cannot offer offline credentials.
+     *
+     * @throws IllegalArgumentException when no token is presented or no provider matches.
+     * @throws JWTDecodeException when the access token is opaque, so the issuer cannot be read from it.
+     */
+    IdentityProvider resolveProvider(String authHeader) {
+        if (providers.size() == 1) {
```
astsiapanay: **[correctness — CONFIRMED] resolveProvider skips its own null-header contract**

This method's Javadoc promises `@throws IllegalArgumentException when no token is presented`, but when exactly one identity provider is configured (this line) it returns that provider unconditionally before ever inspecting `authHeader` — `resolveProvider(null)` does not throw as documented.

**Scope:** currently unreachable as a live bug — every real caller (`OfflineCredentialsController`'s methods) runs only after `rejectNonUserCaller()` has confirmed `context.getUserId() != null`, which is only ever set after a non-null Authorization header was already parsed for this same request. But the contract is false today, no test covers the single-provider case (the existing null-header test uses a 2-provider fixture), and it's a latent trap for the next caller that doesn't share that upstream guarantee.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
```diff
@@ -0,0 +1,233 @@
...
+    private String authHeader() {
+        return context.getRequest().getHeader(HttpHeaders.AUTHORIZATION);
+    }
+
+    private CredentialsDescriptor descriptor() {
+        return CredentialsDescriptorFactory.offlineCredentials(context);
+    }
+
+    /** These credentials belong to a person, so the caller must be one — not an app, not a userless key. */
+    private boolean rejectNonUserCaller() {
```
astsiapanay, line 220: **[reuse — CONFIRMED] rejectNonUserCaller duplicates existing per-request-key checks**

This method's per-request-key rejection re-implements a check already inlined three times in `ExternalServiceCredentialsController` (`signIn`, `signOut`, `getOboCredentials`) instead of extracting one shared helper both controllers call — both already share `ExternalServiceErrorHandler.respond(...)`, so the precedent for a shared helper exists.

**Cost:** a future change to the per-request-key rejection policy (status code, message wording, or the underlying condition) has to be found and updated in 4 separate places across 2 controllers; missing one leaves an inconsistent rejection for that endpoint.

server/src/test/java/com/epam/aidial/core/server/util/CredentialsPathReservationTest.java
```diff
@@ -0,0 +1,62 @@
...
+        String[] parts = CredentialsLocatorFactory.parseExternalServiceScope(
+                "applications/offline/external_services/dial");
+
+        assertEquals("offline", parts[0]);
+        assertEquals("dial", parts[1]);
+        assertNotEqualsReservedId("applications/config/offline/external_services/dial");
+    }
+
+    @Test
+    void everyAppScopedIdIsPathShaped() {
```
astsiapanay, line 57: **[test-coverage — CONFIRMED] Toolset collision test checks a string literal, not real code**

This test's toolset assertion (`assertTrue("toolsets/config/my-toolset".contains("/"))`) only checks a hardcoded string literal, unlike its application/external-service siblings in this file which call the real `CredentialsLocatorFactory.parseExternalServiceScope`. The class Javadoc claims the reserved `"offline"` id stays uncollidable "no matter how an application, toolset or external service is named," but the toolset half of that claim is unverified by any actual toolset code path.

**Cost:** a pre-existing test, `CredentialsLocatorFactoryTest.fromAnyUrl_ResourceFromConfig`, already derives this exact string from real code (`CredentialsLocatorFactory.fromAnyUrl`). If a future change to that id-shaping logic ever dropped the `/` separator for toolset-scoped ids, this reservation test would keep passing while the collision-safety guarantee it claims to enforce silently broke.

server/src/main/java/com/epam/aidial/core/server/data/OfflineCredentialsStatus.java
```diff
@@ -0,0 +1,36 @@
...
+ * Whether the caller has offline credentials and, when they do not, what chat needs to build the authorization URL.
+ * Carries only non-secret settings — the client secret stays in core, which performs the exchange.
+ *
+ * <p>{@code available} distinguishes "not connected yet" from "cannot connect": the caller's identity provider may
+ * carry no offline client at all, and chat should not offer a connect flow that can only end in an error.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record OfflineCredentialsStatus(boolean connected, boolean available, @JsonProperty("connect") Connect connect) {
+
+    public static OfflineCredentialsStatus of(boolean connected, ResourceAuthSettings offlineClient) {
```
astsiapanay, line 19: **[simplification — CONFIRMED] Redundant boolean logic in of()**

The branch guarded by `connected || offlineClient == null` returns `connected || offlineClient != null` for `available` — but that guard forces `offlineClient==null` whenever `connected` is false, so the disjunct can never actually flip the result away from plain `connected`. This reduces to `available = connected || offlineClient != null` computed once, with a single early return for the `Connect` case.

No incorrect output today (confirmed via truth table against this method's one real call site, which always passes `offlineClient=null` when `connected=true`), but the redundant double-condition relies on the caller's convention rather than the method's own logic to stay correct.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
```diff
@@ -0,0 +1,233 @@
...
+            tags = {"External Services"},
+            responses = {
+                    @ApiResponse(code = 200, description = OpenApiDescriptions.RESPONSE_SUCCESS,
+                            body = @ApiSchema(implementation = Boolean.class)),
+                    @ApiResponse(code = 400),
+                    @ApiResponse(code = 401),
+                    @ApiResponse(code = 500)
+            }
+    )
+    public Future<?> signIn() {
```
astsiapanay, line 104: **[simplification — PLAUSIBLE] signIn() decodes caller's JWT twice; catch blocks drifted**

`signIn()` resolves the caller's identity provider twice per request in multi-provider deployments: once via `requireOfflineClient()` (line 183) and again via `verifyIssuedForCaller()`'s `resolveIdTokenUserId` call (line 161) — both funnel through the same `resolveProvider(authHeader)`, decoding the identical access token a second time. The two near-duplicate resolution wrappers, `requireOfflineClient()` and `offlineClientOrNull()` (line 201), have also already drifted: one catches only `JWTDecodeException`, the other catches `JWTDecodeException | IllegalArgumentException`.

**Impact:** verified low — the repeated decode is a local base64+JSON parse over a small provider list (no I/O), and today's callers are shielded from the catch-block asymmetry by upstream guarantees. Flagged as a maintainability risk (two divergent copies to keep in step) rather than a live bug.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
```diff
@@ -0,0 +1,233 @@
...
+                    @ApiResponse(code = 500)
+            }
+    )
+    public Future<?> getStatus() {
+        if (rejectNonUserCaller()) {
+            return Future.succeededFuture();
+        }
+        taskExecutor.submit(() -> {
+            // Same test the app listing uses, so the two endpoints cannot disagree about one user.
+            boolean connected = proxy.getResourceAuthSettingsService().hasUnexpiredCredentials(descriptor());
```
astsiapanay: **[conventions — CONFIRMED] getStatus() breaks this class's own field-capture convention**

The constructor captures `resourceCredentialsService` and `tokenValidator` as fields from `proxy`, but this line instead calls `proxy.getResourceAuthSettingsService()` inline rather than capturing a third field the same way — `Proxy` exposes all three as equally stable final fields, so there's no lifecycle reason for the asymmetry. Not a functional bug, but an unexplained inconsistency worth a one-line fix (add the field).

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
```diff
@@ -502,12 +507,23 @@ private void logClaims(Map<String, Object> claims) {
         }
     }
 
+    /** The user id this provider would derive from an ID token, via the same {@code userIdPath} as a request. */
+    String extractUserIdFromIdToken(String idToken) {
+        return extractStringClaim(claimsOf(decodeJwtToken(idToken)), userIdPath);
+    }
+
+    /** Whether this provider claims the given issuer. False when no {@code issuerPattern} is configured. */
+    boolean matchesIssuer(String issuer) {
```
astsiapanay, line 519: **[conventions — CONFIRMED] Comment purely restates the code below it**

`matchesIssuer`'s Javadoc ("Whether this provider claims the given issuer. False when no issuerPattern is configured.") adds no information beyond what the method name and its one-line three-term boolean body already convey — unlike its neighbor `hasIssuerPattern`, whose comment states a real interpretive point ("a non-match proves nothing") not visible in the code.

Per CLAUDE.md's Code Style rule, "Use comments sparingly. Only comment complex code." — a 3-clause boolean isn't complex code by the repo's own bar.

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
```diff
@@ -48,6 +49,18 @@ public void addResourceCredentials(CredentialsDescriptor credentialsDescriptor,
...
                                        ResourceSignInRequest resourceSignInRequest,
                                        String userId) {
+        addResourceCredentials(credentialsDescriptor, resourceAuthSettings, resourceSignInRequest, userId, credentials -> { });
+    }
+
+    /**
+     * As above, but runs {@code verifier} on the issued credentials <b>before</b> anything is stored, so a caller
+     * can refuse them. The ID token is verification material and is cleared rather than persisted.
+     */
+    public void addResourceCredentials(CredentialsDescriptor credentialsDescriptor,
```
astsiapanay, line 59: **[simplification — PLAUSIBLE] addResourceCredentials verifier param is broader than its one caller needs**

This new 5-arg overload takes a fully generic `Consumer<ResourceCredentials> verifier` added for exactly one real implementation anywhere in the repo (`OfflineCredentialsController::verifyIssuedForCaller`), which both throws and mutates the credentials object.

That said, the verifier does need access to the transient `idToken` field in the narrow window before it's nulled for persistence, which is a real structural reason it lives inside the service rather than the controller — a genuine tradeoff rather than a clear-cut mistake. Flagging for awareness in case a second caller's needs diverge from this shape.

astsiapanay (commented):
Automated code review (max effort: 10 finder angles + 1-vote verification + gap sweep). 12 verified findings posted inline below, ranked by severity.

Two additional verified findings from the same review are **not** included here because they belong to #1818 ("add DIAL_NATIVE auth type"), not this PR's diff:
- `DialNativeAuthSettingsValidator` (new in #1818) duplicates `NoneAuthSettingsValidator` verbatim instead of reusing it.
- `ExternalServiceService.putExternalService` validates new auth settings *before* an unconditional secret-preservation step reattaches the old client secret, with no re-check that the auth type didn't change — so converting a service to `DIAL_NATIVE` can silently leave its old OAuth secret persisted underneath a type whose entire premise is "no credential is declared." This is arguably the most severe finding of the whole review, but the touched file (`ExternalServiceService.java`) isn't part of either PR's diff, so it has no line to anchor a comment to here — flagging it in this summary instead.

ai-dial-actions:
<h1>Dependency Review</h1>
None

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
DmytroZaichenkoDev: fixed

server/src/main/java/com/epam/aidial/core/server/log/ExternalServiceAuditLog.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
DmytroZaichenkoDev: valid, but deliberate for now: resolving an opaque token requires a userinfo call, which would add async I/O into this sync path. We can address in a follow-up if we need. For now let's try to make it work with at least few basic idps.

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
DmytroZaichenkoDev: the consumers are in #1821, where redemption resolves the provider by the recorded issuer to refresh the owner's credentials with no caller token present.

server/src/main/java/com/epam/aidial/core/server/security/AccessTokenValidator.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/data/OfflineCredentialsStatus.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
DmytroZaichenkoDev: intentional: sign-in gives specific 400s, status only reports available:false — hence different catches. The extra decode is a cheap local parse.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/security/IdentityProvider.java
DmytroZaichenkoDev: Fixed

credentials/src/main/java/com/epam/aidial/core/credentials/service/ResourceCredentialsService.java
DmytroZaichenkoDev: Agree with the tradeoff as you stated it: the verifier has to see idToken in the window before it's cleared for persistence, so it lives inside the service. The breadth is what the one caller needs — it both refuses (throws) and records the issuer (mutates). Will narrow the shape if a second caller's needs diverge.

server/src/main/java/com/epam/aidial/core/server/controller/OfflineCredentialsController.java
DmytroZaichenkoDev: Partially agree: the three checks in ExternalServiceCredentialsController pre-date this PR, their messages are endpoint-specific, and getCredentials has the opposite polarity (it requires a per-request key). A shared helper would abstract four slightly different behaviors — I'd keep as is; happy to extract one for the same-polarity sites in a cleanup if you feel strongly.

server/src/test/java/com/epam/aidial/core/server/util/CredentialsPathReservationTest.java
DmytroZaichenkoDev: Fixed

astsiapanay (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1827, opened 2026-08-12
fix: Fix OpenAPI schema for LocalizedValue fields

### LINKED ISSUE #1812: display_name/intro/description fields in OpenAPI spec typed as LocalizedValue object instead of oneOf[string, map]
### Name and Version

0.46.1

1. Open open_api_core.yaml
2. Check the `display_name`, `description`, `intro` fields on Application, Model, ToolSet, Interceptor (and their *Data variants)
3. It's defined as $ref to LocalizedValue: { localeMap, plainValue }
4. Generate TS SDK from this spec — display_name becomes { localeMap?, plainValue? }

`display_name`, `description` and `intro` fields should be typed as a plain string or a locale map (e.g. { "en": "text" }), since that's what core actually returns. In OpenAPI this should be `oneOf: [string, object]`, not a ref to LocalizedValue.

### What do you see instead?

`display_name`, `description` and `intro` fields is typed as { localeMap, plainValue } object, which doesn't match the real response and produces wrong types in the generated SDK.

Also inconsistent: ExternalService uses plain string for the same kind of field, while other schemas use LocalizedValue.

### Additional information

_No response_

Katerina-Charakhovich:
### Applicable issues

- fixes #1812 

### Checklist

### DIFF
--- a/config/build.gradle
+++ b/config/build.gradle
@@ -1,4 +1,6 @@
 dependencies {
+    implementation project(':openapi-annotations')
+
     // External dependencies
     implementation("com.fasterxml.jackson.core:jackson-databind:${jackson_core_version}")
     implementation("com.networknt:json-schema-validator:${networknt_version}")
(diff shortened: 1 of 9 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1830, opened 2026-08-12
feat: add config-file migration endpoint #1784

### LINKED ISSUE #1784: Slice D — Config-file migration endpoint POST /v1/admin/config/file/migrate

Part of #1781.

## Goal

## Endpoint

### Request

```jsonc
{
}
```

     `platform`).
     insufficient for these types.

```jsonc
200 OK
{ "results": [
] }
```

## Testing

  lock are exercised):
    short name;
    `Config`);
  - non-admin is rejected.

## Depends on

KirylKurnosenka:
Adds an admin-triggered, on-demand endpoint that copies file-defined config entities into the `platform` blob bucket, so they become API-managed without changing what clients see or send. Supports dry-run and is idempotent on re-invocation.

### Applicable issues

- fixes #1784

### Description of changes

- New `POST /v1/admin/config/file/migrate` (`ConfigFileMigrateController`), admin-only, migrating models/interceptors/roles/keys/routes/applications/toolsets/schemas/catalog_schemas/settings on request, with a `types` filter and `dryRun` flag.
- Reuses `AdminApplyController`'s per-kind write pipeline (`validateOnly`, `mutateScratch`, `newScratch`, and a newly extracted `applyEntries`) instead of duplicating per-type encryption/validation/dup-checks; `applyEntries` and the `DEPENDENCY_ORDER_COMPARATOR` it now shares with `applyBatch` are also reused as-is by the migration endpoint.
- New `SchemaMigrationNaming` helper mints a deterministic canonical blob name for migrated schemas (`sha256($id)` + sanitized display-name slug), since a schema's canonical id is intentionally decoupled from its `$id`.
- Response statuses distinguish dry-run from real outcomes (`migrated`/`skipped`/`failed` vs. `would_migrate`/`would_skip`/`would_fail`) so a result can't be mistaken for an actual write.
- New `ConfigFileMigrateApiTest` integration coverage: happy path, idempotent re-run, dry-run (writes nothing), toolset secret round-trip, same-display-name schema disambiguation, settings migration, `types` filtering, invalid type (400), non-admin (403).

### Notes for reviewers

- **Base branch**: this PR targets `feat/issue-1783` (#1813) rather than `development`, since this endpoint's presence/idempotency checks and the `applySingle`/`EntityResult` reuse depend on that branch's short-name-derivation and schema `$id`-alias-index work. This lets both PRs be reviewed in parallel.
- **Reminder — once #1813 merges**: rebase this branch onto `development` before switching the base:
  ```
  git fetch origin
  git rebase --onto origin/development feat/issue-1783 feat/issue-1784
  git push --force-with-lease
  ```
  Then change this PR's base to `development`. This keeps the diff correct regardless of whether #1813 merges as a merge commit or a squash.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA ca2723a23266a3e0e9282cdc56a24ce1a333721c.</blockquote>
None

KirylKurnosenka:
/deploy-review

>GitHub actions run: [33052669271](https://github.com/epam/ai-dial-ci/actions/runs/33052669271)

| [deploy-review](https://chat-ai-dial-core-pr-1830.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/33052669271/job/98451716475) |
| chat | [Failed ❌](https://github.com/epam/ai-dial-ci/actions/runs/33052669271/job/98454679548) |

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -495,6 +495,27 @@ public class ControllerSelector {
                     proxy.getLockService());
             return controller::handle;
         });
+        post(RouteTemplate.CONFIG_FILE_MIGRATE, (proxy, context, pathMatcher) -> {
+            ConfigAuthorizationService authService = new AdminRoleAuthorizationService(proxy.getAccessService());
```
astsiapanay: Why do need to allocate a new service on every request?

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -495,6 +495,27 @@ public class ControllerSelector {
...
+            MergedConfigStore mergedConfigStore = (MergedConfigStore) proxy.getConfigStore();
```
astsiapanay: Perhaps it seems to be good to cast an interface to implementation. 
Do you want to extend the interface instead?

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -495,6 +495,27 @@ public class ControllerSelector {
...
+            // Reuses AdminApplyController's per-kind write pipeline (applySingle/validateOnly/
+            // mutateScratch/newScratch) rather than re-implementing per-type writes; `context` here
+            // is only ever used by AdminApplyController.handle()/process(), neither of which this
+            // controller calls, so sharing an instance across the two controllers is safe.
+            AdminApplyController applier = new AdminApplyController(
```
astsiapanay: That's a very bad idea to pass a controller as an argument to another controller. It's ok to reuse a common method over inheritance or a static helper method.

server/src/main/java/com/epam/aidial/core/server/data/config/migration/ConfigFileMigrateStatus.java
```diff
@@ -0,0 +1,17 @@
+package com.epam.aidial.core.server.data;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum ConfigFileMigrateStatus {
+    MIGRATED,
```
astsiapanay, line 7: Please add a description what each status means

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Similar to other "admin" endpoints

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Similar to other "admin" endpoints

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33108995403)

server/src/main/java/com/epam/aidial/core/server/data/config/migration/ConfigFileMigrateStatus.java
KirylKurnosenka: Added

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Refactored

server/src/main/java/com/epam/aidial/core/server/service/ConfigValidateService.java
```diff
@@ -0,0 +1,137 @@
+package com.epam.aidial.core.server.service;
```
astsiapanay: can we put similar services to a single package?

server/src/main/java/com/epam/aidial/core/server/service/ConfigValidateService.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Precheck engine behind {@code /v1/admin/validate} and the {@code precheck=true} phase of
+ * {@code /v1/admin/apply}: validates a manifest against {@code scratch} without writing anything.
+ * Shares manifest-shape helpers with {@link ConfigApplyService} via {@link ConfigManifestSupport}.
+ */
+public class ConfigValidateService {
```
astsiapanay: ConfigValidationService?

server/src/main/java/com/epam/aidial/core/server/service/ConfigManifestSupport.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Manifest-shape helpers shared by {@link ConfigApplyService} (real-apply) and
+ * {@link ConfigValidateService} (precheck) — parsing/naming rules, dependency ordering, scratch
+ * construction/mutation, and cross-entity validation that neither side owns exclusively.
+ */
+public final class ConfigManifestSupport {
```
astsiapanay: Annotate with @UtilityClass

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigApplyService.java
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Real-apply engine behind the write phase of {@code /v1/admin/apply} and
+ * {@code /v1/admin/config/file/migrate} — writes each config entity to blob storage directly or
+ * via {@link ApplicationService}/{@link ToolSetService}. Shares its precheck validation with
+ * {@link ConfigValidateService} and manifest-shape helpers (parsing, scratch, dependency
+ * ordering) with {@link ConfigManifestSupport}.
+ */
+@Slf4j
+public class ConfigApplyService {
```
astsiapanay, line 50: I guess all three classes could be merged into a single one, e.g. EntityConfigurationService.

Does the idea sound good to you?

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
```diff
@@ -0,0 +1,45 @@
...
+import com.epam.aidial.core.storage.http.HttpStatus;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * Pure JSON (de)serialization helpers for admin config entities, shared by
+ * {@code ConfigResourceController}, {@code ConfigApplyService} and {@code ConfigValidateService}.
+ * Deliberately dependency-free (no service/controller state) so both layers can depend on it.
+ */
+public final class ConfigEntityCodec {
```
astsiapanay, line 16: It could be moved to the package service.config

server/src/main/java/com/epam/aidial/core/server/data/ConfigFileMigrateRequest.java
```diff
@@ -0,0 +1,5 @@
+package com.epam.aidial.core.server.data;
```
astsiapanay: I suppose similar classes should be put to the same package

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
astsiapanay, line 16: We already have a bunch of classes having JSON helper methods:
ProxyUtil, JsonUtil and ConfigEntityCodec. The JSON methods of ProxyUtil and ConfigEntityCodec could be merged into JsonUtil. Let's don't make a new helper for a single use case. Secondly some methods like treeToEntity are very specific and doesn't seem to be generic helper methods.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigFileMigrateController.java
```diff
@@ -0,0 +1,474 @@
...
+                .onFailure(error -> context.respond(HttpStatus.BAD_REQUEST,
+                        "Failed to read request body: " + error.getMessage()));
+        return Future.succeededFuture();
+    }
+
+    private void process(Buffer body) {
+        ConfigFileMigrateRequest request;
+        try {
+            String text = body.toString(StandardCharsets.UTF_8);
+            request = ProxyUtil.MAPPER.readValue(text.isEmpty() ? "{}" : text, ConfigFileMigrateRequest.class);
```
astsiapanay, line 139: Wouldn't empty request body be treated as a validation error?

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -476,23 +478,44 @@ public class ControllerSelector {
...
             AdminValidateController controller = new AdminValidateController(
-                    context, authService, mergedConfigStore,
-                    proxy.getResourceService(), proxy.getTaskExecutor());
+                    context, authService, mergedConfigStore, validateService, proxy.getTaskExecutor());
             return controller::handle;
         });
         post(RouteTemplate.CONFIG_APPLY, (proxy, context, pathMatcher) -> {
             ConfigAuthorizationService authService = new AdminRoleAuthorizationService(proxy.getAccessService());
             MergedConfigStore mergedConfigStore = (MergedConfigStore) proxy.getConfigStore();
+            ConfigApplyService applyService = new ConfigApplyService(
```
astsiapanay: Please instantiate the service in advance while the server is being initialized. We have a clear pattern in the Core: don't create service instances per request.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -476,23 +478,44 @@ public class ControllerSelector {
...
+                    mergedConfigStore, proxy.getResourceService(),
+                    mergedConfigStore.getSecretFieldProcessor(),
+                    mergedConfigStore.isSoftValidation(),
+                    proxy.getApiKeyStore(),
+                    proxy.getApplicationService(),
+                    proxy.getToolSetService());
+            ConfigValidateService validateService = new ConfigValidateService(
```
astsiapanay: the same thing

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
astsiapanay: We should change that practice because that wastes resources.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
astsiapanay: replied above

server/src/main/java/com/epam/aidial/core/server/service/ConfigApplyService.java
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * Applies each entry via {@link #applySingle}, mutating {@code scratch} after every success so
+     * later entries in the same list see earlier ones, then flushes every in-memory change as a
+     * single partial-update swap. Shared by the real-apply phase of {@code /v1/admin/apply} and by
+     * {@code /v1/admin/config/file/migrate}, which builds its own entry list and scratch (with any
+     * shadowed file entries already removed) before calling this.
+     */
+    public List<EntityResult> applyEntries(List<AdminManifest> entries, Config scratch) {
+        List<EntityResult> results = new ArrayList<>();
+        // Slice 4S.4: collect partial-update changes per applied entity; flush as one applyBatch
```
astsiapanay: Where to find the slice 4S.4?
I guess you refer to nonexistent documents.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Should we change that in this PR for all admin endpoints?

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: To be honest I didn't catch what you are suggesting

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigApplyService.java
KirylKurnosenka: I would reject this. Each class has now narrower scope. Combining all classes into single one will lead to around 650 lines of code and wider responsibility. I don't think that sounds good at least for manual support.

server/src/main/java/com/epam/aidial/core/server/controller/ConfigFileMigrateController.java
KirylKurnosenka: Similar to `AdminApplyController` and `AdminValidateController`

server/src/main/java/com/epam/aidial/core/server/service/ConfigValidateService.java
KirylKurnosenka: Put into `service.config` package

server/src/main/java/com/epam/aidial/core/server/service/ConfigValidateService.java
KirylKurnosenka: Renamed

server/src/main/java/com/epam/aidial/core/server/service/ConfigManifestSupport.java
KirylKurnosenka: Annotated

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
KirylKurnosenka: Moved

server/src/main/java/com/epam/aidial/core/server/data/ConfigFileMigrateRequest.java
KirylKurnosenka: Placed into `data.config.migration` package

server/src/main/java/com/epam/aidial/core/server/service/ConfigApplyService.java
KirylKurnosenka: Removed reference

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33162919033)

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Refactored

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Refactored

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Refactored

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33166543452)

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
KirylKurnosenka: Those methods were just extracted from controller in order to not create inverted dependency `service -> controller`, since they are also used in services. Also those methods throw `HttpException` which is not quite good, but changing exception to some generic/domain one would lead to bigger refactoring with catching and rethrowing where it's needed. Moreover serialization/deserialization depends on what `ObjectMapper` is used: whereas `serializeForBlob` explicitly shows which mapper is used, `treeToEntity` isn't. So I think they could not be simply placed in generic `JsonUtil`/ProxyUtil.

KirylKurnosenka:
/deploy-review

>GitHub actions run: [33167985944](https://github.com/epam/ai-dial-ci/actions/runs/33167985944)

| [deploy-review](https://chat-ai-dial-core-pr-1830.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/33167985944/job/98837889947) |
| chat | [Failed ❌](https://github.com/epam/ai-dial-ci/actions/runs/33167985944/job/98840265216) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
astsiapanay: Proxy.getConfigStore() returns the interface. Later you cast the interface to a concrete implementation. That looks like an anti pattern. Either remove the interface or extend the interface. Or let's talk in private

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
```diff
@@ -474,25 +476,30 @@ public class ControllerSelector {
...
-            return controller::handle;
+            ConfigValidationService validationService = proxy.getConfigValidationService();
+            return new AdminValidateController(
+                    context, authService, mergedConfigStore, validationService, proxy.getTaskExecutor());
         });
         post(RouteTemplate.CONFIG_APPLY, (proxy, context, pathMatcher) -> {
-            ConfigAuthorizationService authService = new AdminRoleAuthorizationService(proxy.getAccessService());
+            ConfigAuthorizationService authService = proxy.getConfigAuthService();
             MergedConfigStore mergedConfigStore = (MergedConfigStore) proxy.getConfigStore();
+            ConfigApplyService applyService = proxy.getConfigApplyService();
```
astsiapanay: I would suggest pass the proxy as an arg to the controller. Since it has all references to services. That could make lambdas more compact.

It's nice to have.

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
astsiapanay: Right. Perhaps they are not helpers at all. We can inject the methods to controller classes. As I understand the main intention is code de-duplication. However they are very specific and can't be reused in any other places.

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
KirylKurnosenka: Done

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
KirylKurnosenka: Those methods were placed in `ConfigResourceController`
image
and used in `AdminApplyController` as well. When we extracted business logic from `AdminApplyController` controller into corresponding services, those methods started to be used in services, which created illegal (at least for me) dependency (services depend on controllers). That's why I placed those methods in separate util class, so they can be used in services and controllers

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/ControllerSelector.java
astsiapanay: let's postpone in the next PR

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java
KirylKurnosenka: Refactored

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1836, opened 2026-08-17
feat: masked client secret in api responses #1508

### LINKED ISSUE #1508: Masked client secret return in api responses from core
### Name and Version

0.42.4

Masked client secret return in api responses from core

Masked client secret return in api responses from core

_No response_

DmytroZaichenkoDev:
Read responses now carry `client_secret_hint` — the trailing characters of the stored OAuth client secret — so a management UI can show which secret is configured without core ever revealing it. The secret itself stays absent from every response, and the new field is read-only, so the existing "omitted means preserve" write contract is unchanged.

### Applicable issues

- fixes #1508

### Description of changes

**The field**

- `client_secret_hint` on `ResourceAuthSettings`, `@JsonProperty(READ_ONLY)` so it serializes but is never bound from a request.
- How much is revealed scales with the secret's length so the fragment never approaches the whole value (the cap-what-you-show shape of PCI DSS 4.0 §3.4.1): nothing below 8 characters, 2 characters below 12, 4 from 12 on. Every secret a real authorization server issues is 32–64 characters (RFC 6749 §10.10 forces ~128 bits), so the width is constant in practice and the hint says nothing about the secret's size. The 2-character band covers legacy values and the shortest secrets the write validation below still accepts (8–11 characters), where it reveals a quarter of the value at most.
- No `***` sentinel in `client_secret` itself: that would collide with the "null/absent = keep the stored secret" contract used by `ResourceAuthSettingsService`, `ExternalServiceService`, `UserExternalServiceService` and `SecretFieldProcessor`, and it was already tried and removed once (c4c4bd0d).

**Who sees it**

- Write access on the toolset/application resource, manage access on the external-services API, admin on the config API.
- Detail reads only. `GET /v1/applications/{appId}/external-services` — the listing — carries no hint and does no decryption behind it: a listing answers which services exist, not which secret each holds, and every response carrying a fragment of a secret is another place it reaches a log, a proxy cache or browser history. The single-service GET, the toolset and application GETs, and PUT responses all keep it.
- Consumer surfaces (`/openai/toolsets`, `/openai/applications`) are untouched — what a signed-in user needs is answered by the existing auth status fields, not by knowing whether a secret exists. `ResourceAuthSettingsData` deliberately does not carry the field.

**Decrypt before hinting**

- Stored secrets are ciphertext on these read paths, so the hint costs a decrypt (`ToolSetService.redactAuthSettings`, `ExternalServiceService.decryptSecretsForResponse`). A blob that fails to decrypt — legacy plaintext, rotated key — yields no hint rather than failing the read or hinting at ciphertext. Config/platform apps and the admin config API read the merged `Config`, which already holds plaintext.
- Decryption happens only on the read paths that expose the hint; write and consent paths re-read the blob themselves and skip it.
- Cost: the content encryption key is cached per bucket (`CachedContentEncryptionKeyManager`, enabled by default, 10k entries, 10-minute expiry), so the marginal cost per service is one AEAD over a 32–64 byte string, not a key fetch. The single-service GET decrypts only the service it serves; the application GET decrypts its inline services, bounded by services-per-app. Deployments that disable that cache pay a real key fetch per decrypt.

**Write-side validation**

- New `ClientSecretValidation`: caller-supplied `client_secret` must be non-blank, free of control characters, and at least 8 characters. There was no value checking at all before — `"   "` was accepted and stored.
- API writes only. Dynamic client registration keeps whatever the authorization server returned (validation runs before enrichment), and config-file entries are untouched because `ConfigPostProcessor` drops rather than rejects.
- Validation runs before preserve-on-omit in every write path, so updates that omit the secret — and entities already holding a shorter one — are unaffected. Only an explicit re-submission of a sub-8-character secret is now rejected.

**Tests and docs**

- New unit tests for the tier boundaries, read-only binding and the validation rules; API tests covering writer-sees-hint, read-only-share-does-not, listing-does-not, GET→PUT round-trip preserving the real secret (asserted through the outgoing token request), the config API's own read path, and both rejection paths.
- `docs/open_api_core.yaml` regenerated; the generator emits `readOnly: true` for the field. The schema is shared, so it still declares `client_secret_hint` on the listing's element type even though listings do not populate it — OpenAPI cannot express "detail reads only" on a shared schema.

**Known gaps**

- Redaction lives in two places: POJO mutation (`ResourceAuthSettings.redactSecrets*`) and a hand-maintained JSON field list (`ConfigResourceController.redactSecretFields`). Both are documented as needing to move together. A third projection surface, `FileConfigController.handleSingle`, redacts neither — config-file entities are not encrypted at rest, so it returns `client_secret` in plaintext to admins. Pre-existing and outside this diff; worth a separate issue.
- `HINT_MIN_SECRET_LENGTH` (8) equals `ClientSecretValidation.MIN_LENGTH` (8), so the shortest secret a write accepts today is also the one with the worst disclosure ratio — 2 of 8 characters. Deliberate: raising the floor would hide the hint for secrets real issuers never emit anyway.
- The hint puts crypto on read paths that previously did none, which brings two side effects. `CredentialEncryptionService.encryptOrDecrypt` calls `getOrCreateKey` before any length check, so a toolset whose stored `client_secret` is *legacy plaintext that happens to parse as base64* will, in a bucket that has never encrypted anything, generate and persist a CEK as a side effect of a GET — the decrypt then fails and no hint is emitted. The key is the one the next write would have created anyway. Separately, the per-service cost is a cheap AEAD only on a CEK cache hit; on a miss (cold start, eviction, 10-minute expiry) `getOrCreateKey` takes a distributed lock and does a blob read plus a KMS unwrap, once per bucket per window. Judged not worth guarding: the alternative is a `minEncryptedLength` pre-check in shared encryption code or a third copy of the classification logic.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (commented):
Automated review pass (medium effort, 8 finder angles + verification). No correctness bugs survived verification; two initially-promising leads were checked against the actual code and refuted (decrypt-before-validate reordering was already the pre-PR behavior; ApplicationController's missing hint is intentional per the PR description's "consumer surfaces are untouched" note). Findings below are efficiency/cleanup/doc-accuracy items.

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
```diff
@@ -345,6 +352,32 @@ private ExternalServiceData toData(String appId, String serviceId, ExternalServi
                 .setAuthSettings(safe);
     }
 
+    /**
+     * Decrypts inline secrets so {@link #toData} derives the hint from the same code path for every service.
+     * {@code inline} must be the same value {@code manageableServices} was given — a caller limited to its own
+     * user-authored services must not have the inline ones decrypted on its behalf. Nothing to do for a config
+     * app (the merged config already holds plaintext) or for user-authored services (decrypted per resource).
+     */
+    private void revealInlineSecrets(ResolvedApp resolved, boolean inline) {
```
astsiapanay: **[efficiency]** `listExternalServices`/`getExternalService` now unconditionally decrypt every inline service's `client_secret` (via `revealInlineSecrets`/`revealInlineSecret`) on every manager request, purely to compute a hint — these read paths previously did zero crypto work. For an application with many inline external services, a manager repeatedly listing services now pays one decrypt (CEK fetch/create + AEAD decrypt) per service per request. Same pattern in `ResourceController.getApplicationData` (line ~562, `decryptSecretsForResponse`) and `ToolSetService.redactAuthSettings` (line 112) for the toolset GET path.

server/src/main/java/com/epam/aidial/core/server/service/ExternalServiceService.java
```diff
@@ -119,9 +121,14 @@ private static CredentialsLocator applicationCredentialsLocator(ResourceDescript
...
     }
 
+    /**
+     * Writes the definition and returns it as persisted, {@code clientSecret} in plaintext — the caller must
+     * redact it before it leaves the process.
+     */
     public ExternalService putExternalService(ResourceDescriptor resource, String serviceId, ExternalService service, String author) {
         verifyApplication(resource);
         MutableBoolean typeChanged = new MutableBoolean(false);
+        MutableObject<String> storedSecret = new MutableObject<>();
```
astsiapanay: **[simplification]** `putExternalService` captures the persisted `client_secret` into a `MutableObject<String>` before `encryptSecrets` mutates `service` in place, then writes it back afterward — and `UserExternalServiceService.put` (lines 83-97) duplicates this identical encrypt-then-restore-plaintext pattern verbatim. A future fix to how the persisted secret is restored after encryption would need to be applied in both places separately.

server/src/main/java/com/epam/aidial/core/server/service/ExternalServiceService.java
```diff
@@ -173,6 +186,36 @@ public void decryptSecrets(ResourceDescriptor resource, Application application)
         processSecrets(resource, application, false);
     }
 
+    /**
+     * Decrypts in place for a read that wants the {@code clientSecretHint}. A secret that fails to decrypt
+     * (legacy plaintext, rotated key) is dropped rather than left as ciphertext, so it yields no hint while
+     * the read still succeeds and every other service keeps its own.
+     */
+    public void decryptSecretsForResponse(ResourceDescriptor resource, Application application) {
```
astsiapanay, line 189: **[reuse]** `decryptSecretForResponse` (single-service, line 200) re-implements the null/empty checks that `eachServiceWithSecret` (line 243) already performs for the plural `decryptSecretsForResponse` (this line), instead of routing the single-service case through the same shared iteration/filter logic. A future change to the skip conditions only needs to be made in `eachServiceWithSecret` for the list path to pick it up, but the copy-pasted guard here would silently keep the old behavior for the single-service GET.

config/src/main/java/com/epam/aidial/core/config/ResourceAuthSettings.java
```diff
@@ -77,11 +92,60 @@ public class ResourceAuthSettings {
     private Boolean dynamicallyRegistered;
 
     /**
-     * Returns a copy with credential material ({@code clientSecret}, {@code codeVerifier}) removed — the single
-     * place read responses strip secrets, so a newly added secret field only has to be cleared here. The
-     * receiver is left unchanged.
+     * Returns a copy with credential material removed; the receiver is left unchanged. A newly added secret
+     * field has to be cleared here <em>and</em> in {@code ConfigResourceController.redactSecretFields}, which
```
astsiapanay, line 96: **[altitude]** Secret-field redaction is implemented twice — once as POJO mutation here (`redactSecrets`/`redactSecretsKeepingHint`) and once as a hand-maintained JSON-node field list in `ConfigResourceController.redactSecretFields` (line ~1725) — with this javadoc admitting both must be updated whenever a secret field is added. A future secret field redacted here but not mirrored into `ConfigResourceController.redactSecretFields`'s `settings.remove(...)` calls would leak through the config API's GET endpoints even though every other read path is safe.

config/src/test/java/com/epam/aidial/core/config/ResourceAuthSettingsTest.java
```diff
@@ -0,0 +1,113 @@
...
+        // (32-64 characters) the hint says nothing about the secret's size.
+        assertEquals(ResourceAuthSettings.HINT_LENGTH,
+                oauthSettings("0123456789ab").withoutSecretsKeepingHint().getClientSecretHint().length());
+        assertEquals(ResourceAuthSettings.HINT_LENGTH,
+                oauthSettings("x".repeat(64)).withoutSecretsKeepingHint().getClientSecretHint().length());
+    }
+
+    @Test
+    void testShortSecretRevealsFewerCharacters() {
+        // 8-11 characters: below what ClientSecretValidation lets in today, so only legacy values land here.
```
astsiapanay: **[test-coverage]** This comment claims the 8-11 character hint band is "below what ClientSecretValidation lets in today," but `ClientSecretValidation.MIN_LENGTH` is 8, so 8-11 character secrets are exactly what current write validation accepts, not below it — contradicting the PR description itself ("the shortest secrets the write validation below still accepts (8-11 characters)"). A brand-new 10-character secret submitted today will only ever show a 2-character hint; the comment's rationale for the short-hint tier is factually wrong and could mislead someone tuning `HINT_MIN_SECRET_LENGTH`/`FULL_HINT_SECRET_LENGTH` later.

config/src/main/java/com/epam/aidial/core/config/ResourceAuthSettings.java
DmytroZaichenkoDev: Accurate. The split is structural: `redactSecretFields` works on the `ObjectNode` projection of the merged config, not a `ResourceAuthSettings` instance, so it can't reuse the POJO methods.

Collapsing them means round-tripping the node back through `ResourceAuthSettings` — which relies on `client_secret` deserializing while `client_secret_hint` (`READ_ONLY`) doesn't. That swaps an explicit `remove("client_secret")` for an implicit Jackson invariant in the one place whose job is redaction, so I'd rather not do it here.

Filing separately; noted under Known gaps in the description.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/ExternalServiceService.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/service/ExternalServiceService.java
DmytroZaichenkoDev: Fixed

config/src/test/java/com/epam/aidial/core/config/ResourceAuthSettingsTest.java
DmytroZaichenkoDev: Fixed

server/src/main/java/com/epam/aidial/core/server/controller/ExternalServiceManagementController.java
DmytroZaichenkoDev: Fixed

astsiapanay (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1840, opened 2026-08-19
fix: resolve limit and consent paths for deployment names that are not valid URIs

oleksii-donets:
### Description of changes

A model named `anthropic.claude-opus-4-8[1m]` makes `GET /v1/user/usage` and `GET /v1/user/limits` answer **500 for every caller who can access it**, and breaks admission, accounting and consent for that deployment:

```
java.lang.RuntimeException: java.net.URISyntaxException: Illegal character in path at index 25: anthropic.claude-opus-4-8[1m]
  at UrlUtil.decodePath(UrlUtil.java:72)
  at ResourceDescriptorFactory.fromEncoded(ResourceDescriptorFactory.java:34)
  at RateLimiter.getResourceDescription(RateLimiter.java:377)
  at RateLimiter.getLimitAbsolutePath(RateLimiter.java:246)
  at RateLimiter.collectUserStats(RateLimiter.java:170)
```

#### Root cause

The limit record path is assembled from an entity name and handed to `fromEncoded`, which validates every segment by constructing a `java.net.URI`. `[` and `]` are illegal in a URI path (legal only in an authority, for IPv6 literals), and nothing stops such a name: `ConfigPostProcessor` only rejects names holding a separator, and its `^[A-Za-z0-9-_]+$` pattern applies to toolset keys and external-service ids, not models. **A URI syntax check was being applied to a string that is not a URI.**

Worth noting what is *not* wrong with the name: it passes `isValidFilename` — `[` and `]` are not in `INVALID_FILE_NAME_CHARS`. The URI check is the entire defect, so nothing else is relaxed here.

The same helper backs admission (`checkTokenLimit`, `checkRequestLimit`), accounting (`increase`, `updateCostLimits`) and reporting (`getLimitStats`, `getUserStats`), so chat completions, `/anthropic/v1/messages`, `/openai/v1/responses`, the MCP and route surfaces and `/v1/deployments/{id}/limits` all failed too. `/v1/user/limits` and `/v1/user/usage` are the worst case, because `listAccessibleModels` walks *every* accessible model. `ConsentService.getResourceDescription` had the identical defect and failed even earlier, at the consent gate.

#### Why the name is not simply encoded first

`RoleBasedEntity.getName()` has mixed provenance:

| Source | `getName()` |
|---|---|
| File-config models / routes | plain configuration text, e.g. `anthropic.claude-opus-4-8[1m]` |
| Blob-sourced custom applications | an already url-encoded resource url containing `/`, e.g. `applications/<bucket>/my%20app` |

`increase()` has no type guard and is called with `context.getDeployment()`, so applications already hold nested records under `limits/applications/<bucket>/<app>/tokens`. Escaping the name would collapse that into one segment, get it rejected as a filename, and **silently stop all application usage and cost accounting** (that failure is only logged). Dropping the decode step instead (`fromDecoded`) would move the stored records of every escaped name to a new key and desynchronise consent, whose read and write sides currently agree only because both decode.

#### The change

- **`ResourceDescriptorFactory.fromEntityPath`** (new): decodes segments leniently and does not require the path to be a valid URI. Every other check is kept — including the file name charset, because the resulting key still has to be storable by each blob provider, and because such a path can come from a request (a consent deployment id does). Since only that one check is relaxed, any name that resolves today resolves to a **byte-identical** key, so no existing counter is orphaned. `fromEncoded` is unchanged and keeps rejecting non-URI input for client-supplied resource urls.
- **`RateLimiter`** and **`ConsentService`** build their descriptors with it.
- **`ProxyContext.hasNextInterceptor`** decodes the deployment name the same lenient way `decodedSourceDeployment` already does. Without this, completions still returned 500 — the end-to-end test caught it.

#### Tests

- `ResourceDescriptorFactoryTest` — the bracketed name resolves while `fromEncoded` still rejects it; the file name charset is still enforced (`{`, `}`, `"`, encoded or raw); broken percent-encoding kept verbatim; separator, control char, empty segment and >900-byte paths still rejected. Plus a key-stability spot-check over representative shapes (plain, `my%20app`, non-ASCII, single-segment) asserting both factories agree — the general guarantee comes from the two sharing one decode call and differing only in whether it may throw, not from this test.
- `RateLimiterTest` — full cycle for `anthropic.claude-opus-4-8[1m]` (admission, accounting, `/v1/user/usage` and per-deployment reporting), and an application-shaped name proving its record path is unchanged and read/write stay symmetric.
- `LimitApiTest` — end-to-end on a dedicated config: 200 from `/v1/user/limits`, `/v1/user/usage` and `/v1/deployments/{id}/limits`, with counters advancing after a completion.
- `ConsentServiceTest` — bracketed id, and an encoded application id pinning the decoded record path.

#### Verified against a running core

Before: all four request shapes above returned 500 with the exact stack trace from the incident. After: the completion returns 200 and counters advance to 30 tokens / 1 request, with the record stored verbatim at `limits/anthropic.claude-opus-4-8[1m]/tokens` in both blob storage and Redis.

#### Follow-ups deliberately left out

- `UrlUtil.decodePath` wraps `URISyntaxException` in a bare `RuntimeException` while the adjacent fragment/query check throws `IllegalArgumentException`. That inconsistency makes some malformed-path surfaces answer 500 where they mean 400 (e.g. `PublicationController.decodeRule`). Fixing it globally also changes two `catch (IllegalArgumentException) { /* ignore */ }` sites in `ApplicationSchemaService`/`CatalogSchemaService` from "hard error" to "silently skip the resource", so it needs its own change and test sweep.
- `/v1/user/limits` and `/v1/user/usage` answer for the entire config, so a name that *decodes* to something no record path can hold — `gpt%2Fbeta`, `gpt%00x`, or one long enough to exceed the 900-byte limit — would still fail them for every caller. Not user-reachable (names a user controls are `getUrl()` output, singly escaped, and Guava escapes `%`), so this is an admin-config typo rather than the reported bug; per-deployment error isolation belongs in its own change.
- `ConsentService.acceptConsent` does not call `findDeployment`, unlike `buildConsent`, so any authenticated caller can write an arbitrary-key consent record in their own bucket. Pre-existing; adding the check changes API behavior and deserves its own PR.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
+++ b/server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
@@ -373,8 +373,15 @@ public class RateLimiter {
         return getResourceDescription(bucketLocation, path);
     }
 
+    /**
+     * Builds a descriptor for a limit record from an internal relative path, which may carry an entity name.
+     * {@link RoleBasedEntity#getName()} is plain configuration text for a file defined model or route, but an
+     * already url encoded resource url for a custom application - so such a path can never be validated as a
+     * URI (a model named {@code claude-opus-4-8[1m]} is legal config), while it must still be decoded, or a
+     * record written for an application under its encoded name would not be found again.
+     */
     private ResourceDescriptor getResourceDescription(String bucketLocation, String path) {
-        return ResourceDescriptorFactory.fromEncoded(ResourceTypes.LIMIT, bucketLocation, bucketLocation, path);
+        return ResourceDescriptorFactory.fromEntityPath(ResourceTypes.LIMIT, bucketLocation, bucketLocation, path);
     }
 
     private RateLimitResult checkLimit(ProxyContext context, Limit limit, RoleBasedEntity roleBasedEntity) {
--- a/server/src/main/java/com/epam/aidial/core/server/ProxyContext.java
+++ b/server/src/main/java/com/epam/aidial/core/server/ProxyContext.java
@@ -253,7 +253,9 @@ public class ProxyContext {
 
     public boolean hasNextInterceptor() {
         // initial call to the deployment or the interceptor calls another deployment
-        String decodedName = UrlUtil.decodePath(deployment.getName());
+        // a name is url form for an application and plain config text for a model, so it is decoded the same
+        // lenient way as the source deployment above - a model named "claude-opus-4-8[1m]" is not a valid URI
+        String decodedName = UrlUtil.tryDecodePath(deployment.getName());
         if (apiKeyData.getInterceptors() == null || !decodedName.equals(getInitialDeployment())) {
             return !interceptors.isEmpty();
         } else { // make sure if a next interceptor is available from the list
--- a/server/src/main/java/com/epam/aidial/core/server/service/ConsentService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/ConsentService.java
@@ -125,8 +125,12 @@ public class ConsentService {
                 && Boolean.TRUE.equals(deployment.getFeatures().getConsentRequired());
     }
 
+    /**
+     * The id is already decoded - a deployment name from the config, or a path parameter the route decoded -
+     * so it is not a url and cannot be validated as one.
+     */
     private static ResourceDescriptor getResourceDescription(ProxyContext context, String deploymentId) {
         String bucketLocation = BucketBuilder.buildInitiatorBucket(context);
-        return ResourceDescriptorFactory.fromEncoded(ResourceTypes.USER_CONSENT, bucketLocation, bucketLocation, deploymentId);
+        return ResourceDescriptorFactory.fromEntityPath(ResourceTypes.USER_CONSENT, bucketLocation, bucketLocation, deploymentId);
     }
 }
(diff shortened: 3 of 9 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -373,8 +373,15 @@ private ResourceDescriptor getResourceDescription(ProxyContext context, String p
         return getResourceDescription(bucketLocation, path);
     }
 
+    /**
```
astsiapanay: Let's don't add unnecessary javadoc

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -373,8 +373,15 @@ private ResourceDescriptor getResourceDescription(ProxyContext context, String p
...
+     * Builds a descriptor for a limit record from an internal relative path, which may carry an entity name.
+     * {@link RoleBasedEntity#getName()} is plain configuration text for a file defined model or route, but an
+     * already url encoded resource url for a custom application - so such a path can never be validated as a
+     * URI (a model named {@code claude-opus-4-8[1m]} is legal config), while it must still be decoded, or a
+     * record written for an application under its encoded name would not be found again.
+     */
     private ResourceDescriptor getResourceDescription(String bucketLocation, String path) {
```
astsiapanay: Let's rename the method to `toResourceDescriptor`. The current name is wrong.

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1842, opened 2026-08-19
fix: validate empty apiKeyHeader in toolset auth settings #1538

### LINKED ISSUE #1538: [Toolsets] There is no validation for empty `apiKeyHeader`
### Name and Version

dial-core

1. Send request to create toolset with empty `apiKeyHeader`

Image

Request should be rejected with 400 response: "Field `API_KEY_HEADER` is required for API_KEY authentication.'

### What do you see instead?

Toolset is successfully created

### Additional information

_No response_

DmytroZaichenkoDev:
Creating a toolset with `"api_key_header": ""` under `API_KEY` authentication was accepted instead of rejected, storing a toolset that can never authenticate. Field presence in the auth-settings validator was decided with a null check, so an empty or blank string counted as a supplied value.

### Applicable issues

- fixes #1538

### Description of changes

- `BaseAuthSettingsValidator.extractFields` now treats a blank string (and an empty list for `scopes_supported`) as *not provided*, instead of only `null`. A blank `api_key_header` therefore fails the existing required-field check and the request is rejected with `400 Field 'API_KEY_HEADER' is required for API_KEY authentication.`
- Fixed in the shared helper rather than only for `apiKeyHeader`: the same null check let a blank `client_id`/`client_secret` satisfy the OAuth required-field checks.
- This also relaxes the *forbidden*-field direction — a blank value no longer trips it, so a client that leaves `api_key_header: ""` in the payload while switching to `NONE` is no longer rejected. Blank consistently means "not set" in both directions.
- Tests: unit coverage in `ApiKeyAuthSettingsValidatorTest` for empty and whitespace-only headers plus blank forbidden fields, and an end-to-end `ToolSetApiTest.testCreateToolSetWithBlankApiKeyHeader` reproducing the issue against the toolset PUT endpoint. All new tests were verified to fail without the fix.

### Checklist

### DIFF
--- a/credentials/src/main/java/com/epam/aidial/core/credentials/validation/BaseAuthSettingsValidator.java
+++ b/credentials/src/main/java/com/epam/aidial/core/credentials/validation/BaseAuthSettingsValidator.java
@@ -2,8 +2,10 @@ package com.epam.aidial.core.credentials.validation;
 
 import com.epam.aidial.core.config.ResourceAuthSettings;
 import com.epam.aidial.core.credentials.service.ResourceAuthSettingsChangeMode;
+import org.apache.commons.lang3.StringUtils;
 
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Set;
 
 public abstract class BaseAuthSettingsValidator implements AuthSettingsValidator {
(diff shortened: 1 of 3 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1844, opened 2026-08-19
feat: skip API key sign-in when the submitted key is unchanged #1843

### LINKED ISSUE #1843: [Toolsets] Skip API key sign-in when the submitted key is unchanged
`POST /v1/ops/toolset/signin` with `authentication_type: API_KEY` always validates and re-stores the key, even when the submitted key is identical to the one already stored at that credentials level.

**Expected:** if the key matches what is stored, skip the work and return success — no MCP probe against the toolset endpoint, no re-encryption/rewrite of the credentials.

Entry point: `SecuredResourceService.signIn` → `validateApiKey` (the MCP `initialize` + `tools/list` probe) and `ResourceCredentialsService.addResourceCredentials`.

DmytroZaichenkoDev:
`POST /v1/ops/toolset/signin` with `authentication_type: API_KEY` now short-circuits when the submitted key is identical to the one already stored at that credentials level: no MCP `initialize` + `tools/list` probe against the toolset endpoint, and no re-encryption/rewrite of the credentials record. Any other case — no record, a different key or header, a different auth type, or a record that cannot be read — goes through the existing validate-and-store path unchanged.

### Applicable issues

- fixes #1843

### Description of changes

- `SecuredResourceService.signIn` resolves the credentials descriptor first, then checks `isStoredApiKeyUnchanged` before probing or storing; on a match it returns success immediately.
- A record counts as unchanged only when it is an `API_KEY` record at the requested level with the same `apiKey` **and** the same `apiKeyHeader` from the current auth settings. For `USER` level it must additionally match the caller's user id and the request's `offlineUsageConsent`, so a consent flip or a header config change is still persisted.
- The pre-check is best-effort: if the stored record cannot be read (undecryptable, malformed, transient storage error) it falls back to the normal store path, preserving the previous behavior where sign-in simply overwrites such a record. Without this, a malformed blob would have turned every subsequent sign-in into a 400.
- `validateApiKey` split into `validateApiKeyFormat` (blank / illegal control characters) and `probeApiKey` (the MCP probe). Format validation still runs first, before any storage read, so malformed keys are rejected exactly as before.
- Authorization is unaffected — `verifyAccess` runs in the controller before `signIn`, so the skip is reachable only by callers who could already overwrite the credential.
- Tests cover: unchanged key skips both the probe (asserted via request count against a mock server) and the store; changed key, changed header, and flipped `USER`-level consent all still store; an unreadable stored record still stores.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/service/SecuredResourceService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/SecuredResourceService.java
@@ -1,10 +1,12 @@
 package com.epam.aidial.core.server.service;
 
 import com.epam.aidial.core.config.AuthenticationType;
+import com.epam.aidial.core.config.CredentialsLevel;
 import com.epam.aidial.core.config.SecuredResource;
 import com.epam.aidial.core.config.ToolSet;
 import com.epam.aidial.core.credentials.data.credentials.CredentialsDescriptor;
 import com.epam.aidial.core.credentials.data.credentials.CredentialsLocator;
+import com.epam.aidial.core.credentials.data.credentials.ResourceCredentials;
 import com.epam.aidial.core.credentials.data.credentials.ResourceSignInRequest;
 import com.epam.aidial.core.credentials.data.credentials.ResourceSignOutRequest;
 import com.epam.aidial.core.credentials.service.ResourceCredentialsService;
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1851, opened 2026-08-21
chore: derive code style rules in CLAUDE.md from PR review history

oleksii-donets:
Expands the `Code Style` section of `CLAUDE.md` from three lines into conventions mined from ~117 review comments across 32 recent PRs. Every rule was then verified against the codebase, and the ones that asserted conventions the code does not actually follow were dropped rather than shipped.

### Applicable issues

- fixes #

### Description of changes

- **Naming** — verb methods / noun classes; `get*` reserved for accessors that do not change value between calls; `Base` prefix preferred for *new* abstract classes only; map variables must name their contents, accepting both existing house idioms (`apiKeysByCanonicalId`, `prefixToHash`); explicit carve-out never to rename Jackson-bound `config` members, since those names are the public `aidial.config.json` keys.
- **Vert.x** — new section, and the largest gap in the previous text: never block an event-loop thread (route blocking work through `AsyncTaskExecutor.submit`), no locks on the loop, propagate failures through the `Future` chain, treat the hot-reloaded `Config` as immutable and shared, log with SLF4J placeholders and never log credentials or prompt bodies.
- **Scope discipline** — change only what the issue requires; completeness of your own change is in scope while unrelated cleanup is not; report rather than widen; moving code can change behaviour, not just location.
- **Error handling** — no speculative handling in new code, but do not delete existing guards to satisfy that; invalid request input throws, while config load and background reconciliation must not fail wholesale on one bad record; match documented status codes and pass upstream statuses through.
- **Cross-cutting changes** — extending an enum such as `AuthenticationType` means sweeping every site; a new controller mirrors its siblings' rate-limit, MCP handshake, auth and audit steps; cheap guards before expensive work; identical validation on every write path; `tryLockResource` in background loops and `lockResource` on request paths.
- **Tests** — behaviour-bearing classes get unit tests (not DTOs/POJOs); assert against the real code path; test cross-replica behaviour with a foreign `senderPodId` rather than a real multi-pod setup, per `MergedConfigStoreReplicaUpdateTest`.
- **Documentation** — corrects a factual error: `docs/open_api_core.yaml` is generated and auto-committed by CI (`./gradlew replaceSpec` in `.github/workflows/pr.yml`) and must never be hand-edited.

Rules deliberately **excluded** after verification, to keep the file trustworthy:

- "Do not inline method calls as arguments" — not a Java or Google Java Style convention, absent from `checkstyle.xml`, and violated roughly once every 33 lines of main source, including by the reviewer-dictated block now live at `RateLimiter.java:170-172`.
- A prescriptive `<value>By<key>` map-naming mandate — the form appears in ~3% of map declarations; `xToY` is equally established.
- "Logic lives in the class that owns the data" — this argument was declined by PR authors more than once, and acting on it in #1808 moved a cutoff ahead of a Redis lookup, which is why frequently-written records can report zero usage.
- An `@Nullable` mandate — the `config` module has no annotation dependency on its compile classpath, and three different `Nullable` types are already in use across modules.

No Java files are touched. `./gradlew checkstyleMain checkstyleTest` passes.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

CLAUDE.md
```diff
@@ -86,4 +86,116 @@ See `sample/` for examples of both files; `server/src/main/resources/aidial.sett
 
 ## Code Style
 
-Use comments sparingly. Only comment complex code.
+When two rules here conflict, the narrower one wins, and *Scope discipline* beats everything
```
astsiapanay: let's move the section to a separate file like docs/CODE_STYLE.md and reference it from CLAUDE.md with a one-liner, so CLAUDE.md stays short (it's loaded into context on every session).

oleksii-donets: Done

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

astsiapanay (approved):

PR #1859, opened 2026-08-26
feat: support defaultHeaders on deployment and interfaces levels

khshanovskyi:
### Applicable issues

- 

### Description of changes

Adds `defaultHeaders` to models, applications and interceptors — the header counterpart of `defaults`.

- Deployment-level `defaultHeaders` apply to every interface the deployment serves; `interfaces.<type>.defaultHeaders` is laid over them for that interface only (overrides a repeated name, adds new ones, keeps the rest). Names are matched case-insensitively.
- A header the request already carries is never replaced — a client-sent value always wins.
- The headers are materialized on the inbound request, so DIAL Core reads them as if the client had sent them (`X-DIAL-CACHE-POLICY` set this way drives upstream cache pinning, not just what the adapter receives) and they reach the deployment under the same copy rules and exclusions as client headers.
- Headers DIAL Core sets itself always win: they are written after the client headers are copied, with `putHeader`.
- Applied once per request at the head of the chain — with interceptors configured that is the hop to the first interceptor, from where they travel down the chain. An interceptor's own `defaultHeaders` are applied on the hop that calls it and take precedence.

Covers `chat/completions`, `completions`, `embeddings`, `POST /openai/v1/responses`, `POST /anthropic/v1/messages` (+ `count_tokens`) and both interceptor controllers. Documented in `docs/dynamic-settings/{models,applications,interceptors}.md`; `docs/open_api_core.yaml` regenerated.

### Checklist

- [x] All tests pass (`./gradlew test` green, checkstyle clean, OpenAPI spec regenerated)

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/Deployment.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Deployment.java
@@ -139,6 +139,7 @@ public abstract class Deployment extends RoleBasedEntity {
         if (interfaceHeaders.isEmpty()) {
             return defaultHeaders;
         }
+        // TreeMap for its case-insensitive comparator, not for ordering: an interface entry has to override a deployment-level header spelled in another case
         Map<String, String> merged = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         merged.putAll(defaultHeaders);
         merged.putAll(interfaceHeaders);
--- a/config/src/main/java/com/epam/aidial/core/config/Deployment.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Deployment.java
@@ -9,6 +9,7 @@ import lombok.EqualsAndHashCode;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 @Data
 @EqualsAndHashCode(callSuper = true)
@@ -61,6 +62,14 @@ public abstract class Deployment extends RoleBasedEntity {
      * Default parameters are applied if a request doesn't contain them in OpenAI Responses API call.
      */
     private Map<String, Object> responsesDefaults = Map.of();
+    /**
+     * Headers added to a request that carries none under that name, for every interface the deployment
+     * serves. Overlaid per interface by {@code interfaces.<type>.defaultHeaders}, see
+     * {@link #resolveDefaultHeaders}.
+     */
+    @JsonAlias({"defaultHeaders", "default_headers"})
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private Map<String, String> defaultHeaders = Map.of();
     /**
      * List of interceptors to be called for the deployment
      */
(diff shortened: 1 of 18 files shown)


ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33003435438)

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33049000099)

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33049769583)

config/src/main/java/com/epam/aidial/core/config/Deployment.java
```diff
@@ -118,4 +127,21 @@ public boolean hasCatalogSchemaId() {
...
+     * with {@code interfaces.<type>.defaultHeaders} laid over them. Names are compared case-insensitively,
+     * so an interface entry overrides a deployment-level header however either spells it.
+     */
+    public Map<String, String> resolveDefaultHeaders(InterfaceType type) {
+        DeploymentInterface declared = interfaces == null ? null : interfaces.get(type.getValue());
+        Map<String, String> interfaceHeaders = declared == null ? Map.of() : declared.getDefaultHeaders();
+        if (interfaceHeaders.isEmpty()) {
+            return defaultHeaders;
+        }
+        Map<String, String> merged = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
astsiapanay, line 143: Why do we need to keep the order?

config/src/main/java/com/epam/aidial/core/config/Deployment.java
```diff
@@ -118,4 +127,21 @@ public boolean hasCatalogSchemaId() {
...
+
+    /**
+     * The default headers in force for the interface type: the deployment-level {@link #defaultHeaders}
+     * with {@code interfaces.<type>.defaultHeaders} laid over them. Names are compared case-insensitively,
+     * so an interface entry overrides a deployment-level header however either spells it.
+     */
+    public Map<String, String> resolveDefaultHeaders(InterfaceType type) {
+        DeploymentInterface declared = interfaces == null ? null : interfaces.get(type.getValue());
+        Map<String, String> interfaceHeaders = declared == null ? Map.of() : declared.getDefaultHeaders();
+        if (interfaceHeaders.isEmpty()) {
```
astsiapanay, line 139: The application-level `defaultHeaders` apply to every interface the application serves. `interfaces.openaiChatCompletions.defaultHeaders` is laid over them for that interface only: a name it repeats is overridden, a new name is added, and every other application-level header still applies.

astsiapanay: That contradicts to the code

config/src/main/java/com/epam/aidial/core/config/Deployment.java
khshanovskyi: Added the comment `TreeMap for its case-insensitive comparator, not for ordering: an interface entry has to override a deployment-level header spelled in another case`

astsiapanay (approved):

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33053510588)

PR #1864, opened 2026-08-31
feat: support interfaces and baseUrl in upstream config

khshanovskyi:
### Applicable issues

- 

### Description of changes

Model providers that expose several LLM APIs for the same model (Fireworks serves both the OpenAI and the Anthropic API) previously needed two model deployments in Core, because an upstream had one untyped `endpoint`. That field also silently backed the Anthropic Messages API, which is confusing given `endpoint` normally means the OpenAI ChatCompletions endpoint — and the same confusion would repeat for every future API.

This adds a typed `interfaces` map and a `baseUrl` to `upstreams[]`:

- **`upstreams[].interfaces`** keys the same types as the deployment-level map — `openaiChatCompletions`, `openaiResponses`, `openaiEmbeddings`, `anthropicMessages` — but each value carries a full `endpoint` rather than a base url, because an upstream is the provider itself and no ingress path is routed into it.
- **`upstreams[].baseUrl`** completes any declared interface that omits `endpoint`, appending the path from that API's own spec — `/v1/chat/completions`, `/v1/embeddings`, `/v1/responses`, `/v1/messages` — ignoring the `/openai` and `/anthropic` ingress keywords. This is the one difference from deployment-level `interfaces.{type}.base_url`, which is instead concatenated with the request path to Core.
- **`key`, `extraData` and `secretExtraData` are overridable per interface.** An interface-level value wins over the upstream-level one; each field falls back independently, so an interface overriding only `key` still inherits `extraData`. `extraData` and `secretExtraData` are merged into `X-UPSTREAM-EXTRA-DATA` as before.
- **Backward compatible.** `interfaces` is a whitelist, not a default: an interface type absent from the map still falls back to the legacy field serving it — `responsesEndpoint` for `openaiResponses`, `endpoint` for every other type — so existing upstreams keep working unchanged.

Two config-load rules are enforced, and a model violating either is rejected: every declared interface must resolve to a url (its own `endpoint`, or `baseUrl`), and an upstream declaring `interfaces` must declare an `id` — that shape has no `endpoint` to fall back on as its identifier for `X-UPSTREAM-ID` routing and prompt-cache pinning.

Also fixes two pre-existing NPEs that this shape makes easier to hit:

- `X-UPSTREAM-ID` lookup threw `NullPointerException` (500) when any upstream in the list had neither `id` nor `endpoint` — the documented Bedrock/VertexAI/direct-Anthropic shape, where the upstream carries only credentials. It now skips the unaddressable upstream and answers 400 for a genuinely unknown id.
- The prompt-cache entry write threw `NullPointerException: map value can't be null` in Redis when the upstream had no `endpoint`, silently losing every cache pin.

`SecretFieldProcessor` now also descends into map-valued fields, so the new per-interface `key`/`secretExtraData` are encrypted at rest, stripped from admin GET responses, and preserved on a PUT that omits them — the same guarantees the upstream-level secrets already had.

Docs updated in `docs/dynamic-settings/models.md` (new `models.<model_name>.upstreams.interfaces` section) and `sample/aidial.config.json`; `docs/open_api_core.yaml` regenerated via `./gradlew replaceSpec`.

### Checklist

- [x] All tests pass (full `./gradlew test` green; `checkstyleMain`/`checkstyleTest` clean)

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/InterfaceType.java
+++ b/config/src/main/java/com/epam/aidial/core/config/InterfaceType.java
@@ -14,17 +14,40 @@ import java.util.List;
 public enum InterfaceType {
 
     @JsonAlias({"openai_chat_completions"})
-    OPENAI_CHAT_COMPLETIONS("openaiChatCompletions", List.of("prefix.body.tools", "prefix.body.messages")),
+    OPENAI_CHAT_COMPLETIONS(
+            "openaiChatCompletions",
+            "/v1/chat/completions",
+            List.of("prefix.body.tools", "prefix.body.messages")
+    ),
     // an embeddings request carries no prompt prefix to cache, so it contributes no cache keys
     @JsonAlias({"openai_embeddings"})
-    OPENAI_EMBEDDINGS("openaiEmbeddings", List.of()),
+    OPENAI_EMBEDDINGS(
+            "openaiEmbeddings",
+            "/v1/embeddings",
+            List.of()
+    ),
     @JsonAlias({"openai_responses"})
-    OPENAI_RESPONSES("openaiResponses", List.of("prefix.body.tools", "prefix.body.instructions", "prefix.body.input")),
+    OPENAI_RESPONSES(
+            "openaiResponses",
+            "/v1/responses",
+            List.of("prefix.body.tools", "prefix.body.instructions", "prefix.body.input")
+    ),
     @JsonAlias({"anthropic_messages"})
-    ANTHROPIC_MESSAGES("anthropicMessages", List.of("prefix.body.tools", "prefix.body.system", "prefix.body.messages"));
+    ANTHROPIC_MESSAGES(
+            "anthropicMessages",
+            "/v1/messages",
+            List.of("prefix.body.tools", "prefix.body.system", "prefix.body.messages")
+    );
 
     @JsonValue
     private final String value;
+
+    /**
+     * The path this interface's own API spec serves it at, with no DIAL ingress keyword
+     * ({@code /openai}, {@code /anthropic}) in front of it. Appended to {@code Upstream#baseUrl}
+     * to address a provider that hosts the API where its spec says it should be.
+     */
+    private final String apiPath;
 
     /**
      * The node order used to build upstream cache keys, hardcoded per the wire format of this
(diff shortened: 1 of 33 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/upstream/UpstreamRouteProvider.java
```diff
@@ -75,7 +75,9 @@ public UpstreamRoute get(Deployment deployment, CacheBreakpointContext breakpoin
         if (upstreamId != null && !upstreamId.isBlank()) {
             Upstream selected = null;
             for (Upstream upstream : upstreams) {
-                String id = Objects.requireNonNullElse(upstream.getId(), upstream.getEndpoint());
+                // an upstream with neither is unaddressable, not a reason to fail the whole lookup:
```
astsiapanay: Why did you change the line here?

Doesn't the change produce the same result?

khshanovskyi: They're identical except when id and endpoint are both null — requireNonNullElse throws there rather than returning null:                                                                                      
                                                                                                                                                                                                                 
```
  public static <T> T requireNonNullElse(T obj, T defaultObj) {                                                                                                                                                  
      return (obj != null) ? obj : requireNonNull(defaultObj, "defaultObj"); // NPE, not null                                                                                                                    
  }    
```                                                                                                                                                                                                          
                                                                                                                                                                                                                 
  That case is reachable in practice: Bedrock / VertexAI / direct-Anthropic upstreams carry only credentials ("upstreams": [{"key": "..."}]), so they have neither field. Any request sending X-UPSTREAM-ID      
  against such a model hit an NPE → 500 — and it depended on ordering, since the loop breaks on match: an endpoint-less upstream listed before the requested one killed a lookup that would otherwise have       
  succeeded.                                                                                                                                                                                                     
                                                                                                                                                                                                                 
  With the ternary that upstream is simply skipped as unaddressable, and a genuinely unknown id gets the intended 400. Covered by                                                                                
  UpstreamRouteProviderTest.testGet_UpstreamId_UnknownWhenAnUpstreamHasNeitherIdNorEndpoint.

khshanovskyi: @astsiapanay rollbacked

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

Issue #1863, opened 2026-08-28
[DIAL 2.0] Introduce tenant-rooted storage layout (addressing only)


DmytroZaichenkoDev:
## Why

Resources live under three unrelated roots — `Users/<id>/`, `public/`, `Keys/<id>/` — each with its own access rules, and there is no level above them that represents an organization. That blocks multi-tenancy and makes access hard to reason about.

## What this covers

The first step only: a single tenant-rooted storage layout, and the mapping from API paths onto it. Today's deployment becomes that same layout with one default tenant, so nothing changes for users.

- one tree with a tenant root, holding personal and shared spaces
- the API path to storage path mapping, in one place
- existing buckets and URLs keep working unchanged

This is the first phase of the storage, authorization and multi-tenancy work. Later phases build on it.

## Out of scope

- authorization changes — the current model keeps deciding access
- data migration
- tenants, teams and channels as user-facing features
- complex resource packaging

## Acceptance

- no change to any client-visible URL or bucket token
- existing behavior and tests unchanged
- new layout behind a flag, off by default

PR #1866, opened 2026-08-31
feat: add tenant-rooted storage layout behind a setting #1863

DmytroZaichenkoDev:
Part 1 of 2 for #1863 (addressing only). Part 2 is #1868.

## What

Adds a tenant-rooted storage layout, off by default, and routes physical path composition through it.

```
Users/u1/files/documents/notes.txt   →  .org/<tenant>/.users/u1/.files/documents/notes.txt
public/files/policies/risk.pdf       →  .org/<tenant>/.files/policies/risk.pdf
Keys/<project>/files/report.csv      →  .org/<tenant>/.keys/<project>/.files/report.csv
platform/files/x                     →  .files/x        (platform scope is the root)
```

- `TenantLayoutTransform` — the conversion, total and reversible in both directions.
- `StorageLayout` + `LegacyStorageLayout` — `ResourceDescriptor` composed the bucket location and the resource-type folder inline in **two** places (`getAbsoluteFilePath` and `resolveByPath`); both now go through one layout.
- `TenantRootedStorageLayout` + `StorageLayouts` — the new layout and the holder for the active one.
- `AiDial` selects it at start-up from a new settings section:

```json
"storageLayout": {
  "tenantRooted": false,
  "defaultTenant": "default"
}
```

## Invariant

**Off by default, so physical paths are unchanged.** `ResourceDescriptorPathTest` pins the current paths; `TenantRootedLayoutApiTest` drives the resource API with the flag **on** through the real stack — round trip, listing, deletion, and the blob landing under `.org/test-tenant/…` in a reserved type folder — which is the only way to show the descriptor, the cache and the blob store agree on the same paths.

## Notes for review

- The active layout is **process-wide mutable state**. `ResourceDescriptor` is constructed in ~63 files and carries no configuration, so there is nowhere else to put it without touching every construction site. Open to alternatives.
- `AiDial` sets the layout **unconditionally** on every start, including the legacy case, so it is always determined by settings rather than inherited from a previous `start()` in the same JVM.
- Applications and toolsets keep **separate** folders (`.applications`, `.toolsets`) rather than both mapping to `.deployments` as the design sketch shows: merging two groups into one folder is not invertible, and a working reverse transform is required for migration.
- `storageLayout` is a new top-level section rather than a field on `storage` — `Storage` is jclouds provider configuration, this is a logical concern — and deliberately **static** settings, not hot-reloaded config.
- `defaultTenant` defaults to `"default"`; that string becomes a physical path segment, so it is worth an explicit decision.
- `Users/` and `Keys/` prefixes are repeated in `TenantLayoutTransform` because `storage` cannot depend on `server`'s `BucketBuilder`.

## Not included

No migration sits behind this flag. Enabling it against populated storage re-addresses everything and the existing data is not reachable at the new paths — the README says so. Three things must land before it can be flipped anywhere real: re-encryption of path-bound secrets (see #1868 for the first half), a decision on invitation ids, which embed the physical path, and a measurement of the existing path-length distribution against the 900-byte ceiling, which tenant-rooting eats into.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
⚠️ Dependency review workflow failed - results may be outdated. [Check logs](https://github.com/epam/ai-dial-core/actions/runs/33386957204)

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

storage/src/main/java/com/epam/aidial/core/storage/resource/TenantLayoutTransform.java
```diff
@@ -0,0 +1,124 @@
...
+ * stays in {@link ResourceDescriptor#getAbsoluteFilePath()}.
+ *
+ * <p>The conversion is total and reversible in both directions, so a migrated path can always be mapped
+ * back to its origin.
+ *
+ * <p>Legacy locations are produced by the server-side bucket builder; the prefixes are repeated here
+ * because this module cannot depend on it.
+ */
+@UtilityClass
+public class TenantLayoutTransform {
```
astsiapanay, line 19: TenantLayoutTransformer?

README.md
```diff
@@ -232,6 +232,8 @@ without putting the raw address in the log.
 | storage.bucket              |     -      |    No    | Blob storage bucket.                                                                                                                                                                               |
 | storage.overrides.*         |     -      |    No    | Key-value pairs to override storage settings. `*` might be any specific blob storage setting to be overridden. Refer to [examples](#temporary-credentials-1) in the sections below.                |
 | storage.createBucket        |   false    |    No    | Indicates whether bucket should be created on start-up.                                                                                                                                            |
+| storageLayout.tenantRooted  |   false    |    No    | Places every bucket under a tenant root instead of storing it at the top level. Changing it on a populated deployment re-addresses existing data and requires migration; leave disabled otherwise.  |
```
astsiapanay, line 235: If the storage layout is a part of the module `storage` does it make sense to make the setting a part of the storage config?

`storage.layout.tenantRooted`

storage/src/main/java/com/epam/aidial/core/storage/resource/StorageLayouts.java
```diff
@@ -0,0 +1,22 @@
...
+    private static volatile StorageLayout active = LegacyStorageLayout.INSTANCE;
+
+    private StorageLayouts() {
+    }
+
+    public static StorageLayout resolveActive() {
+        return active;
+    }
+
+    public static void useLayout(StorageLayout layout) {
```
astsiapanay, line 19: Can we switch between layouts in runtime?

storage/src/main/java/com/epam/aidial/core/storage/resource/TenantLayoutTransform.java
DmytroZaichenkoDev: Fixed: 
https://github.com/epam/ai-dial-core/pull/1889/changes/74fe4b033c533c3014d7a9cdfd8cf7e83312870b

README.md
DmytroZaichenkoDev: Fixed:
https://github.com/epam/ai-dial-core/pull/1889/changes/41f441cab56242bf31d88b2168e901786f18f188

storage/src/main/java/com/epam/aidial/core/storage/resource/StorageLayouts.java
DmytroZaichenkoDev: No, by design: the layout is fixed at startup before any resource I/O, because a runtime flip re-addresses live data out from under the Redis cache, the dirty write-behind queue, and per-resource locks (all keyed by physical path). The setter exists so the comparison tests can boot two instances in one JVM. Per-bucket migration in P2 will come as a composite layout consulted per resolution, not a runtime switch.

astsiapanay: That's a dangerous method by nature. I'm thinking how we can make us safe. Can we come up with more robust mechanism so anyone couldn't change the layout in runtime?

PR #1869, opened 2026-08-31
test: drive the resource API under the tenant-rooted layout #1863

DmytroZaichenkoDev:
Part 4 of 4 for #1863. Stack: layout seam → tenant layout → secret AAD → **this**.

## What

An integration test that enables `storageLayout.tenantRooted` and drives the resource API through the real stack.

- conversation round trip (`PUT` then `GET` returns the same body)
- listing and deletion
- the blob is physically stored under `.org/test-tenant/…` in a reserved type folder

## Why

Parts 1 to 3 verify the conversion in isolation. Nothing had yet run the descriptor, the Redis cache and the blob store together with the flag on — which is the only way to show they agree on the same paths.

## Notes for review

- The layout is process-wide, so the test restores the legacy layout in `@AfterEach`. Any test going through `AiDial.start()` also resets it, since the layout is set unconditionally there.
- Found while writing this: `MAX_PATH_SIZE` (900 bytes) is checked against the **physical** path, and tenant-rooting lengthens every path by roughly 15 bytes for a default-tenant user resource. Resources within that margin of the ceiling today would migrate but then fail validation. The ceiling is correct as written — it guards the blob store's key limit — so this is a migration constraint to measure before cutover, not something to relax here.


DmytroZaichenkoDev:
Folded into #1866, so the end-to-end test sits with the change it proves.

PR #1871, opened 2026-08-31
feat: support baseUrl and mode in deployment interfaces config

khshanovskyi:
### Description of changes

Deployment-level counterpart of #1864: a `baseUrl` on the deployment, and a `mode` on each `interfaces` entry.

- **`baseUrl` on the deployment** — the adapter root shared by every `interfaces` entry that declares no `base_url` of its own. In most pass-through cases one adapter serves every interface and only the path differs (`/openai/deployments/{name}/chat/completions`, `/openai/v1/responses`, `/anthropic/v1/messages`), so the supported interfaces can simply be listed with no url each.
- **`interfaces.<type>.base_url` is now optional** and overrides the deployment-level one for that interface only. A base url is resolved as: interface `base_url` → deployment `baseUrl` → legacy `endpoint`/`responsesEndpoint`.
- **`interfaces` stays the whitelist.** A deployment `baseUrl` on its own declares no interface, and an entry with no base url at either level is neither served (503) nor advertised in `/v1/deployments`. An interface mapped to `null` reads exactly like an absent one.
- **`mode`: `passthrough` | `translator`** — makes explicit whether the interface is forwarded as it arrived or handed to a translator. Absent means `passthrough` and is omitted from serialized config, so every config written before this keeps its behaviour and is not rewritten on round-trip.
- **A `translator` interface does not charge its tokens to limits.** The translator calls Core back to have the completion served, and that inner request already carries the usage to the same initiator bucket, so charging both would count it twice.

```json
{
  "models": {
    "openai-gpt-5.4-mini": {
      "overrideName": "gpt-5.4-mini",
      "type": "chat",
      "baseUrl": "http://dial-openai-adapter/",
      "interfaces": {
        "openaiChatCompletions": { "mode": "passthrough" },
        "openaiResponses": null,
        "anthropicMessages": {
          "mode": "passthrough",
          "baseUrl": "http://dial-bedrock-adapter/"
        }
      },
      "upstreams": []
    }
  }
}
```

The root-level `translators` config and the per-interface `translator` field are **not** part of this PR. When they land, translator routing slots into the resolution chain between the deployment-`baseUrl` fallback and the legacy fallback, and `isInterfaceDeclared` needs the same branch so translated interfaces stay visible in the deployment listing.

Docs updated: `docs/dynamic-settings/models.md` (resolution order, `baseUrl`, `mode`, example), plus the inherited `baseUrl` in `applications.md` and `interceptors.md`.

### Checklist

- [x] All tests pass — `DeploymentEndpointUtilTest` (+5 cases), `DeploymentTest` (+5), `DeploymentPostControllerTest` (+3: translator not charged, passthrough charged with and without an explicit mode), full `./gradlew test` suite, checkstyle clean

### DIFF
--- a/config/src/main/java/com/epam/aidial/core/config/Application.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Application.java
@@ -179,6 +179,7 @@ public class Application extends Deployment {
         this.setInvalid(source.getInvalid());
         this.setName(source.getName());
         this.setEndpoint(source.getEndpoint());
+        this.setBaseUrl(source.getBaseUrl());
         this.setInterfaces(source.getInterfaces());
         this.setDisplayName(source.getDisplayName());
         this.setDisplayVersion(source.getDisplayVersion());
--- a/config/src/main/java/com/epam/aidial/core/config/Deployment.java
+++ b/config/src/main/java/com/epam/aidial/core/config/Deployment.java
@@ -18,6 +18,14 @@ public abstract class Deployment extends RoleBasedEntity {
     private String endpoint;
     private String responsesEndpoint;
     /**
+     * Source (adapter) root shared by every {@link #interfaces} entry declaring no {@code base_url} of its
+     * own — in the common case one adapter serves all of them and only the ingress path differs. It stands
+     * for no interface by itself: {@code interfaces} alone says which ones the deployment serves.
+     */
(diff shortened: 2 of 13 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1879, opened 2026-09-01
fix: handle case when role limits are not set at all #1880

### LINKED ISSUE #1880: NullPointerException when a role's limits are not set
### Description

A `Role` can be configured in `aidial.config.json` without a `limits` section at all (i.e. `role.getLimits()` returns `null` rather than an empty map). Two code paths assume `limits` is always a non-null map and dereference it directly, causing an NPE:

1. **`ConfigPostProcessor.processRoles`** (`server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java`) iterates `role.getLimits().entrySet()` unconditionally while logging role limits during config load/reload. If `getLimits()` is `null`, this throws NPE during config reload.
2. **`RateLimiter.getLimit`** (`server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java`) calls `role.getLimits().get(name)` inside a `Optional.map`, which likewise NPEs when the role has no limits configured, breaking rate limiting for that role.

### Steps to reproduce

1. Add a role to `aidial.config.json` without a `limits` field, e.g.:
   ```json
   "roles": {
     "default": {}
   }
   ```
2. Trigger a config load/reload, or send a request as a user assigned to that role.

### Expected behavior

Config loads successfully and rate limiting falls back to the default limit when a role has no limits configured, instead of throwing an NPE.

### Actual behavior

`NullPointerException` is thrown, at least during config reload (`ConfigPostProcessor.processRoles`), and potentially during request-time rate limit checks (`RateLimiter.getLimit`).

### Suggested fix

Null-check `role.getLimits()` before iterating/dereferencing in both `ConfigPostProcessor.processRoles` and `RateLimiter.getLimit`.

KirylKurnosenka:
Fixes a NullPointerException that occurs when a role is configured without any rate limits.

### Applicable issues

- fixes #1880

### Description of changes

- `ConfigPostProcessor.processRoles`: guard against `role.getLimits()` returning `null` before iterating over it during config load/reload.
- `RateLimiter.getLimit`: chain `Role::getLimits` through `Optional.map` instead of dereferencing it directly, so a role without limits falls back to the default limit instead of throwing NPE.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
+++ b/server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
@@ -390,8 +390,11 @@ public final class ConfigPostProcessor {
             Role role = entry.getValue();
             role.setName(name);
             log.debug("Start loading role `{}`", role.getName());
-            for (Map.Entry<String, Limit> limitEntry : role.getLimits().entrySet()) {
-                log.debug("Loading {} for deployment `{}`", limitEntry.getValue(), limitEntry.getKey());
+            Map<String, Limit> roleLimits = role.getLimits();
+            if (roleLimits != null && !roleLimits.isEmpty()) {
+                for (Map.Entry<String, Limit> limitEntry : roleLimits.entrySet()) {
+                    log.debug("Loading {} for deployment `{}`", limitEntry.getValue(), limitEntry.getKey());
+                }
             }
             log.debug("End loading role `{}`", role.getName());
         }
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay:
The issue should have the label `bug`

KirylKurnosenka:
> The issue should have the label `bug`

Created the corresponding bug #1880, change pr title prefix from `chore` to `fix`

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1882, opened 2026-09-01
feat: support translators for deployment interfaces

khshanovskyi:
### Description of changes

Follow-up to #1871, which added `interfaces.<type>.mode`. This adds the translator half: a deployment can serve an API it does not speak itself by handing the request to a service that converts it.

- **`translators` at the root of the config** — a map of `{in, out, baseUrl}`. `in` is the API a client calls DIAL Core on, `out` the API it is converted to, and `baseUrl` the translator service. The ingress path is appended to `baseUrl` exactly as it is to a deployment's own base URL.
- **`interfaces.<type>.translator`** — the name of a `translators` entry, or a definition written inline as `{out, baseUrl}`. Whichever form was written is what a config round-trip gives back, so a named reference is never frozen into a copy of what it resolved to.
- **Either a base URL or a translator, never both.** `mode` decides which, and a config where the two disagree is rejected on load rather than resolved by precedence — routing and limits both read `mode`, so a base URL winning for routing while `mode` exempted the request from limits would be an uncharged pass-through request. A translated interface therefore falls back to nothing: not its own `base_url`, not the deployment `baseUrl`, not the legacy `endpoint`.
- **Limits.** A translated request is checked against every limit and counts towards none. Tokens, cost and `requestHour`/`requestDay` are all carried by the call the translator makes back to Core, on the same initiator, so one client call is counted exactly once. It is not a way around a quota: an exhausted caller is still rejected on the translated interface.
- **`X-DIAL-DEPLOYMENT-ID`** is now sent for a translated request on the interfaces that did not already send it (Anthropic Messages always did), so the translator knows which deployment to call back for after `overrideName` rewriting.

Validation rejects only config that no reload can fix — a translator declared alongside a `base_url`, `mode: translator` with no translator, an `in` that is not the interface it is declared under, a translator converting an interface to itself (its output would arrive back on the interface it came from, and nothing bounds that at runtime), and an `out` the deployment does not serve pass-through. A reference that merely resolves to no URL — an unknown name, or an entry with no `baseUrl` — leaves that one interface unserved instead, exactly as it does for an application or interceptor.

Named references are linked on every config load, so editing a `translators` entry reaches every deployment referencing it. Three places assemble a `Config` field by field and had to carry the registry with them — `MergedConfigStore.rebuild`, `MergedConfigStore.shallowClone` and `ConfigManifestSupport.newScratch` — and model resurrection re-links, since it rebuilds a model from its stored payload.

Not included: `translators` is file-config-only, with no `ResourceTypes` entry, so unlike models, applications, interceptors and routes it cannot be managed through the config API or migrated. Worth its own change.

Docs: new `docs/dynamic-settings/translators.md`, plus the translator field and resolution order in `models.md` and the limits note in `roles.md`.

### Checklist

- [x] All tests pass — full `./gradlew test` suite, checkstyle clean. New coverage: `TranslatorTest` (registry and both reference shapes), `DeploymentEndpointUtilTest` (translator routing, no fallback, linking), `ConfigPostProcessorTest` (each validation rule, linking, unknown reference tolerated), `MergedConfigStoreTest` and `MergedConfigStorePartialUpdateTest` (registry survives the merge, resurrection re-links), `RateLimiterTest` and `DeploymentPostControllerTest` (translated requests checked but not counted)


ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA 9a0f02806f6fdc8d858c413a9707bdc875576f67.</blockquote>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

config/src/main/java/com/epam/aidial/core/config/InterfaceType.java
```diff
@@ -68,11 +68,23 @@ public enum InterfaceType {
...
+     * that only a newer Core, or another component, understands.
+     */
+    public static InterfaceType find(String value) {
         for (InterfaceType type : values()) {
             if (type.value.equals(value)) {
                 return type;
             }
         }
-        throw new IllegalArgumentException("Unknown interface type: " + value);
+        return null;
```
astsiapanay, line 90: Why do we return nullable?

config/src/main/java/com/epam/aidial/core/config/Translator.java
```diff
@@ -0,0 +1,56 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Translator {
+
+    /**
+     * The interface type the translator accepts, as {@link InterfaceType#getValue()}. Absent only inline,
+     * where the entry the definition sits under is what it converts from; a registry entry is declared under
+     * no interface, so it has to say. Types this Core does not know are kept as written, the same way an
+     * {@code interfaces} key is.
+     */
+    private final String in;
```
astsiapanay: Should it be a type of InterfaceType?

config/src/main/java/com/epam/aidial/core/config/Translator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * The interface type it converts to, which the deployment has to serve itself: the translator calls
+     * Core back on that interface to have the completion served.
+     */
+    private final String out;
```
astsiapanay: The same

config/src/main/java/com/epam/aidial/core/config/databind/TranslatorRefDeserializer.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Reads {@code interfaces.<type>.translator} in either shape it is written: a string naming a
+ * {@code translators} entry, or an object defining one inline.
+ */
+public class TranslatorRefDeserializer extends JsonDeserializer<TranslatorRef> {
+
+    @Override
+    public TranslatorRef deserialize(JsonParser p, DeserializationContext ctx) throws IOException {
+        if (p.getCurrentToken() == JsonToken.VALUE_STRING) {
+            return TranslatorRef.named(p.getValueAsString());
```
astsiapanay, line 21: Why do we need this mechanism?

docs/dynamic-settings/translators.md
```diff
@@ -0,0 +1,95 @@
...
+* `baseUrl`: **Required.** The root URL of the translator service. Each request is forwarded to `baseUrl` + **the exact ingress path it was received on**, exactly as a deployment's own base URL is. A trailing slash is normalized.
+
+**Example**
+
+```json
+"translators": {
+    "anthropicMessagesToOpenaiResponses": {
+        "in": "anthropicMessages",
+        "out": "openaiResponses",
+        "baseUrl": "http://dial-bedrock-translator/to-responses"
```
astsiapanay, line 28: Is it base URL or a complete endpoint?
It looks like a complete URL in the example.

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -324,6 +337,139 @@ public static void validateUpstreamInterfaces(Model model, List<ValidationWarnin
...
+     * declared under no interface, so {@code in} is the only thing tying it to one. {@code out} and
+     * {@code baseUrl} need no check here — {@link Translator} rejects an entry missing either as it is read.
+     */
+    private static void validateTranslators(Config config) {
+        for (Map.Entry<String, Translator> entry : config.getTranslators().entrySet()) {
+            Translator translator = entry.getValue();
+            if (translator == null) {
+                throw new IllegalStateException("Translator '" + entry.getKey() + "' is empty");
+            }
+            if (translator.getIn() == null) {
```
astsiapanay, line 351: Do we need to validate the field `out`?

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -324,6 +337,139 @@ public static void validateUpstreamInterfaces(Model model, List<ValidationWarnin
...
+            }
+        }
+    }
+
+    /**
+     * Validates a model's {@code interfaces}. An entry is served either by a base url or by a translator,
+     * never by both and never by neither, and {@code mode} is what says which — routing and limits both
+     * read it, so a config where it disagrees with the fields around it is rejected rather than resolved.
+     */
+    public static void validateDeploymentInterfaces(Model model, List<ValidationWarning> warnings) {
```
astsiapanay, line 399: Can we validate a case when one translator recursively invokes another forming a cycle?

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
```diff
@@ -210,6 +211,16 @@ protected Future<Void> collectTokenUsage(Buffer responseBody) {
         return trackDeploymentStats(context.getDeployment().getName(), ownUsage, true);
     }
 
+    /**
+     * Checks the initiator's limits before the request is forwarded. A translated request is checked against
+     * every one of them, but counts towards none: the call the translator makes back to Core is what carries
+     * both the usage and the request slot, so counting here as well would charge one client call twice.
+     */
+    protected Future<RateLimitResult> checkLimits(Deployment deployment) {
+        boolean translated = DeploymentEndpointUtil.resolveMode(deployment, interfaceType()) == InterfaceMode.TRANSLATOR;
```
astsiapanay: Why do we calculate limits in the mode translator?

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
```diff
@@ -88,6 +88,16 @@ public Future<Void> increase(
...
+        return limit(context, roleBasedEntity, true);
+    }
+
+    /**
+     * @param countRequest whether the call consumes a slot of the caller's {@code requestHour}/{@code requestDay}
+     *                     quota. False for a call another one accounts for: a translated request is checked
+     *                     against every limit, but the call the translator makes back to Core is the one
+     *                     counted, so a single client call never spends two slots.
+     */
+    public Future<RateLimitResult> limit(ProxyContext context, RoleBasedEntity roleBasedEntity, boolean countRequest) {
```
astsiapanay: Why do we always  calculate token limits but conditionally requests only?

server/src/main/java/com/epam/aidial/core/server/util/DeploymentEndpointUtil.java
```diff
@@ -52,6 +54,10 @@ public boolean isInterfaceDeclared(Deployment deployment, InterfaceType type) {
         if (resolveInterfaceBaseUrl(deployment, type) != null) {
             return true;
         }
+        // a translated interface is served by its translator alone, so one with none linked serves nothing
+        if (resolveMode(deployment, type) == InterfaceMode.TRANSLATOR) {
```
astsiapanay: That might be surprising to see a such exception. The method name is isInterfaceDeclared. So why do we have the exception here?

config/src/main/java/com/epam/aidial/core/config/databind/TranslatorRefDeserializer.java
astsiapanay: It's a polymorphic type.

config/src/main/java/com/epam/aidial/core/config/InterfaceType.java
astsiapanay: Let's annotate with @Nullable

khshanovskyi: Done — annotated `@Nullable` (needed `jakarta.annotation:jakarta.annotation-api` on the `config` module, which had no nullability dependency).

On why it is nullable at all: it existed so an untyped `out` could be skipped rather than rejected. Now that `in`/`out` are typed that reason is gone, and the one remaining caller resolves an `interfaces` map key, whose key space really is open.

config/src/main/java/com/epam/aidial/core/config/Translator.java
khshanovskyi: Done — `in` and `out` are `InterfaceType` now.

Worth flagging the trade-off: an unknown value fails the config parse instead of parsing inert, where an `interfaces` **map key** stays tolerated (unchanged). I think that is right for these two — they are matched against real interfaces rather than just carried — and it is what makes the `out` check total, see the cycle thread. `FileConfigStore` catches the parse failure and keeps the previous config live. Happy to revert to `String` if you would rather keep the tolerance.

config/src/main/java/com/epam/aidial/core/config/Translator.java
khshanovskyi: Same — `out` is `InterfaceType`.

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
khshanovskyi: The case is caught, but the message hid it: it reported *"The model does not serve 'openaiResponses'"*, which reads like a missing interface rather than a loop. Split into its own warning now — an `out` landing on a translated interface says the callback would be handed to a translator again.

More importantly, typing `out` closed the hole that made a longer cycle reachable at all: `validateTranslatorOutput` used to bail out on `InterfaceType.find(out) == null`, so an unknown `out` skipped the check entirely.

No graph walk is needed after that. `out` must land on a **pass-through** interface, so a two-hop chain cannot be configured, so no multi-interface cycle can form — your example is rejected at both ends. The one-interface cycle (`in == out`) was already checked. Test added for the two-interface case.

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
khshanovskyi: Because not being counted is not the same as not being checked — a translated interface should not be a way around an exhausted quota. Every limit is still checked before the request is forwarded; only the counting is skipped, since the translator's callback carries the usage and the request slot.

Reworked how that is decided, though: `mode == TRANSLATOR` was spelled out at two call sites. It now lives on `InterfaceMode.isChargedToInitiator()`, so a mode added later answers the question once at its declaration, and the pre-flight check and the post-response charge share one predicate instead of drifting apart.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
khshanovskyi: Because tokens are never counted here, on any path — they are not known until the response is read, so `increase` adds them afterwards, while a request slot is spent at admission. So there is one flag rather than two: a call another one accounts for opts out of tokens and cost by not calling `increase`, and out of the request slot by passing `false` here. Both are still checked either way.

That was invisible in the code; stated it on `limit` now.

server/src/main/java/com/epam/aidial/core/server/util/DeploymentEndpointUtil.java
khshanovskyi: Agreed, and it was a duplicate as well — `resolveLegacyEndpoint` already owns that rule (a translated interface never falls back to `endpoint`/`responsesEndpoint`). `isInterfaceDeclared` now asks that resolver instead of reimplementing the exception, so no mode is special-cased in it. Behaviour is identical and the existing tests pass unmodified.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/controller/BaseDeploymentPostController.java
khshanovskyi: Correction to my reply above — the design changed in `ca5b7524`, and what I wrote no longer describes the code.

I argued the translated call had to be checked so it would not be a way around a quota. That was wrong: it is checked, just one hop later. The translator calls Core back on the `out` interface, and that callback is an ordinary pass-through request checked against the **initiator's** limits (`BucketBuilder.buildInitiatorBucket` resolves from the user/project the per-request key carries). A caller over quota gets a 429 there.

So the translated call is a routing hop to the translator service, not a request the caller is accountable for, and it now touches the limits not at all — neither read nor written. Only real requests are checked and charged, and a translated interface still cannot be used to exceed a quota.

Trade-off, for the record: the 429 now reaches the client through the translator rather than straight from Core, and the translator pays a round-trip for a request that was going to be refused. Documented in `translators.md`.

server/src/main/java/com/epam/aidial/core/server/limiter/RateLimiter.java
khshanovskyi: Same correction — `countRequest` is gone as of `ca5b7524`.

Your question was the right one to ask. The asymmetry existed because the outer call still checked while not counting, which needed a flag for requests but not for tokens. Now the outer call does nothing at all, so `RateLimiter.limit` is a single method again with no flag threading down to `RequestRateLimit.check` — that part of the file is back to what it was before this PR.

The mode-level predicate is now `InterfaceMode.isSubjectToLimits()`, and both the pre-flight check and the post-response charge skip on the same terms, so they cannot drift apart.

astsiapanay (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

Issue #1870, opened 2026-08-31
[DIAL 2.0] Verify behaviour is unchanged on the new storage layout


DmytroZaichenkoDev:
Follows #1863, which added the tenant-rooted storage layout behind a setting that is off by default.

## Why

Before any stored data moves to the new layout, we need to be able to *show* that behaviour is unchanged — not assert it. Today there is no way to answer "does the platform behave identically on the new layout?" other than by reading the code and hoping.

Getting migrated permissions wrong is the risk that matters most here, and it is the one least visible by inspection.

## What this covers

Instruments only. Nothing is migrated and nothing is switched on.

- **Replay and compare** — record real API traffic, replay it against both layouts, and diff the responses.
- **Access decisions** — check that the same subject gets the same answer for the same resource under both layouts, across the ways access can be granted today (own resources, sharing, public rules, per-request grants, admin).
- **Verify a copied bucket** — confirm everything arrived, and that content which was encrypted before is still readable afterwards.

## Out of scope

- moving any data
- the migration tool itself
- any change to authorization or to the API surface

## Acceptance

- replaying recorded traffic against both layouts shows no differences, or every difference is deliberate and written down
- a bucket copied to the new layout can be checked as complete and readable
- all of it runs without enabling the new layout for anyone

PR #1889, opened 2026-09-02
fix: map system buckets and keep durable artifacts stable under the tenant-rooted layout #1863

DmytroZaichenkoDev:
Stacked on #1866.

`TenantLayoutTransform` handled four bucket locations; the codebase builds descriptors against four more, and each threw under the tenant-rooted layout. These paths are composed off the request path, so the failures took out a subsystem rather than a request:

| Location | What breaks |
|---|---|
| `deployment_cost_stats/` | per-request cost accounting, on the `/v1/deployments` path |
| `background_jobs/` | background job scanning |
| `response_mappings/` | the response id index |
| `api_key_data/` | per-request keys, so every application caller |

They now map to a reserved `.system/` segment at the root, above any tenant — that is what preserves today's behaviour, since the job scheduler scans its bucket whole and all four are keyed by globally unique ids. Whether this state should become tenant-scoped belongs to the phase that turns multi-tenancy on.

The locations move to `ResourceDescriptor` so a layout can enumerate them, and their callers now reference those constants; a second literal next to the caller is how these came to be missed. An unregistered location still throws rather than falling back silently.

**Durable artifacts must not follow the physical path.** The seam made `getAbsoluteFilePath()` layout-dependent, which breaks anything derived from it and then stored:

- **Invitation ids** encoded the physical path, so `getInvitationResource` could not parse a pre-flip id back and share-by-link returned 404.
- **Encryption AAD** at three sites — `SecretFieldProcessor` (every `@EncryptedField` on config-managed entities), `UserExternalServiceService` (user-authored client secrets), `BackgroundJobService` (the per-request key inside a job record). Ciphertext keeps the AAD it was sealed with; after a layout flip the AAD re-derives differently and decryption tag-mismatches. For config entities the failure is silent: the entity drops out of the merged config as invalid.

All four now use `getStableFilePath()`, which is byte-identical to the legacy physical path under any layout. `SecretFieldProcessorLayoutTest` proves it with real AES-GCM: a secret encrypted under the legacy layout decrypts after the flip, and ciphertext bound to the tenant-shaped physical path does not.

**Public sub-buckets.** `ApplicationService` keys a public function app's deployment folder as a synthesized sub-bucket, `public/deployments/<id>/`, which the transform rejected — every deploy of a public app threw under the tenant-rooted layout. Public sub-buckets now keep their legacy suffix under the tenant root; dotted segments are rejected so the mapping stays reversible, since dotted names there are reserved for principal branches and resource-type folders. The private (`Users/<sub>/deployments/<id>/`) and review (`Users/<sub>/publications/<pub>/deployments/<id>/`) variants already worked via multi-segment principal ids and are pinned by `ApplicationDeploymentLayoutTest` alongside the public one.

**Reserved names are now enforced.** The tenant-rooted tree gives `.org`/`.system`/`.users`/`.keys` and dotted type folders one namespace, and full-path parsing depends on the reserved names staying unambiguous. No `ResourceTypes` group takes one today, but nothing kept a future one off them; the transform now rejects the collision, and the every-type round-trip test doubles as the build-time tripwire.

Found by the P1 verification instruments (#1870) and a follow-up audit of physical-path consumers.


storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceDescriptor.java
```diff
@@ -106,8 +132,24 @@ public String getDecodedUrl() {
...
     public String getAbsoluteFilePath() {
+        return getStoragePrefix(StorageLayouts.resolveActive()) + getPathWithinType();
+    }
+
+    /**
+     * The path {@link #getAbsoluteFilePath()} produces under the legacy layout, whichever layout is active.
+     * Anything durable derived from a path — an identifier handed to a user, an encryption AAD — must use
+     * this: a physical path is free to change when the layout does, and the stored artifact is not.
+     */
+    public String getStableFilePath() {
```
astsiapanay, line 143: getStatbleFilePath is confusing. It had better to assign explicitly name, e.g. getLegacyFilePath.

storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceDescriptor.java
```diff
@@ -23,6 +24,31 @@ public class ResourceDescriptor {
...
 
+    /**
+     * Prefixes of the two principal bucket-location shapes, {@code Users/<sub>/} and {@code Keys/<project>/}.
+     * The server-side bucket builder formats locations with them, and a storage layout recognizes principal
+     * locations by them — one set of literals for both, or the two drift.
+     */
+    public static final String USERS_LOCATION_PREFIX = "Users" + PATH_SEPARATOR;
+    public static final String KEYS_LOCATION_PREFIX = "Keys" + PATH_SEPARATOR;
+
+    public static final String DEPLOYMENT_COST_STATS_BUCKET = "deployment_cost_stats";
```
astsiapanay, line 35: Perhaps we want to reorganize the constants into a new enum SystemResourceRegistry

storage/src/main/java/com/epam/aidial/core/storage/resource/StorageLayout.java
```diff
@@ -9,5 +9,5 @@ public interface StorageLayout {
 
     String resolveLocationPrefix(String bucketLocation);
 
-    String resolveTypeFolder(String group);
+    String resolveTypeFolder(String typeGroup);
```
astsiapanay, line 12: What's group's type? We changed the name and semantics?

storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceDescriptor.java
DmytroZaichenkoDev: Fixed

storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceDescriptor.java
DmytroZaichenkoDev: Fixed

storage/src/main/java/com/epam/aidial/core/storage/resource/StorageLayout.java
DmytroZaichenkoDev: No semantics change — the parameter was always ResourceType.group() (the storage folder name, as opposed to urlSegment()); the rename just aligned the interface with the transformer's vocabulary, and it's now documented on the method.

astsiapanay (approved):

PR #1893, opened 2026-09-02
test: verify a bucket migrated to the tenant-rooted layout #1870

DmytroZaichenkoDev:
Stacked on #1892.

Copies a legacy tree into the tenant-rooted layout, checks everything arrived, then **boots a tenant-rooted core onto the copy** with an empty cache and reads it all back.

The only instrument that crosses layouts. The other two start empty and each run reads only its own writes, so neither can see a failure that happens only to data written under one layout and read under the other — which is the whole of what P2 is about to do.

What it compares:

| | |
|---|---|
| Inventory | every source path, mapped through the transform, exists at the destination; nothing extra |
| Content | the bytes |
| Metadata | `content-encoding`, `content-type`, `etag`, `author`, `created_at` |
| `public/rules/rules` | **by name**, not as a row in a count — its absence fails open, and every object count would still add up |
| Read-back | what each resource returned before the move vs after |

Read-back is a before/after comparison rather than an assertion of 200: the question is whether migration changed anything, not whether every fixture happens to be readable.

`BucketCopier` is a stand-in for the migrator P2 will build, not that migrator. It exists so the verifier has something to verify, and so the questions the verifier asks are settled before the real one lands. Swap it when P2 arrives.

**What building it turned up:** a byte-perfect copy still produces an unreadable bucket. Whether an object is compressed is recorded beside it rather than inside it, along with its etag, author and creation time — so a copy that keeps every byte and drops the metadata reads back as a parse error. Only some resources are affected, since small ones are stored uncompressed, which is what makes it easy to miss.

Inventory and checksums both passed while that was true. That is why the verifier compares metadata as well, and it is the thing P2's migrator most needs to get right.

### DIFF
--- a/server/src/test/java/com/epam/aidial/core/server/layout/BucketCopier.java
+++ b/server/src/test/java/com/epam/aidial/core/server/layout/BucketCopier.java
@@ -1,0 +1,115 @@
+package com.epam.aidial.core.server.layout;
+
+import com.epam.aidial.core.storage.resource.TenantLayoutTransform;
+import lombok.SneakyThrows;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.UserDefinedFileAttributeView;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
+/**
+ * Copies a blob tree from the legacy layout into the tenant-rooted one.
+ *
+ * <p>A stand-in for the migrator P2 will build, not the migrator itself: it copies bytes and rewrites paths,
+ * with none of the throughput, resumability or re-encryption a real one needs. It exists so the verifier has
+ * something to verify before the real one lands, and so the questions the verifier asks are settled first.
+ */
+@Slf4j
+@UtilityClass
+public class BucketCopier {
+
+    /**
+     * Splits a physical path into the bucket location and what follows, by finding the resource-type folder.
+     * The two halves are what {@link TenantLayoutTransform} converts, and a path is only made of those two
+     * plus the resource path within the type.
+     */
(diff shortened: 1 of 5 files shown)


server/src/test/java/com/epam/aidial/core/server/layout/BucketCopier.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Copies a blob tree from the legacy layout into the tenant-rooted one.
+ *
+ * <p>A stand-in for the migrator P2 will build, not the migrator itself: it copies bytes and rewrites paths,
+ * with none of the throughput, resumability or re-encryption a real one needs. It exists so the verifier has
+ * something to verify before the real one lands, and so the questions the verifier asks are settled first.
+ */
+@Slf4j
+@UtilityClass
+public class BucketCopier {
```
astsiapanay, line 26: So we verify the logic against local FS. That won't be true for blob storage providers.

astsiapanay, line 26: Any plans to check all scenarios agains AWS, GCP and Azure?

astsiapanay (approved):

server/src/test/java/com/epam/aidial/core/server/layout/BucketCopier.java
DmytroZaichenkoDev: Deliberately FS-only in P1 — CI has no cloud credentials, and a provider matrix here would mostly test jclouds. The verifier is designed to be re-pointed: P2's real migrator replaces the BucketCopier stand-in and runs this verification against the actual provider during each staged cutover.

DmytroZaichenkoDev: Deliberately FS-only in P1 — CI has no cloud credentials, and a provider matrix here would mostly test jclouds. The verifier is designed to be re-pointed: P2's real migrator replaces the BucketCopier stand-in and runs this verification against the actual provider during each staged cutover.

PR #1897, opened 2026-09-03
chore: optimize RuleService#getCachedRules, align RULES TTL #1896

### LINKED ISSUE #1896: Optimize RuleService#getCachedRules: drop metadata pre-check, align RULES TTL with APPLICATION
## Context

`RuleService#getCachedRules()` (`server/src/main/java/com/epam/aidial/core/server/service/RuleService.java:127-143`) is called on every `listRules`, `getAllowedPublicResources`, and `filterForbidden` invocation. Today it does **two** sequential calls:

1. `resources.getResourceMetadata(PUBLIC_RULES)` — a metadata-only lookup, purely to check whether the cached `updatedAt` is stale.
2. Only if stale: `resources.getResourceWithMetadata(PUBLIC_RULES, EtagHeader.ANY)` — metadata + full body.

Both calls hit the same underlying path in `ResourceService` (`storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java`): Redis first (`redisGet`), falling back to blob storage (S3/Azure/GCS) — plus a distributed lock — only on a Redis miss (`getResourceBytesWithMetadata`, lines 554-578). On a Redis *hit*, `getResourceWithMetadata` is just a single `redisGet` call with no extra locking or blob access — exactly as cheap as `getResourceMetadata`, except it also returns the body for free. So doing the metadata-only pre-check first is pure overhead on the (normal) hit path: it turns one Redis round trip into two, for no benefit, since the body would need fetching almost every time anyway once TTL is fixed (see below).

The fix has two parts:

**A. Remove the metadata pre-check and the local `AtomicReference` cache entirely.** Since every call already has to hit Redis via `getResourceWithMetadata(...)` to get an up-to-date body, keeping a separate `updatedAt`-keyed local cache buys nothing — it's another comparison/CAS around a Redis call that's already fast (once part B lands) and already returns the current body. Simplify `getCachedRules()` to just call `resources.getResourceWithMetadata(...)` and decode the result directly, every time. This collapses two Redis round trips (metadata pre-check + conditional body fetch) into one, and removes the now-pointless caching layer.

**B. Raise the RULES Redis TTL to match APPLICATION (30 days).** Currently `RULES` has a 5-minute TTL (`ResourceTypes.java:15`), so the Redis entry keeps expiring under normal traffic, forcing the expensive blob-storage fallback + lock on a regular basis. Public rules are small, so there's no storage-cost reason to evict them quickly — `APPLICATION` and several other resource types already use a 30-day TTL (`ResourceTypes.java:18,22,27-34`) for exactly this reason. This keeps the rules resource resident in Redis, so the single `getResourceWithMetadata` call in `getCachedRules()` almost always hits the fast Redis path instead of the blob-storage fallback.

## Changes

1. **`server/src/main/java/com/epam/aidial/core/server/service/RuleService.java`**
   - Remove the `cachedRules` field (line 41, plus its javadoc comment at lines 38-40) and the now-unused `AtomicReference` import (line 25).
   - Rewrite `getCachedRules()` (lines 127-143) to a single direct fetch-and-decode, no caching:
     ```java
     private Map<String, List<Rule>> getCachedRules() {
         Pair<ResourceItemMetadata, String> resource = resources.getResourceWithMetadata(PUBLIC_RULES, EtagHeader.ANY);
         return decodeRules(resource == null ? null : resource.getValue());
     }
     ```
   - `ResourceItemMetadata` is no longer referenced anywhere in the class once this lands, so that import (line 11) is removed too; `Pair` (line 17) stays in use for the local variable. `EtagHeader` (line 15) is unchanged.

2. **`storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceTypes.java:15`**
   Change:
   ```java
   RULES("rules", true, TimeUnit.MINUTES.toMillis(5)),
   ```
   to:
   ```java
   RULES("rules", true, TimeUnit.DAYS.toMillis(30)),
   ```
   This is the enum default consumed by `AiDial.getResourceSettings()` (`server/src/main/java/com/epam/aidial/core/server/AiDial.java:382-391`) unless overridden by `resources.resourceTypesExpiration.RULES` in settings (no such override exists in the shipped defaults, so this enum change takes effect directly).

3. **`README.md`** — add a `resourceTypesExpiration.RULES` row next to the existing `APPLICATION`/`TOOL_SET` rows (around line 266-267), documenting the new 30-day default, since RULES currently has no row at all in that table:
   ```
   | resources.resourceTypesExpiration.RULES |      | 30 days  | Define expiration time for public rules                                                            |
   ```

No test updates are required — no existing test asserts the RULES TTL value or mocks `getResourceMetadata`/`getResourceWithMetadata` call counts for `RuleService` (confirmed via repo-wide search), and `PublicationApiTest.listRules()` only checks functional behavior (which is preserved: `getCachedRules()` always returns rules decoded from the current Redis/blob-storage state, just without an intermediate local cache).

## Verification

- `./gradlew :storage:compileJava :server:compileJava` — confirm compilation.
- `./gradlew :server:test --tests "com.epam.aidial.core.server.PublicationApiTest"` — confirm existing rules behavior (including `listRules`) still passes.
- `./gradlew checkstyleMain` — confirm style compliance for the rewritten method.

astsiapanay:
Optimizes `RuleService#getCachedRules()` to make a single Redis round trip instead of two, and raises the `RULES` resource TTL to match `APPLICATION` so it stays resident in Redis.

### Applicable issues

- fixes #1896

### Description of changes

- `RuleService#getCachedRules()` previously called `getResourceMetadata` to check staleness and then, only if changed, `getResourceWithMetadata` to fetch the body — two sequential Redis calls on every invocation. Now it calls `getResourceWithMetadata` directly and unconditionally, and uses the returned `updatedAt` against a local `AtomicReference` cache to decide whether to re-decode the JSON body, avoiding the extra metadata pre-check while still skipping unnecessary re-parsing.
- Raised `ResourceTypes.RULES`'s Redis TTL from 5 minutes to 30 days, matching `APPLICATION` and several other small, infrequently-changing resource types. Public rules are small, so there's no storage-cost reason to evict them quickly — keeping them resident in Redis avoids the more expensive blob-storage fallback that a short TTL forces under normal traffic.
- Documented the new `resources.resourceTypesExpiration.RULES` default in `README.md`.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,7 @@ without putting the raw address in the log.
 | resources.resourceTypesExpiration.PROMPT |      |  5 mins  | Define expiration time for prompts                                                                 |
 | resources.resourceTypesExpiration.APPLICATION |      | 30 days  | Define expiration time for applications                                                            |
 | resources.resourceTypesExpiration.TOOL_SET |      | 30 days  | Define expiration time for toolsets                                                                |
+| resources.resourceTypesExpiration.RULES |      | 30 days  | Define expiration time for public rules                                                            |
 
 </details>
 
--- a/server/src/main/java/com/epam/aidial/core/server/service/RuleService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/RuleService.java
@@ -125,15 +125,13 @@ public class RuleService {
     }
 
     private Map<String, List<Rule>> getCachedRules() {
-        ResourceItemMetadata meta = resources.getResourceMetadata(PUBLIC_RULES);
-        long key = (meta == null) ? Long.MIN_VALUE : meta.getUpdatedAt();
+        Pair<ResourceItemMetadata, String> resource = resources.getResourceWithMetadata(PUBLIC_RULES, EtagHeader.ANY);
+        long key = (resource == null) ? Long.MIN_VALUE : resource.getKey().getUpdatedAt();
         Pair<Long, Map<String, List<Rule>>> current = cachedRules.get();
 
         if (current == null || current.getKey() != key) {
-            Pair<ResourceItemMetadata, String> resource = resources.getResourceWithMetadata(PUBLIC_RULES, EtagHeader.ANY);
-            Pair<Long, Map<String, List<Rule>>> next = (resource == null)
-                    ? Pair.of(Long.MIN_VALUE, decodeRules(null))
-                    : Pair.of(resource.getKey().getUpdatedAt(), decodeRules(resource.getValue()));
+            Map<String, List<Rule>> decoded = decodeRules(resource == null ? null : resource.getValue());
+            Pair<Long, Map<String, List<Rule>>> next = Pair.of(key, decoded);
 
             cachedRules.compareAndSet(current, next);
             current = next;
--- a/storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceTypes.java
+++ b/storage/src/main/java/com/epam/aidial/core/storage/resource/ResourceTypes.java
@@ -12,7 +12,7 @@ public enum ResourceTypes implements ResourceType {
     SHARED_BY_ME("shared_by_me", true, TimeUnit.MINUTES.toMillis(5)),
     INVITATION("invitations", true, TimeUnit.MINUTES.toMillis(5)),
     PUBLICATION("publications", true, TimeUnit.MINUTES.toMillis(5)),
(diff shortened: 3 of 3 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

PR #1906, opened 2026-09-04
fix: skill recreate after delete fails with 412 #1904

### LINKED ISSUE #1904: [Assets-Skills] Deleting a skill and re-creating one with the same name fails with 412 "Resource already exists"
### Name and Version

Core: latest

1. Create a skill by PUT-ing its folder to `/v2/skills/{bucket}/{path}` with header `If-None-Match: *` (create-only semantics), e.g. path `monitor-ci`.
2. Delete it: `DELETE /v2/skills/{bucket}/{path}`.
3. Immediately PUT a new skill to the exact same path (`monitor-ci`) again, with `If-None-Match: *`.

After a successful delete, the path is free again and a new skill can be created under the same name.

### What do you see instead?

The create request is rejected with `412 Precondition Failed`, body:
```
Resource already exists
```
even though the skill was already deleted and does not appear in any skill listing/metadata endpoint. The `If-None-Match: *` check appears to still see the deleted resource — stale metadata/index not fully cleaned up by the delete — blocking re-use of the name.

### Additional information

Also noting the error body is a raw plain-text string rather than Core's usual JSON error envelope (`{ error: { message } }`), which is a secondary inconsistency worth fixing alongside this.

### Confidential information

I confirm that do not share any confidential information

astsiapanay:
After deleting a skill and immediately re-creating one at the same path with `If-None-Match: *`, the create-only PUT was incorrectly rejected with `412 Resource already exists`, even though the skill no longer appeared in any listing.

### Applicable issues

- fixes #1904

### Description of changes

- `ComplexResourceService.delete(...)` only tombstones the marker resource (`state = "deleting"`) and defers actual removal to the sweep service's grace period. Every other read path in `ComplexResourceService` (`getMarker`, `classify`, `listChildren`, ...) treats a `deleting` marker as absent via the `isActive(...)` helper, but the etag lookup used for `If-Match`/`If-None-Match` precondition validation (`readAggregateEtag`) did not, so it still returned a non-null etag for a deleted-but-not-yet-swept resource, tripping the `If-None-Match: *` check.
- Fixed `readAggregateEtag` to return `null` for an inactive (tombstoned) marker, consistent with the rest of the class.
- Added a regression test (`SkillResourceApiTest#testRecreateAfterDeleteWithIfNoneMatch`) covering create → delete → recreate with `If-None-Match: *`.

The secondary inconsistency noted in the issue (412 body being raw plain text instead of the JSON error envelope) is existing, consistent behavior across all `/v1` and `/v2` resource endpoints, not specific to this bug, so it was left out of scope for this fix.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/resource/ComplexResourceService.java
@@ -970,7 +970,7 @@ public class ComplexResourceService {
 
     private String readAggregateEtag(ResourceDescriptor marker) {
         FolderResourceMarker document = readMarker(marker, false);
-        return document == null ? null : document.getEtag();
+        return isActive(document) ? document.getEtag() : null;
     }
 
     private FolderResourceMarker readMarker(ResourceDescriptor marker, boolean lock) {
--- a/server/src/test/java/com/epam/aidial/core/server/SkillResourceApiTest.java
+++ b/server/src/test/java/com/epam/aidial/core/server/SkillResourceApiTest.java
@@ -195,6 +195,16 @@ public class SkillResourceApiTest extends ResourceBaseTest {
     }
 
     @Test
+    void testRecreateAfterDeleteWithIfNoneMatch() {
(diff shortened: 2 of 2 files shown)


Oleksii-Klimov (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1909, opened 2026-09-04
fix: resolve translator references at request time instead of materializing them into deployments

khshanovskyi:
### Description of changes

- Follow-up to #1882: a deployment's named `translator` reference was materialized into `TranslatorRef.definition` at config load, so every path putting a deployment into a live config had to relink it (`ConfigPostProcessor.linkTranslators`, `MergedConfigStore` resurrection, `validateSingleModel`), and an edit to a `translators` entry only reached referencing deployments after a relink or reload.
- `TranslatorRef` is now an immutable name-or-inline reference with a single `resolve(Map<String, Translator>)` method; the name is looked up in the registry on each request, mirroring how interceptors are resolved by name via `config.interceptors`. `linkTranslators` and all relink call sites are deleted.
- Only the URL-answering methods of `DeploymentEndpointUtil` (`resolveServingEndpoint`, `resolveRequestUri`, `resolveResponsesBaseUri`) take the translators registry, threaded from `context.getConfig().getTranslators()` at the request-path controllers. `resolveMode` and `isInterfaceDeclared` are unchanged, so the listing chain (`FeaturesData`, `ApplicationData`, model/deployment controllers) keeps its signatures.
- Behavior change: an interface with `mode: translator` is advertised by carrying a translator reference, whether or not the name resolves — an unregistered name is now listed in `/v1/deployments` and answers `503` on request, instead of being hidden from listings (consistent with how an application whose backend is missing still lists). `docs/dynamic-settings/translators.md` updated, including the edit-propagation rule: references are resolved on each request, never frozen into a copy.
- As a side effect, resource-backed deployments (blob models, custom applications) referencing a translator by name now resolve it correctly without depending on any relink pass.

### Checklist

- [x] All tests pass (`:config:test`, `:server:test`, `checkstyleMain`, `checkstyleTest`)


ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA bc4903e2dd44cdbec4cd4cb00353ba5527761f51.</blockquote>
None

astsiapanay (approved):

PR #1911, opened 2026-09-04
fix: config file migrate result reports kind/resourceUrl/key instead of an ambiguous id

### LINKED ISSUE #1784: Slice D — Config-file migration endpoint POST /v1/admin/config/file/migrate

KirylKurnosenka:
## Summary
- `ConfigFileMigrateResult.id` used to mean different things depending on entity type and status: the canonical platform-bucket id for most entries, but the raw schema `$id` for skipped/failed schemas.
- Replaces it with `kind` (entity kind), `resourceUrl` (the percent-encoded admin API path, relative with no leading `/v1/`, at which the resulting entity is reachable — matching the `ResourceDescriptor#getUrl()` convention; percent-encoding matters because a managed type's short name comes straight from an arbitrary file-config key, e.g. a model id containing `[`/`]`, and isn't restricted the way `ENTITY_NAME_PATTERN` restricts a PUT/DELETE name) and `key` (the entity's identifier in the source file config — a short name, or a schema's `$id`).
- A schema whose blob name can never be derived still reports which schema failed via `key`, without a canonical id or a fake blob path.
- Keys never report `key` (a key entity's file-side identity is its raw secret, which must never be echoed back).

Closes #1784

## Test plan
- [x] `ConfigFileMigrateApiTest` extended to assert `kind`/`resourceUrl`/`key` across migrated/skipped/failed results for models, interceptors, keys, and schemas
- [x] New test covering a schema whose blob name can never be derived (`key` reported, no `resourceUrl`)
- [x] New test covering idempotent schema migration reporting the matched blob name instead of the raw `$id`
- [x] New test covering a model with a URI-unsafe name (`anthropic.claude-opus-4-8[1m]`): asserts the reported `resourceUrl` is percent-encoded (`models/platform/anthropic.claude-opus-4-8%5B1m%5D`) and that a GET against that exact path returns 200
- [x] Keys never report `key`, asserted across migrated/skipped/failed branches
- [x] `./gradlew :server:test --tests "com.epam.aidial.core.server.ConfigFileMigrateApiTest"` passes locally (18 tests)


ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA 6ed06fd87fb7cfe610ee6d5e0d120c9d56526655.</blockquote>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

docs/open_api_core.yaml
```diff
@@ -14863,12 +14863,16 @@ components:
...
-          type: string
         reason:
           type: string
         status:
           $ref: "#/components/schemas/ConfigFileMigrateStatus"
+        blobName:
+          type: string
+        fileId:
+          type: string
+        kind:
```
astsiapanay, line 14920: Add description

docs/open_api_core.yaml
```diff
@@ -14863,12 +14863,16 @@ components:
     ConfigFileMigrateResult:
       type: object
       properties:
-        id:
-          type: string
         reason:
           type: string
         status:
           $ref: "#/components/schemas/ConfigFileMigrateStatus"
+        blobName:
```
astsiapanay: How much blobName is different to fileId?

docs/open_api_core.yaml
```diff
@@ -14863,12 +14863,16 @@ components:
...
+          type: string
+        fileId:
```
astsiapanay: That may confuse a user more than we have one ID in the response.

docs/open_api_core.yaml
KirylKurnosenka: Added: `kind`, `blobName`, and `fileId` now each have a `description` in 7ee30ccc.

docs/open_api_core.yaml
KirylKurnosenka: fileId is where the entity came from (its name/short-name in the source file config, or a schema's $id); blobName is where it ended up (or would end up) as a platform-bucket resource. They're independent: a schema whose blob name can never be derived still reports fileId with no blobName, and a key never reports fileId at all (see next comment). Added descriptions to spell this out in 7ee30ccc.

docs/open_api_core.yaml
KirylKurnosenka: Understandable concern, but the old single id was actually the bug this PR fixes: it meant a canonical blob path for most results, but the raw file $id for skipped/failed schemas - callers couldn't tell which kind of identifier they were looking at. Splitting it makes each field mean exactly one thing, always. Added descriptions in 7ee30ccc to make that explicit; happy to rename the fields if blobName/fileId still isn't clear enough.

ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA 6fe1ac689615bde00c684e54a14aa8254b786071.</blockquote>
None

docs/open_api_core.yaml
```diff
@@ -14863,12 +14863,19 @@ components:
     ConfigFileMigrateResult:
       type: object
       properties:
-        id:
-          type: string
         reason:
           type: string
         status:
           $ref: "#/components/schemas/ConfigFileMigrateStatus"
+        blobName:
```
astsiapanay: The name doesn't seem to reflect what data it actually holds.
You say `Canonical id of the resulting platform-bucket blob (e.g. models/platform/gpt-4). `
When I read the description I understand blobName is gpt-4.

docs/open_api_core.yaml
astsiapanay: I would keep id as is. But I add a new field `source`: `config` or `storage`. That was intention to distinguish where the entity came from

KirylKurnosenka: Renamed:
1. `blobName` -> `resourceUrl` (which is url encoded)
2. `fileId` -> `key`

docs/open_api_core.yaml
KirylKurnosenka: Renamed:
1. `blobName` -> `resourceUrl` (which is url encoded)
2. `fileId` -> `key`

ai-dial-actions:
<h1>Dependency Review</h1>
<blockquote>⚠️: No snapshots were found for the head SHA 3f7e6cf192fb92cdaeea455c057d567034024365.</blockquote>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

docs/open_api_core.yaml
KirylKurnosenka: Since PR is approved, resolving comments on my own

docs/open_api_core.yaml
KirylKurnosenka: Since PR is approved, resolving comments on my own

docs/open_api_core.yaml
KirylKurnosenka: Since PR is approved, resolving comments on my own

docs/open_api_core.yaml
KirylKurnosenka: Since PR is approved, resolving comments on my own

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1915, opened 2026-09-07
feat: support defaults on interfaces level

khshanovskyi:
Adds `defaults` to an `interfaces.<type>` entry, so a deployment can default body parameters per interface instead of only per API through the deployment-level `defaults`/`responsesDefaults`. An entry declaring `defaults` states the whole set for that interface and replaces the deployment-level one; the deployment-level fields stay as the fallback they have always been, which keeps every existing config behaving exactly as before.

### Description of changes

- Add `defaults` to `DeploymentInterface`, resolved by `Deployment#resolveDefaults(InterfaceType)`. The two levels are **not** merged: the deployment-level sets hold parameters of one API each, so they cannot be laid under an interface of another. The priority is interface entry first, deployment level second.
- Keep the deployment-level fields serving the interfaces they were written for: `defaults` for `openaiChatCompletions` and `openaiEmbeddings`, `responsesDefaults` for `openaiResponses`. Neither holds Anthropic parameters, so neither reaches `anthropicMessages`, which defaults from `interfaces.anthropicMessages.defaults` or nowhere.
- Take the resolved map in `RequestObject#applyDefaults` instead of the whole `Deployment`, so which defaults are in force is decided once, by interface, rather than by each request shape. `MessagesApiRequest` now applies them instead of skipping them.
- Resolve the fronted deployment's defaults and default headers under the interface the client called: an intercepted `embeddings` request no longer resolves them as chat completions, while the interceptor's own settings keep resolving under the interface it is addressed on.
- Document the field in `models.md`, `applications.md` and `interceptors.md`, and add it to the `DeploymentInterface` OpenAPI schema.

For the configuration below:

```json
"openai-gpt-5.4-mini": {
    "type": "chat",
    "baseUrl": "http://dial-openai-adapter/",
    "defaults": {
        "temperature": 1,
        "custom_fields": { "configuration": "foo.baz" }
    },
    "interfaces": {
        "openaiChatCompletions": { "mode": "passthrough" },
        "openaiResponses": { "mode": "passthrough" },
        "anthropicMessages": {
            "mode": "passthrough",
            "defaults": { "temperature": 0.5 }
        }
    }
}
```

| Request | Effective `defaults` | Source |
|---|---|---|
| `POST /openai/deployments/{name}/chat/completions`, `POST /openai/deployments/{name}/embeddings` | `{"temperature": 1, "custom_fields": {"configuration": "foo.baz"}}` | deployment-level `defaults`, since neither entry declares its own |
| `POST /openai/v1/responses` | `{}` | `responsesDefaults` is absent and the entry declares none; the deployment-level `defaults` does not serve this interface |
| `POST /anthropic/v1/messages` | `{"temperature": 0.5}` | the entry declares `defaults`, so that is the whole set |

### Checklist

- [x] All tests pass (2642 tests, 0 failures, 0 errors; checkstyle clean)

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/function/enhancement/ApplyDefaultDeploymentSettingsFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/enhancement/ApplyDefaultDeploymentSettingsFn.java
@@ -17,25 +17,15 @@ import java.util.Map;
 public class ApplyDefaultDeploymentSettingsFn extends BaseRequestFunction<RequestObject> {
 
     /**
-     * The interface the deployment this hop calls is addressed on. On the hop that calls an interceptor
-     * that is the interceptor's own interface, which is not necessarily the one the client called.
+     * The interface the client called. Every deployment this hop applies settings for resolves them under
+     * it - the interceptor it calls as much as the deployment the chain fronts - whatever interface the
+     * hop is routed on.
      */
-    private final InterfaceType servingInterface;
-    /**
-     * The interface the client called. The deployment the interceptor chain fronts resolves its settings
-     * under this one however the hops in front of it are addressed.
-     */
-    private final InterfaceType requestedInterface;
+    private final InterfaceType interfaceType;
 
     public ApplyDefaultDeploymentSettingsFn(Proxy proxy, ProxyContext context, InterfaceType interfaceType) {
-        this(proxy, context, interfaceType, interfaceType);
-    }
-
-    public ApplyDefaultDeploymentSettingsFn(Proxy proxy, ProxyContext context,
-                                            InterfaceType servingInterface, InterfaceType requestedInterface) {
         super(proxy, context);
-        this.servingInterface = servingInterface;
-        this.requestedInterface = requestedInterface;
+        this.interfaceType = interfaceType;
     }
 
     @Override
@@ -49,7 +39,7 @@ public class ApplyDefaultDeploymentSettingsFn extends BaseRequestFunction<Reques
         request.clearInterceptorSettings();
         Deployment deployment = context.getDeployment();
         if (deployment instanceof Interceptor) {
-            applyDefaults(request, deployment, servingInterface);
+            applyDefaults(request, deployment);
         }
     }
 
@@ -60,7 +50,7 @@ public class ApplyDefaultDeploymentSettingsFn extends BaseRequestFunction<Reques
                 String deploymentId = context.getInitialDeployment();
                 deployment = proxy.getDeploymentService().findDeployment(context, deploymentId);
             }
-            applyDefaults(request, deployment, requestedInterface);
+            applyDefaults(request, deployment);
         }
     }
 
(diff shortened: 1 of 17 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/function/enhancement/ApplyDefaultDeploymentSettingsFn.java
```diff
@@ -16,11 +16,26 @@
 @Slf4j
 public class ApplyDefaultDeploymentSettingsFn extends BaseRequestFunction<RequestObject> {
 
-    private final InterfaceType interfaceType;
+    /**
+     * The interface the deployment this hop calls is addressed on. On the hop that calls an interceptor
+     * that is the interceptor's own interface, which is not necessarily the one the client called.
+     */
+    private final InterfaceType servingInterface;
```
astsiapanay: Why do we handle defaults differently for interceptors and upstreams?

astsiapanay: Should it be the same interface? servingInterface?

khshanovskyi: No upstream is involved here — the two fields are the two deployments this function resolves settings for on a single hop: `context.getDeployment()` (the interceptor) in `applyInterceptorDefaults`, and the deployment the chain fronts, looked up via `getInitialDeployment()`, in `applyDeploymentDefaults`. They are separate config entries with separate `interfaces` maps.

They only diverge for embeddings. `buildUri` sends an intercepted embeddings request over the interceptor's chat completions interface, so on that hop the interceptor is addressed on chat completions while the fronted model is serving embeddings. Using `servingInterface` for both would resolve the model's defaults and `defaultHeaders` as chat completions and skip its `interfaces.openaiEmbeddings.defaults` — the bug this PR fixes, pinned by `testDefaults_AtFirstInterceptorOfAnEmbeddingsRequest`. Using `requestedInterface` for both would instead stop applying an interceptor's own `interfaces.openaiChatCompletions.defaultHeaders` to embeddings traffic, which is current behaviour.

Happy to take the second option if you prefer a single field — it is simpler, at the cost of that behaviour change.

Separately, this caught an inaccurate claim in the javadoc one file over: I had written that chat completions is the only interface an interceptor may declare, but `ConfigPostProcessor` validates `interfaces` only for models, not for interceptors or applications. Corrected in 6b51a5ae to say it is the only one `buildUri` addresses it on.

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/function/enhancement/ApplyDefaultDeploymentSettingsFn.java
khshanovskyi: You are right - collapsed to a single `interfaceType` in 86452336.

The distinction only arose because `buildUri` routes an intercepted `/embeddings` request over the interceptor's `openaiChatCompletions` base url. Resolving both deployments under *that* is what breaks: the model's `interfaces.openaiEmbeddings.defaults` gets skipped, which is the bug this PR fixes. Resolving both under the interface the client called is correct, and costs only this: an interceptor's `interfaces.openaiChatCompletions.defaultHeaders` no longer applies to embeddings traffic, its deployment-level `defaultHeaders` still does.

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1916, opened 2026-09-07
fix: ignore the client-supplied model name for model deployments

adubovik:
## The issue

The Azure-style chat completions endpoint names the deployment in the **path**:

```
POST /openai/deployments/{deployment_id}/chat/completions
```

so the `model` field in the body has no routing role at all. DIAL Core never read it — and never rewrote it either (unless the deployment declared an `overrideName`), so it was forwarded to the model adapter verbatim.

That leaves the choice of the actual model in the hands of the client. Given this config:

```json
{
  "models": {
    "gpt-4o-mini": {
      "endpoint": "http://openai-adapter/openai/deployments/gpt-4o-mini/chat/completions",
      "pricing": { "unit": "token", "prompt": "0.15", "completion": "0.60" }
    }
  }
}
```

a user with access to the cheap deployment could send:

```
POST /openai/deployments/gpt-4o-mini/chat/completions
{
  "model": "gpt-4o",
  "messages": [...]
}
```

DIAL Core resolves the deployment `gpt-4o-mini` from the path — access control, rate limits, token limits and **pricing** are all taken from that deployment — and then hands the adapter a body asking for `gpt-4o`. Any adapter that honours the body's `model` (most OpenAI-compatible ones do, and the DIAL OpenAI adapter passes it through to Azure/OpenAI) serves the more expensive model. The user is charged the cheap deployment's price, the analytics attribute the spend to `gpt-4o-mini`, and the same trick bypasses per-model access rules — a user who is not entitled to `gpt-4o` at all can reach it through any deployment they can call.

The principle this breaks: **model routing is the domain of DIAL Core**. The deployment config is the only mapping between DIAL deployment ids and upstream model names; the client must not be able to influence it.

## The fix

`EnhanceDeploymentRequestFn` now rewrites the body's `model` for every **model** deployment:

* `overrideName` when the model declares one (unchanged behaviour, still wins);
* the DIAL deployment id otherwise (new).

Applications, interceptors and toolsets are untouched — they keep passing the body through, and keep applying their own `overrideName`, exactly as before. The rewrite happens in the existing enhancement chain, so the other request transformations still take place around it: `defaults` are applied first and can no longer set `model` for a model deployment (Core owns that field now), and `overrideName` is applied as part of the same step.

## API: before and after

Config used in the examples:

```json
{
  "models": {
    "dial-model-id": {
      "endpoint": "http://openai-adapter/openai/deployments/upstream-model-id/chat/completions"
    },
    "dial-model-id-with-override": {
      "endpoint": "http://openai-adapter/openai/deployments/upstream-model-id/chat/completions",
      "overrideName": "upstream-model-id"
    }
  }
}
```

### 1. Azure Chat Completions — model deployment, no `overrideName`

Request to DIAL Core:

```
POST /openai/deployments/dial-model-id/chat/completions
{"model": "foo-bar", "messages": [...]}
```

| | body sent to the adapter |
|---|---|
| before | `{"model": "foo-bar", "messages": [...]}` |
| after  | `{"model": "dial-model-id", "messages": [...]}` |

A body with no `model` at all now also arrives as `{"model": "dial-model-id", ...}`.

### 2. Azure Chat Completions — model deployment with `overrideName`

```
POST /openai/deployments/dial-model-id-with-override/chat/completions
{"model": "foo-bar", "messages": [...]}
```

| | body sent to the adapter |
|---|---|
| before | `{"model": "upstream-model-id", ...}` |
| after  | `{"model": "upstream-model-id", ...}` (unchanged) |

### 3. Azure Completions / Embeddings

Same endpoints family, same controller, same change:

```
POST /openai/deployments/dial-embedding-id/embeddings
{"model": "foo-bar", "input": "hello"}
```

| | body sent to the adapter |
|---|---|
| before | `{"model": "foo-bar", "input": "hello"}` |
| after  | `{"model": "dial-embedding-id", "input": "hello"}` |

### 4. Applications and toolsets — unchanged

```
POST /openai/deployments/dial-app-id/chat/completions
{"model": "foo-bar", "messages": [...]}
```

| | body sent to the application |
|---|---|
| before | `{"model": "foo-bar", ...}` |
| after  | `{"model": "foo-bar", ...}` (unchanged) |

An application is not a model: it has no upstream model to be tricked into serving, no per-model pricing to skew, and applications legitimately read the body's `model` for their own purposes. Only `overrideName` rewrites it, as before.

## The other APIs

Both were checked. **Neither is exploitable the same way, because in both the body's `model` *is* the deployment selector**: DIAL Core resolves the deployment from that value, so a client asking for a pricier model is simply routed — and priced, rate-limited and access-checked — as that deployment. That is the intended behaviour.

### OpenAI Responses API — `POST /openai/v1/responses`

`ResponsesController.dispatch()` reads `request.getModel()` and resolves the deployment from it, so the body's model and the target deployment cannot disagree. After the fix the value is rewritten to the very same deployment id (or to `overrideName`, as before), so nothing changes on the wire — but the invariant is now enforced by the core rather than being a property of how the endpoint happens to route.

The one path where the two could drift apart is the interceptor round-trip: when an interceptor calls Core back, the deployment comes from the per-request key (`initialDeployment`) and not from the body. `ResponsesInterceptorController` deliberately leaves the body's `model` alone (see `handleRequestBody_doesNotOverrideModelName`), so an unmodified interceptor sends back the id it received and the outcome is unchanged; if an interceptor did rewrite it, the final call to the model adapter is now normalized back to the deployment id.

### Anthropic Messages API — `POST /anthropic/v1/messages`, `POST /anthropic/v1/messages/count_tokens`

`MessagesBaseController.handle()` resolves the deployment from the body's `model` in the same way, and this API has no interceptor flow at all, so there is no path on which the body's model can differ from the deployment. Behaviour is unchanged; `X-DIAL-DEPLOYMENT-ID` is still captured before the enhancement chain runs, so it keeps carrying the DIAL deployment id the client asked for.

All three APIs run the same `EnhanceDeploymentRequestFn`, so the single fix covers them together and keeps them consistent.

## Compatibility with the DIAL OpenAI adapter

The adapter (`ai-dial-adapter-openai`) resolves the upstream model name in this order — stated three times in its README, once per upstream shape:

> `overrideName` → the `model` field of the original chat completion request (if present) → `${ADAPTER_DEPLOYMENT_ID}` (the deployment id in the adapter's own URL path).

Concretely, `endpoints/chat_completion.py:106` and `endpoints/embeddings.py:31` do `request_body["model"] = request_body.get("model") or deployment_id`, and Core delivers `overrideName` by writing it into that same `model` field. So the middle rung of that chain is exactly the field this PR takes out of the client's hands.

**Where the body's `model` actually reaches the provider.** The adapter builds its client from the upstream URL (`utils/parsers.py`):

| Upstream endpoint shape | Client | Does the body's `model` matter? |
|---|---|---|
| `…/openai/deployments/{name}/chat/completions` (Azure, last gen) | `AsyncAzureOpenAI(azure_endpoint, azure_deployment)` | **No** — the deployment is pinned in the base URL, so the OpenAI SDK skips its `/deployments/{model}` path injection (`openai/lib/azure.py::_build_request`) and Azure ignores the field. |
| `…/openai/v1/chat/completions` (Azure next gen), OpenAI Platform, Bedrock, vLLM, Mistral, Model Studio | `AsyncOpenAI` / `AsyncBedrockOpenAI` | **Yes** — the model name is the routing key. |
| `…/openai` + Responses API, `…/v1/messages` (Anthropic) | `AsyncAzureOpenAI` without deployment / `AsyncAnthropic` | **Yes** — for the Azure client with no pinned deployment the SDK builds `/deployments/{model}/…` from the body field. |

For every shape in rows 2 and 3 the README's instruction is: *"Because the deployment ID is not included in the upstream URL, specify it in the `overrideName` field."* Those are also precisely the shapes where the old behaviour let a client pick the provider-side model by hand — up to choosing the Azure deployment through the SDK's path injection.

**Effect of this PR on the adapter:**

* `overrideName` keeps winning, unchanged — it is still written into the body's `model` and still sent as `X-DIAL-OVERRIDE-NAME`. Every configuration in the adapter's README follows this shape and behaves identically.
* The adapter's second rung now always carries a Core-chosen value instead of a client-chosen one. Nothing else in the adapter reads the request's `model` for routing or client selection: `endpoints/responses.py:74` prefers `X-DIAL-OVERRIDE-NAME` and falls back to the same field, tokenizer selection in `/tokenize` and `/truncate_prompt` goes through `get_upstream_model_name` (`dial_api/request.py:69`) with the same precedence, and the deployment type, credentials and vendor lookups all key off the adapter's own path id.
* **One behavioural interaction worth calling out:** because Core now always populates `model`, the adapter's last-resort fallback to `${ADAPTER_DEPLOYMENT_ID}` is no longer reachable for model deployments. It mattered only for a config that (a) points at an upstream URL without a pinned model, (b) omits `overrideName`, (c) uses an adapter path id that differs from the DIAL model name, *and* (d) is called by clients that omit `model` from the body — such a config already sent the DIAL model name upstream for any client that did send `model`. The fix is the one the adapter README already prescribes for those upstreams: set `overrideName`.

## Tests

* `EnhanceDeploymentRequestFnTest` — new; every case runs against all three request shapes (`ChatCompletionRequest`, `ResponsesApiRequest`, `MessagesApiRequest`): a model ignores the client-chosen name, a model still applies `overrideName`, a model gets a `model` field even when the body carried none, applications and interceptors keep the client value, applications still apply `overrideName`.
* `DeploymentPostApiTest.testClientChosenModelNameReplacedForChatCompletions` — new end-to-end test: `{"model":"foo-bar"}` posted to `/openai/deployments/gpt-3-turbo/chat/completions` reaches the upstream as `{"model":"gpt-3-turbo"}`.
* `DeploymentPostControllerTest.testHandleRequestBody_NotOverrideModelName` → renamed to `testHandleRequestBody_OverrideClientChosenModelName` and its body now carries a foreign model name, since "not overriding" is no longer the behaviour.

## Docs

* `docs/dynamic-settings/models.md` — spells out what the adapter receives when `overrideName` is absent.
* `openapi-generator/.../ChatCompletionRequest.yaml` — the `model` property is documented as ignored for model deployments.


ai-dial-actions:
<h1>Dependency Review</h1>
None

adubovik:
Close in favour of a fix in the adapters themselves.

PR #1918, opened 2026-09-07
fix: responses background jobs ignore the interfaces config #1881

### LINKED ISSUE #1881: Responses background jobs ignore the interfaces config and read responsesEndpoint directly
### Description

`BackgroundJobService` resolves the Responses API target by reading the deployment field directly:

```java
if (deployment.getResponsesEndpoint() == null) {
    ...
}
String targetUrl = deployment.getResponsesEndpoint() + "/" + mapping.getUpstreamResponseId();
```

Every other Responses path goes through `DeploymentEndpointUtil.resolveResponsesBaseUri(deployment)`, which resolves `interfaces.openaiResponses` first and only then falls back to the legacy `responsesEndpoint`.

As a result, a model that declares the Responses API through `interfaces` rather than the legacy field cannot use background jobs: the poller sees a null `responsesEndpoint` and skips it, even though `POST /openai/v1/responses` on the same model routes correctly.

### Impact

- A model configured only via `interfaces.openaiResponses` (`base_url`, or the deployment-level `baseUrl`) silently loses background-mode polling.
- The gap widens as `interfaces` becomes the normal way to configure a deployment — the legacy field stays optional.

### Suggested fix

Use `DeploymentEndpointUtil.resolveResponsesBaseUri(deployment)` in `BackgroundJobService` in place of the two direct `getResponsesEndpoint()` reads, and add a test for a deployment configured through `interfaces` only.

### Notes

Pre-existing, unrelated to any in-flight branch. Found while reviewing the deployment `interfaces`/`baseUrl` work (#1871).

khshanovskyi:
Background-mode polling for the Responses API resolved its target by reading the deployment's legacy `responsesEndpoint` field directly, so a deployment declaring the API through `interfaces.openaiResponses` was never polled to completion. `BackgroundJobService` now resolves the endpoint the same way every other Responses path does.

### Applicable issues

- fixes #1881

### Description of changes

- `BackgroundJobService.poll` resolves the target url with `DeploymentEndpointUtil.resolveResponsesBaseUri(deployment, translators)` in place of the two direct `getResponsesEndpoint()` reads, so `interfaces.openaiResponses` (its own `base_url`, or the deployment-level `baseUrl`) is honoured and the legacy field stays a fallback.
- The upstream lookup now passes the same `OPENAI_RESPONSES` endpoint supplier the original request was routed with. For a deployment declaring no `upstreams`, `UpstreamRouteProvider` synthesizes an upstream whose id is the supplied endpoint, and the mapping's `upstreamKey` was issued against the Responses serving endpoint — while the poller asked for `Deployment::getEndpoint`. The two ids never matched, so polling failed with `Unknown upstream id`, or an NPE where `endpoint` was absent, for every upstream-less deployment — legacy `responsesEndpoint` configs included.
- Removed `UpstreamRouteProvider.get(Deployment, CacheBreakpointContext, String)`, whose last production caller was the code above; the five test call sites now pass `Deployment::getEndpoint` explicitly.
- Added `BackgroundJobServiceTest` coverage for a deployment configured through `interfaces` only — once with an interface-level `base_url`, once with the deployment-level `baseUrl` — asserting both the polled url and the endpoint supplier handed to the route provider.

### Checklist

- [x] All tests pass (BackgroundJobServiceTest 21/21, UpstreamRouteProviderTest 9/9, Responses and ResponseItem controller suites, checkstyle clean; four unrelated integration tests flaked in the full run and pass on rerun)

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
+++ b/server/src/main/java/com/epam/aidial/core/server/service/BackgroundJobService.java
@@ -4,6 +4,7 @@ import com.epam.aidial.core.config.Config;
 import com.epam.aidial.core.config.Deployment;
 import com.epam.aidial.core.config.InterfaceType;
 import com.epam.aidial.core.config.Model;
+import com.epam.aidial.core.config.Translator;
 import com.epam.aidial.core.config.Upstream;
 import com.epam.aidial.core.credentials.data.credentials.BucketInfo;
 import com.epam.aidial.core.credentials.encryption.CredentialEncryptionService;
@@ -20,6 +21,7 @@ import com.epam.aidial.core.server.token.TokenStatsTracker;
 import com.epam.aidial.core.server.token.TokenUsage;
 import com.epam.aidial.core.server.token.UsagePerModel;
 import com.epam.aidial.core.server.upstream.UpstreamRouteProvider;
+import com.epam.aidial.core.server.util.DeploymentEndpointUtil;
 import com.epam.aidial.core.server.util.ProxyUtil;
 import com.epam.aidial.core.server.util.ResponseIdUtil;
 import com.epam.aidial.core.server.vertx.AsyncTaskExecutor;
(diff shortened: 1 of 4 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

Issue #1930, opened 2026-09-09
App resource dependencies — scoped access for applications to user resources


serguei-gorokhov:
> **Type:** EPIC / Design

## Issue tracker

Statuses: ⬜ todo · 🚧 in progress · 👀 in review · ✅ merged

| Child issue | Status |
|---|---|
| #1931 — S1: definition model — inert `resourceDependencies` parsing | 👀 in review — #1940 |
| #1932 — S2: write-time validation (+ decoded-segment bans, review hardening) | 👀 in review — #1941 |
| #1933 — S3: admin consent on the existing consent API | 👀 in review — #1942 |
| #1934 — S4: request-start resolution (chat path) | 👀 in review — #1943 |
| #1935 — S5: mint-site coverage on conversation APIs + whole-branch review hardening | 👀 in review — #1944 |
| #1936 — S6: admin-consent status GET + `ConsentGrant` provenance | 👀 in review — #1945 |
| #1937 — S7: target-path placeholder into the bucket slot (grammar fix) | 👀 in review — #1946 |
| #1938 — S8: chained-call resolution at every hop | 👀 in review — #1947 |
| #1939 — S9: MCP-application-proxy mint site | 👀 in review — #1948 |

## Problem

An interactive app (QuickApp, code app, assistant) runs while the user is present, and to do its job it must reach the user's resources — a folder of reference files, a conversation, a prompt library — outside its own appdata sandbox. Today no mechanism is both sufficient and honest:

- **The per-request key (PRK)** is underprivileged: it reaches the app's own appdata, request attachments and the user's ambient shared/public resources, and nothing else.
- **The user's JWT** is overprivileged: any credential that makes DIAL validate the app as the user hands over full user identity — the whole own bucket with write access, every ops endpoint, the whole share graph, and no attribution (costs, limits and trace all bill to the user).
- **External services with `DIAL_NATIVE`** is the wrong mechanism for online access: redemption returns the user's full refreshed IdP access token, and its consent moment is offline-shaped ("act while I'm away") — not an honest ask for an app acting while the user is *present*.

The requirement is **scoped access**: this folder, these access types, this app, consented. This epic delivers the resource-dimension slice of the roadmap's granularly-scoped-PRK step, built almost entirely from machinery that already ships in Core.

### Use cases

[...]

PR #1943, opened 2026-09-09
feat: resolve declared resource dependencies at request start (chat path)

### LINKED ISSUE #1934: Resolve declared resource dependencies at request start (chat path)
Parent: #1930

## What to build

A new enhancement-chain function, `ResolveResourceDependenciesFn`, runs for chat-completion calls to an application, **before** the per-request key is assigned — while the context still carries the originating user, so the reach checks evaluate the USER's permissions, not the app's (load-bearing timing, made loud by an assert).

Per declared record: resolve the target (`current-user/…` against the originating user's bucket, concrete global-view paths as-is), verify fresh against the user's reach, intersect with the content-bound admin-consent record, and bake the passing grants into the key the application will hold.

- **Fail-closed per record:** unreachable or unconsented targets simply do not grant; a `required` record that cannot resolve fails the call with an explicit 403 instead of the app half-working. Reach is all-or-nothing per record — no half-grants.
- **Malformed records** (config-file apps bypass write-time validation) are unresolvable — never a crash, never a grant; the read side re-enforces the write-side vocabulary (kind, READ/WRITE only).
- **Runtime audit** on the `DIAL_RESOURCE_DEPS_AUDIT` stream: one grant line per run, one denial line (unreachable or unconsented), one runtime-fail line when a required record rejects the call.
- **Enforcement is unchanged** — the existing per-request permission rule honors the grants, folder targets by prefix.

Smoke test proves the v1 happy path end to end: admin-authored app declares a personal prompts folder (`current-user/prompts/dep-smoke/`, write, required) → admin grants → user chats → the app's key writes inside the declared folder (200) and is refused outside it (403).

## Acceptance criteria

- [ ] A declaring, admin-consented app called by a user who can reach the target: the PRK carries the grant; the app's subsequent resource call succeeds.
- [ ] Same app, user who cannot reach: no grant, no failure, denial audit line.
- [ ] `required` record failing reach/consent: the call fails with an explicit error + runtime-fail audit line.
- [ ] Apps without a section: bit-for-bit unchanged behavior.
- [ ] The timing assert exists and its test fails if the fn is wired after key assignment.

## Blocked by

#1933

serguei-gorokhov:
### Applicable issues

- fixes #1934
- Parent epic: #1930

### Description of changes

A new enhancement-chain function, `ResolveResourceDependenciesFn`, runs for chat-completion calls to an application, **before** the per-request key is assigned — while the context still carries the originating user, so the reach checks evaluate the USER's permissions, not the app's (load-bearing timing, made loud by an assert).

Per declared record: resolve the target (`current-user/…` against the originating user's bucket, concrete global-view paths as-is), verify fresh against the user's reach, intersect with the content-bound admin-consent record, and bake the passing grants into the key the application will hold.

- **Fail-closed per record:** unreachable or unconsented targets simply do not grant; a `required` record that cannot resolve fails the call with an explicit 403 instead of the app half-working. Reach is all-or-nothing per record — no half-grants.
- **Malformed records** (config-file apps bypass write-time validation) are unresolvable — never a crash, never a grant; the read side re-enforces the write-side vocabulary (kind, READ/WRITE only).
- **Runtime audit** on the `DIAL_RESOURCE_DEPS_AUDIT` stream: one grant line per run (what was granted), one denial line (unreachable or unconsented), one runtime-fail line when a required record rejects the call.
- **Enforcement is unchanged** — the existing per-request permission rule honors the grants, folder targets by prefix.

Smoke test proves the v1 happy path end to end: admin-authored app declares `current-user/prompts/dep-smoke/` (write, required) → admin grants → user chats → the app's key writes inside the declared folder (200) and is refused outside it (403).

### Behavior changes flagged for review

- An app that declares a `required` dependency can now fail chat-completion calls with an explicit 403 (reach or consent missing) — new failure mode, only for apps that opt in by declaring.
- The timing assert makes wiring-after-key-assignment fail loudly rather than silently resolve against the wrong identity.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
@@ -1,0 +1,210 @@
+package com.epam.aidial.core.server.function;
+
+import com.epam.aidial.core.config.Application;
+import com.epam.aidial.core.config.ResourceAccessType;
+import com.epam.aidial.core.config.ResourceDependency;
+import com.epam.aidial.core.server.Proxy;
+import com.epam.aidial.core.server.ProxyContext;
+import com.epam.aidial.core.server.data.ApiKeyData;
+import com.epam.aidial.core.server.data.AuthBucket;
+import com.epam.aidial.core.server.data.consent.Consent;
+import com.epam.aidial.core.server.data.permission.PerRequestSharedData;
+import com.epam.aidial.core.server.function.request.RequestObject;
+import com.epam.aidial.core.server.log.ResourceDependencyAuditLog;
+import com.epam.aidial.core.server.security.AccessService;
+import com.epam.aidial.core.server.service.ResourceDependencyValidator;
+import com.epam.aidial.core.server.util.BucketBuilder;
+import com.epam.aidial.core.server.util.ResourceDescriptorFactory;
+import com.epam.aidial.core.storage.http.HttpException;
+import com.epam.aidial.core.storage.http.HttpStatus;
+import com.epam.aidial.core.storage.resource.ResourceDescriptor;
+import com.epam.aidial.core.storage.resource.ResourceType;
+import com.epam.aidial.core.storage.resource.ResourceTypes;
+import com.epam.aidial.core.storage.util.UrlUtil;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Request-start resolution of the application's declared resource dependencies (design §7.1):
+ * resolve each declared target, verify it fresh against the originating user's reach, intersect
+ * with the admin-consented set, and bake the passing grants into the per-request key the
+ * application will hold — the app never asks for a credential; the key it already holds gets
+ * richer. A record is a request, not a grant: nothing here widens anything the user cannot
+ * already reach.
+ */
+public class ResolveResourceDependenciesFn extends BaseRequestFunction<RequestObject> {
+
+    public ResolveResourceDependenciesFn(Proxy proxy, ProxyContext context) {
+        super(proxy, context);
+    }
+
+    @Override
+    public Boolean apply(RequestObject request) {
+        if (!(context.getDeployment() instanceof Application application)) {
+            // Interceptor hop or a non-application deployment — dependencies resolve for the
+            // application being called, nothing else.
+            return false;
+        }
+        List<ResourceDependency> declaration = application.getResourceDependencies();
+        if (declaration == null || declaration.isEmpty()) {
+            return false;
+        }
+        // Load-bearing timing: this function runs in the enhancement chain BEFORE the per-request key
+        // is assigned, so the reach checks below evaluate the originating user's permissions. After
+        // assignment the same calls would silently evaluate the app's own key instead — make that
+        // drift loud instead of silent.
+        if (context.getApiKeyData().getPerRequestKey() != null) {
+            throw new IllegalStateException(
+                    "Resource dependencies must be resolved before the per-request key is assigned");
+        }
+        resolve(application, declaration);
+        return false;
+    }
+
+    private void resolve(Application application, List<ResourceDependency> declaration) {
+        String applicationId = application.getName();
+        // The consent record is content-bound to the whole declaration: any change since the grant
+        // re-requires it, and until then nothing resolves.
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+        return Set.copyOf(dependency.getAccess());
+    }
+
+    /**
+     * Resolves a declared target path to a descriptor: a {@code current-user/…} path against the
+     * originating user's own bucket, a concrete global-view path as-is. Null when the record is
+     * malformed — an unresolvable record, never a crash.
+     */
+    @Nullable
+    private ResourceDescriptor resolveTarget(ResourceDependency dependency, AuthBucket userBucket) {
```
astsiapanay, line 170: Let's don't invent something new for parsing DIAL resource URL

server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+        resolve(application, declaration);
+        return false;
+    }
+
+    private void resolve(Application application, List<ResourceDependency> declaration) {
+        String applicationId = application.getName();
+        // The consent record is content-bound to the whole declaration: any change since the grant
+        // re-requires it, and until then nothing resolves.
+        boolean consented = proxy.getConsentService().isAdminConsented(applicationId, declaration);
+        AccessService accessService = proxy.getAccessService();
```
astsiapanay, line 75: the service variables could be class members.

server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+        resolve(application, declaration);
+        return false;
+    }
+
+    private void resolve(Application application, List<ResourceDependency> declaration) {
+        String applicationId = application.getName();
+        // The consent record is content-bound to the whole declaration: any change since the grant
+        // re-requires it, and until then nothing resolves.
+        boolean consented = proxy.getConsentService().isAdminConsented(applicationId, declaration);
```
astsiapanay, line 74: User may consent too. But the admin consent is checked only.

server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+            ResourceDescriptor target = requestedAccess == null ? null : resolveTarget(dependency, userBucket);
+            if (!consented || target == null) {
+                // Fail closed per record: unconsented, malformed (config-file apps bypass write-time
+                // validation) — no grant, no failure, unless required.
+                trackUnresolved(dependency, target, unresolved, requiredFailures);
+                continue;
+            }
+            Set<ResourceAccessType> userAccess =
+                    accessService.lookupPermissions(Set.of(target), context).getOrDefault(target, Set.of());
+            if (userAccess.containsAll(requestedAccess)) {
```
astsiapanay, line 94: The block under the branch `if` could be a separate method

server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+                // Both halves of the delivery: perRequestSharedResources serves the application's own
```
astsiapanay, line 95: There are many unnecessary comments. Let's leave ones which have real meaning

server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -0,0 +1,210 @@
...
+        List<ResourceDependency> declaration = application.getResourceDependencies();
+        if (declaration == null || declaration.isEmpty()) {
+            return false;
+        }
+        // Load-bearing timing: this function runs in the enhancement chain BEFORE the per-request key
+        // is assigned, so the reach checks below evaluate the originating user's permissions. After
+        // assignment the same calls would silently evaluate the app's own key instead — make that
+        // drift loud instead of silent.
+        if (context.getApiKeyData().getPerRequestKey() != null) {
+            throw new IllegalStateException(
```
astsiapanay, line 63: Supposing user makes a chat completion request via DIAL chat. PRK is not available since user is authenticated via JWT. Why do we throw the exception here?

astsiapanay (changes requested):

PR #1944, opened 2026-09-09
feat: resolve dependencies on the remaining conversation mint sites + e2e suite

### LINKED ISSUE #1935: Complete mint-site coverage on conversation APIs and land whole-branch review hardening
Parent: #1930

## What to build

Two halves, deliberately shipped as one story — reviewers should treat the mixed diff as intentional:

**Half 1 — mint-site coverage on the conversation APIs.** `ResolveResourceDependenciesFn` now runs in the same chain position on the Anthropic Messages (`MessagesBaseController.prepareUpstreamRoute`) and Responses (`ResponsesController.handleRequestBody`) entry points, completing the v1 conversation mint-site coverage (route/feature/deploy sites remain a documented v1 limitation — an app invoked through those surfaces gets no dependency grants in v1).

An e2e suite (`ResourceDependencyApiTest`) covers the v1 checklist as test cases: the consented flow, required-without-consent failing the call explicitly, optional-without-consent degrading silently (target stays off-limits), withdrawal stopping the run immediately, declaration change invalidating the grant (content binding), public-target grants following the originating user's own reach exactly (pointer semantics), and the audit events carrying app, targets, user and trace id (the grant line also carries access types).

**Half 2 — whole-branch review hardening** (found by the whole-branch security/review pass; its commit is chronologically adjacent and folds in here):

- **Root-call-only guard semantics.** The timing guard threw on every hop that arrives with a per-request key — the normal state for an interceptor's final call back to a declaring app (breaking even the root user request) and for every chained call into one. Resolution is now honestly root-call-only: hops under a per-request key are skipped — callable, just without grants — because the originating user's reach is not evaluable under a key context. Chained composition is recorded as phase-2 machinery.
- **Internal-type ban.** The placeholder type segment enforces the validator's vocabulary: `ResourceTypes.of()` maps internal engine types, so a `current-user/credentials/` declaration from a config-file or admin-authored app could otherwise resolve into the user's secret-bearing blobs.
- **Union grants.** Two records sharing a target combine rights instead of overwriting each other; malformed concrete paths are caught (unresolvable, never a crash); target permission lookups are batched into one rule-chain walk; consent is checked before any resolution work; the audit's access-type join tolerates null elements. The `perRequestReceivers[app]` half of the bake — which never propagated on plain chained calls and served no reachable flow — is removed; the grant lives solely in the key's own shared map.

New e2e: a chained call into a declaring app stays callable without receiving grants (the case the old guard broke).

## Acceptance criteria

- [ ] Consented flow works on the chat, Anthropic Messages, and Responses entry points; required-without-consent fails the call explicitly (403 + runtime-fail audit line).
- [ ] Optional-without-consent degrades silently; withdrawal stops the run immediately; declaration change invalidates the grant.
- [ ] Public-target grants follow the originating user's own reach exactly (public is not exempt).
- [ ] A chained call into a declaring app stays callable without receiving grants; interceptor-protected declaring apps work on root calls.
- [ ] `current-user/credentials/` (internal types) never resolve; records sharing a target combine rights (union), never overwrite.
- [ ] Audit events carry app, targets, user and trace id — verified by content, not just presence; the grant line also carries access types.
- [ ] Full `:server:test` green on CI Linux (modulo the documented pre-existing Windows-environmental failures that fail identically on pristine `development`).

## Blocked by

#1934

serguei-gorokhov:
### Applicable issues

- fixes #1935
- Parent epic: #1930

### Description of changes

This story deliberately folds **two halves into one PR** — the mixed diff is intentional (the whole-branch review hardening commit is chronologically adjacent to the mint-site commit); reviewers should assess both.

**Half 1 — mint-site coverage on the conversation APIs (`ad900889`).** `ResolveResourceDependenciesFn` now runs in the same chain position on the Anthropic Messages (`MessagesBaseController.prepareUpstreamRoute`) and Responses (`ResponsesController.handleRequestBody`) entry points, completing the v1 conversation mint-site coverage (route/feature/deploy sites remain a documented v1 limitation — an app invoked through those surfaces gets no dependency grants in v1). `ResourceDependencyApiTest` covers the v1 checklist as test cases: the consented flow, required-without-consent failing the call explicitly, optional-without-consent degrading silently, withdrawal stopping the run immediately, declaration change invalidating the grant (content binding), public-target grants following the originating user's own reach exactly, and audit events carrying app, targets, user and trace id (the grant line also carries access types).

**Half 2 — whole-branch review hardening (`2a40b3f6`).** Three real defects and a defense-in-depth gap, all in the composed behavior:

- **Root-call-only guard semantics.** The timing guard threw `IllegalStateException` on every hop that arrives with a per-request key — the normal state for an interceptor's final call back to a declaring app (breaking even the root user request) and for every chained call into one (400/500 on all three interfaces). Resolution is now honestly root-call-only: hops under a per-request key are skipped — callable, just without grants — because the originating user's reach is not evaluable under a key context (`buildUserBucket` flips to the parent app's bucket). Chained composition is recorded as phase-2 machinery.
- **Internal-type ban.** The placeholder type segment now enforces the validator's `PERSONAL_TYPED_ROOTS` vocabulary (constant made public, shared): `ResourceTypes.of()` maps internal engine types, so a `current-user/credentials/` declaration from a config-file or admin-authored app could otherwise resolve into the user's secret-bearing blobs.
- **Union grants.** Two records sharing a target combine rights instead of overwriting each other; malformed concrete paths raise `RuntimeException` shapes that are now caught (unresolvable, never a crash); target permission lookups are batched into one rule-chain walk; consent is checked before any resolution work; the audit's access-type join tolerates null elements. The `perRequestReceivers[app]` half of the bake — which never propagated on plain chained calls and served no reachable flow — is removed; the grant lives solely in the key's own shared map.

New e2e: a chained call into a declaring app stays callable without receiving grants (the case the old guard broke).

### Behavior changes flagged for review

- **Required-dependency hard-fail on chained hops is NOT here** (that arrives with the chained-resolution PR later in the stack) — in this PR, hops under a per-request key are skipped, never failed.
- Internal engine types (`credentials`, `keys`, …) are no longer resolvable via the placeholder, whoever authored the app.
- Records sharing a target now union their grants (previously last-write-wins overwrite).

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
@@ -24,19 +24,28 @@ import com.epam.aidial.core.storage.util.UrlUtil;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
- * Request-start resolution of the application's declared resource dependencies (design §7.1):
- * resolve each declared target, verify it fresh against the originating user's reach, intersect
- * with the admin-consented set, and bake the passing grants into the per-request key the
- * application will hold — the app never asks for a credential; the key it already holds gets
- * richer. A record is a request, not a grant: nothing here widens anything the user cannot
- * already reach.
+ * Request-start resolution of the called application's declared resource dependencies (design
+ * §7.1): resolve each declared target, verify it fresh against the originating user's reach,
+ * intersect with the content-bound admin-consent record, and bake the passing grants into the
+ * per-request key the application will hold — the app never asks for a credential; the key it
+ * already holds gets richer. A record is a request, not a grant: nothing here widens anything
+ * the user cannot already reach.
+ *
+ * <p><b>Root-call only.</b> Resolution runs when the context still carries the originating user
+ * (no per-request key) — that is the only state in which the user's reach is directly
+ * evaluable; under a per-request key the same checks would silently evaluate a deployment's own
+ * key instead. Hops that arrive with a per-request key present — an interceptor's final call
+ * back to the app, or a chained app-to-app call — are skipped: their declarations do not
+ * resolve and their grants do not propagate in v1 (documented limitation; chained composition
+ * needs originating-user evaluation under key contexts, which is phase-2 machinery).
  */
 public class ResolveResourceDependenciesFn extends BaseRequestFunction<RequestObject> {
 
@@ -55,13 +64,10 @@ public class ResolveResourceDependenciesFn extends BaseRequestFunction<RequestOb
         if (declaration == null || declaration.isEmpty()) {
             return false;
         }
-        // Load-bearing timing: this function runs in the enhancement chain BEFORE the per-request key
-        // is assigned, so the reach checks below evaluate the originating user's permissions. After
-        // assignment the same calls would silently evaluate the app's own key instead — make that
-        // drift loud instead of silent.
         if (context.getApiKeyData().getPerRequestKey() != null) {
-            throw new IllegalStateException(
-                    "Resource dependencies must be resolved before the per-request key is assigned");
+            // Not the root user call — see the class javadoc. Skip, never throw: a declaring app
+            // behind an interceptor or in a chain must stay callable, just without grants.
+            return false;
         }
         resolve(application, declaration);
         return false;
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/epam/aidial/core/server/function/ResolveResourceDependenciesFn.java
```diff
@@ -24,19 +24,28 @@
...
 import javax.annotation.Nullable;
 
 /**
- * Request-start resolution of the application's declared resource dependencies (design §7.1):
- * resolve each declared target, verify it fresh against the originating user's reach, intersect
- * with the admin-consented set, and bake the passing grants into the per-request key the
- * application will hold — the app never asks for a credential; the key it already holds gets
- * richer. A record is a request, not a grant: nothing here widens anything the user cannot
- * already reach.
+ * Request-start resolution of the called application's declared resource dependencies (design
```
astsiapanay: Please don't add extensive javadoc. Make it compact and meaningful

PR #1950, opened 2026-09-09
fix: allow erasing write-only secrets with explicit null on PUT #1949

### LINKED ISSUE #1949: Admin API: no way to erase a write-only secret on entity update (PUT)
## Name and Version

AI DIAL Core

## What steps will reproduce the bug?

1. Create a model with a secret extra data on the upstream:
   ```
   PUT /v1/models/{bucket}/my-model
   { "upstreams": [ { "endpoint": "https://vendor.example/v1/chat/completions", "secretExtraData": "{\"field\":\"value\"}" } ], ... }
   ```
   → 200 OK
2. Later, switch the upstream to a plain-string `extraData` (intent: replace the pair):
   ```
   PUT /v1/models/{bucket}/my-model
   { "upstreams": [ { "endpoint": "https://vendor.example/v1/chat/completions", "extraData": "abc" } ], ... }
   ```
   → **422** `When both extraData and secretExtraData are set, both must be JSON objects`

   The omitted `secretExtraData` is silently preserved from the stored entity (omit-preserve in `SecretFieldProcessor.mergeInto`: `current == null || current.isNull()` → copy prior ciphertext), so after the merge both fields are set and `UpstreamExtraDataMerger.validateNoOverlap` (via `ConfigResourceController.checkCrossReferences`) rejects the non-JSON `extraData`.
3. Try `"secretExtraData": null` instead — pre-fix it hits the same preserve branch, so the outcome is identical. No request clears the field; the only workaround is deleting and recreating the model.

## What is the expected behavior?

A client must be able to distinguish three update intents for a secret field:
- omitted → keep the stored secret
- explicit `null` → erase the secret
- a value → replace the secret

With `null` meaning erase, step 2 sent as `{ "extraData": "abc", "secretExtraData": null }` should return 200 and leave the upstream with plain `extraData: "abc"` and no secret extra data.

## What do you see instead?

An explicit `null` is treated identically to an omitted field — the stored secret is silently preserved. There is no request that can remove a secret: GET drops write-only (`@EncryptedField`) fields entirely, so a client cannot read the value to unset it any other way. Once a JSON `secretExtraData` is stored, the upstream can never be switched to a plain-string `extraData` without deleting the entity.

The same applies to `Upstream.key`: after `"key": null` the credential is still sent as `X-UPSTREAM-KEY` on proxied requests. Affected fields: `Upstream.key`, `Upstream.secretExtraData`, `UpstreamInterface.key`, `UpstreamInterface.secretExtraData` (via `Model.upstreams`, `Route.upstreams`, `Upstream.interfaces`) on per-entity PUT. `Key.key` is mandatory on PUT (400 on blank/null), so erase does not apply there. An empty string "" incidentally overwrites the secret with "" — undocumented and unpinned by tests.

KirylKurnosenka:
On a per-entity PUT update there was no way to erase a write-only secret: an explicit `null` was treated identically to an omitted field and silently preserved the stored ciphertext (see #1949 — once a JSON `secretExtraData` was stored, the upstream could never be switched to a plain-string `extraData`). This change makes the three update intents explicit: omitted → preserve, explicit `null` → erase, a literal value (including the empty string) → replace.

### Applicable issues

- fixes #1949

### Description of changes

- `SecretFieldProcessor.mergeInto`: the preserve branch no longer treats an explicit `null` as omit — the request's `NullNode` now flows into the entity and the blob writer omits the field, so the erase carries end-to-end. Renamed `mergePreservingOmittedSecrets` → `mergeUpdateSecrets` to reflect the three-intent contract.
- `ConfigResourceController.handlePut`: updated call site and replaced the stale sentinel-era comment.
- Nested paths (`Model.upstreams`, `Route.upstreams`, `Upstream.interfaces`) are covered automatically by the existing recursion in `mergeArray`/`mergeMap`.
- `Key.key` is unchanged: an explicit `null` still yields 400 via `validateKeyForApiWrite` — erase does not apply to a mandatory field.
- An empty string stays a literal value, not an erase signal (pinned by a test).
- Tests: 5 new `SecretFieldProcessorTest` cases (top-level, upstream array element, interface map entry, empty string, null with no prior value); end-to-end `ModelWriteApiTest.testPutExplicitNullUpstreamKeyErasesSecret` (no `X-UPSTREAM-KEY` header after erase) and `ConfigEntityWriteApiTest.testKeyPutExplicitNullKeyRejected400`.
- OpenAPI: `nullable: true` + descriptions on the `Upstream`/`UpstreamInterface` secret fields documenting the three intents.

### Checklist

### DIFF
--- a/docs/open_api_core.yaml
+++ b/docs/open_api_core.yaml
@@ -16134,6 +16134,7 @@ components:
         key:
           type: string
           writeOnly: true
+          description: "Write-only secret, required on create. On PUT update an omitted key is preserved from the stored entity; an explicit null is rejected with 400."
         project:
           type: string
         role:
@@ -17738,11 +17739,15 @@ components:
         key:
           type: string
           writeOnly: true
+          nullable: true
+          description: "Write-only secret. On PUT update: omitted - preserved from the stored entity; explicit null - erased; a value (including the empty string) - replaces the secret."
         responsesEndpoint:
           type: string
         secretExtraData:
           type: string
           writeOnly: true
+          nullable: true
+          description: "Write-only secret. On PUT update: omitted - preserved from the stored entity; explicit null - erased; a value (including the empty string) - replaces the secret."
         tier:
           type: integer
         weight:
@@ -17761,9 +17766,13 @@ components:
         key:
           type: string
           writeOnly: true
+          nullable: true
+          description: "Write-only secret. On PUT update: omitted - preserved from the stored entity; explicit null - erased; a value (including the empty string) - replaces the secret."
         secretExtraData:
           type: string
           writeOnly: true
+          nullable: true
+          description: "Write-only secret. On PUT update: omitted - preserved from the stored entity; explicit null - erased; a value (including the empty string) - replaces the secret."
     UserInfoResponse:
       type: object
       properties:
--- a/server/src/main/java/com/epam/aidial/core/server/config/SecretFieldProcessor.java
+++ b/server/src/main/java/com/epam/aidial/core/server/config/SecretFieldProcessor.java
@@ -106,9 +106,9 @@ public class SecretFieldProcessor {
         }
     }
 
-    public ObjectNode mergePreservingOmittedSecrets(JsonNode existingBlobNode,
-                                                    JsonNode requestNode,
-                                                    Class<?> entityClass) {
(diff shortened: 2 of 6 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1953, opened 2026-09-10
fix: key stops working after PUT that omits the key field #1927

### LINKED ISSUE #1927: [Catalog-Keys] Key with Admin role stops working after updating Project field
## EPAM AI DIAL Admin version
Admin: N/A

## How to reproduce
1. Navigate to Admin → Catalog → Keys
2. Create a key with the **Admin** role
3. Verify the key works (e.g. send a request via Postman)
4. Open the key and update the **Project** field
5. Try to use the key again

## Actual result
The key becomes non-functional after updating the Project field.

## Expected result
The key should remain valid and functional after updating the Project field.

## Additional information
No response

## Confidential information
I confirm that do not share any confidential information

KirylKurnosenka:
A key PUT whose body omits `key` (the documented preserve-on-omit update) returned 200 but revoked the very bearer it was supposed to preserve: `handlePut` captured `keySecret` before any decryption, so `ApiKeyStore` was re-indexed under the preserved `ENC[...]` ciphertext and the rotation cleanup removed the still-working plaintext secret.

The fix moves the `keySecret` capture and the `addOrUpdateKey`/`removeKey` block to after the existing decrypt-in-place, so the capture is plaintext in every arm (create, rotation, omit).

### Applicable issues

- fixes #1927

### Description of changes

- `ConfigResourceController.handlePut`: capture `keySecret` from the decrypted `Key` entity (after the post-put `decryptFields`) instead of before `encryptFields`, gating the store update on `keyEntity != null`. On preserve-on-omit this keeps the correct plaintext index in `ApiKeyStore` and short-circuits the old-secret removal; rotation cleanup is unchanged. The ordering now matches the DELETE path and the `MergedConfigStore` replica fast path, which both capture the secret after decryption.
- `ConfigEntityWriteApiTest.testKeyPut200PreserveKeyOnOmit`: assert the old bearer still authenticates (`GET /v1/bucket` with `Api-key: <old plaintext>`) after the omitted-key PUT — verified to fail on the pre-fix code and pass with the fix.

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
+++ b/server/src/main/java/com/epam/aidial/core/server/controller/ConfigResourceController.java
@@ -1561,7 +1561,6 @@ public class ConfigResourceController implements Controller {
 
                 String blobBody;
                 Key keyEntity = null;
-                String keySecret = null;
                 String oldSecret = null;
                 Object entity = null;
                 Map<String, String> putEventMetadata = null;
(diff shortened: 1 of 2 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1954, opened 2026-09-10
feat: fixed calendar windows for day/week/month rate limits #1951

### LINKED ISSUE #1951: Fixed calendar windows for day/week/month rate limits
## Why do we need this feature?

Today every rate-limit period (minute/hour/day/week/month) uses a single "floating window" model: usage ages out gradually as time advances, so e.g. the "month" limit is really a rolling 30-day total, not a calendar month. That's the right behavior for short periods where you want smooth, gradually-replenishing limits that absorb bursts — but it's a poor fit for day/week/month limits used for **financial analysis and reporting**, where stakeholders expect calendar-aligned periods (e.g. "this month" = the 1st through the last day of the calendar month, the way monthly billing cycles or Claude Code's own monthly usage reporting work).

Once day/week/month become fixed calendar windows, they gain something a floating window never had: a single, deterministic reset instant. That's worth surfacing to clients of the bulk limits/usage endpoints directly — a user blocked on a monthly cap can be told exactly when they'll be unblocked, not just that they're over.

**Usecases:**

1. Finance/ops needs monthly usage figures that line up with calendar months (and, more generally, with the organization's own reporting calendar) so they can be compared against invoices, budgets, and other calendar-period financial data without manual reconciliation.
2. Deployments whose fiscal/reporting week doesn't start on a fixed day need to align the "week" limit's reset with their own business week (e.g. Monday 9am local time instead of an arbitrary rolling 7 days).
3. Peak-load protection (minute/hour limits) should keep behaving as it does today — smooth and gradually replenishing — since that's a different concern (burst absorption) from calendar-aligned reporting.
4. A chat user blocked by a monthly/weekly/daily cap wants to know exactly when they'll be unblocked, not just that they're over.
5. A billing/finance dashboard wants to show "resets Oct 1" next to a monthly figure, matching the calendar-aligned framing of this design in the first place.
6. A client rendering the quota indicator wants this for free on the same response it already fetches (`GET /v1/user/limits`), not a second call.

## List of technical requirements for this feature

1. MINUTE and HOUR limits keep today's floating-window behavior, unchanged.
2. DAY, WEEK, and MONTH limits become fixed calendar windows: usage resets all at once at a deterministic boundary rather than aging out gradually.
3. The calendar anchor (what time of day a period resets, and — for weeks — which day it starts on) must be configurable, not hardcoded to UTC midnight. Given inputs like "Monday, 9am, Europe/Warsaw" or "Sunday, 12am, America/New_York".
4. The anchor is a single setting for the whole deployment — no per-role or per-individual-user customization, and no self-service API for end users to set their own.
5. The anchor is expressed as a real IANA timezone (e.g. `Europe/Warsaw`), not a fixed UTC offset — so a "day" is not always exactly 24 real hours. DST transitions are a genuine edge case the boundary math must handle: a calendar day can be 23 hours (spring-forward) or 25 hours (fall-back) on the two transition dates each year. Period-start/period-end must therefore be computed with zone-aware date/time arithmetic (e.g. `ZonedDateTime`, which correctly advances by one calendar day in local wall-clock time regardless of how many real hours that spans), not raw millisecond math.
6. This applies uniformly to every limit family that has day/week/month periods today (token limits, cost limits, and the request-count limit's daily period) — anything named "day", "week", or "month" becomes fixed; anything named "minute" or "hour" stays floating.
7. Once a fixed-window limit is hit, the caller is blocked until the exact next period boundary (all-or-nothing reset) — there is no partial/gradual replenishment for day/week/month, unlike minute/hour.
8. This is a hard cutover: floating-window day/week/month behavior is removed outright, with no deployment-level toggle to opt back into it.
9. No change to enforcement or to the 429 "limit exceeded" error response: the `Retry-After` signal keeps its current shape, just becoming deterministic (computed from the fixed calendar boundary) instead of estimated by aging out old usage.
10. The bulk stats endpoints (`GET /v1/deployments/{deployment_name}/limits`, `GET /v1/user/limits`, `GET /v1/user/usage`) do gain new information: reset info is reported for DAY/WEEK/MONTH (fixed windows), where the reset moment is a single deterministic instant. MINUTE/HOUR (floating windows) have no such instant — usage decays gradually, not all at once — so they do not report one.
11. That reset value is an absolute instant (e.g. `"2026-10-01T00:00:00+02:00"`), not a countdown, so a client can display, format, or derive a countdown from it without its own clock math.
12. One shape, no new vocabulary: the same `{total, used}` pair type used for every window today gains the new field, rather than introducing a second type only fixed windows use. A client already parsing `ItemLimitStats`/`CostItemLimitStats` for every window keeps one parser.
13. The reset value must be computable even for a deployment with zero usage (an accessible-but-never-touched deployment is still reported, matching the existing bulk-endpoint behavior of reporting every accessible deployment rather than omitting unused ones) — it cannot depend on a stored usage record existing.

## Suggested solution overview

Introduce one deployment-wide "rate limit schedule" setting with three parts: a timezone, a week start day, and a daily reset time. DAY periods reset every day at the configured local time in the configured timezone; WEEK periods additionally only reset on the configured start day; MONTH periods reset on the 1st calendar day at the configured local time in the configured timezone (correctly spanning variable month lengths). If the schedule isn't explicitly configured, it defaults to UTC, Monday, midnight — which reproduces today's implicit UTC-midnight day/month behavior and only changes the (previously undefined) week-start behavior.

This sits alongside the existing per-role `limits`/`costLimit` configuration, but is deployment-wide rather than per-role, so it lives at the top level of the config document. Field formats:

```json
{
  "type": "object",
  "properties": {
    "timezone": {
      "type": "string",
      "default": "UTC",
      "description": "IANA timezone id the schedule is anchored to, e.g. \"Europe/Warsaw\", \"America/New_York\", or \"UTC\". No fixed pattern (IANA ids don't fit one shape) — validate against the runtime's known zone ids instead. DST-observing zones mean a calendar day is not always exactly 24 real hours; see technical requirement 5."
    },
    "weekStartDay": {
      "type": "string",
      "enum": ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"],
      "default": "Mon",
      "description": "Day the WEEK window starts on, using the well-known 3-letter abbreviation."
    },
    "resetTime": {
      "type": "string",
      "pattern": "^([01][0-9]|2[0-3]):[0-5][0-9]$",
      "default": "00:00",
      "description": "24h time of day (HH:mm) at which DAY/WEEK/MONTH periods reset, local to timezone."
    }
  },
  "additionalProperties": false
}
```

Example, and where it sits relative to the existing per-role config:

```json
{
  "rateLimitSchedule": {
    "timezone": "Europe/Warsaw",
    "weekStartDay": "Mon",
    "resetTime": "09:00"
  },
  "roles": {
    "<role>": {
      "limits": {
        "<model|app>": {
          "minute": "150000",
          "day": "432000000",
          "week": "5000000000",
          "month": "20000000000"
        }
      }
    }
  }
}
```

Omitting `rateLimitSchedule` entirely is equivalent to `{"timezone": "UTC", "weekStartDay": "Mon", "resetTime": "00:00"}`.

Structurally, day/week/month tracking moves from the current "ring buffer of recent sub-intervals" model to a much simpler "current period total + which period this is" model: on each check, recompute which calendar period the current time falls into; if it's a new period compared to what's stored, the counter starts over at zero for that period. This is simpler than the floating-window model precisely because a fixed window doesn't need to track any historical sub-interval detail — it only ever needs "how much has happened in the period we're currently in." Conceptually, where today's stored record for a window is a set of aging sub-interval buckets plus a rolling total, the new stored record is just:

```json
{
  "periodStart": "2026-09-07T09:00:00+02:00",
  "count": 1234567
}
```

`periodStart` is the start of the calendar period this `count` belongs to, as computed from the schedule above. A read simply recomputes what `periodStart` *should* be for "now"; if it doesn't match the stored value, the period has rolled over and `count` is treated as `0` before adding the current request's usage.

Note the offset embedded in `periodStart` (`+02:00` above) is whatever UTC offset the configured timezone actually observed on that date — it reflects the zone's DST state at that instant, not a fixed value. Two `periodStart`s that share the same local wall-clock time (e.g. both "09:00") can carry different offsets if a DST transition happened between them, since `periodStart` always identifies one absolute instant on the timeline.

**Naming note:** today's floating-window record already has a field literally called `sum` (the live rolling total across sub-intervals). The new record's total is deliberately named `count` instead, not `sum`, even though the two never coexist on the same record (a window is either floating or fixed, never both). Reusing `sum` would still work correctly — the mismatched `periodStart` forces a reset before the value is ever read, so a stray old `sum` value could never leak into a new-shape total — but it would leave two fields with the same name and different meanings (one instantaneous-slice total, one period-to-date total) across the codebase and in any raw-blob inspection during rollout, which is exactly the kind of ambiguity worth avoiding for a one-line rename.

Minute and hour periods are untouched — they keep their existing gradual, sliding-window behavior and stored shape.

Because the stored shape for day/week/month usage changes (old records have no `periodStart`/`count` fields in this shape), previously-persisted counters for those periods won't be recognized by the new logic and will be treated as if no usage had been recorded yet for the current period — i.e. counters implicitly reset to zero once on rollout, with no explicit migration step required.

### Surfacing reset time on the limits/usage endpoints

Add a `resetsAt` field to the existing per-window pair type (today just `{total, used}`), populated only for DAY/WEEK/MONTH. For MINUTE/HOUR the field is omitted from the response entirely, rather than serialized as an explicit `null` — there is nothing to report, so nothing is reported. Both `ItemLimitStats` (tokens/requests) and `CostItemLimitStats` (cost) gain the same field, so every one of `LimitStats`'s ten window fields — and `UserLimitStats`'s four top-level global cost pairs — behaves identically: present with a real instant for the three fixed windows, absent for the two floating ones.

```json
{
  "minuteTokenStats": { "total": 100000,   "used": 1200 },
  "dayTokenStats":    { "total": 10000000, "used": 42000, "resetsAt": "2026-09-10T00:00:00+02:00" },
  "weekTokenStats":   { "total": 5000000000, "used": 180000, "resetsAt": "2026-09-14T09:00:00+02:00" },
  "monthTokenStats":  { "total": 20000000000, "used": 640000, "resetsAt": "2026-10-01T00:00:00+02:00" }
}
```

This is computable without reading any stored usage record: `resetsAt` for a fixed window is a pure function of *which* calendar period "now" falls into, under the deployment-wide `rateLimitSchedule` above — the same schedule and boundary math already used for `periodStart`. It is simply "the start of the *next* period after now," using the identical zone-aware arithmetic already specified for computing the *current* period's `periodStart`. Since it depends only on the current instant and the schedule (not on any per-deployment counter), it costs nothing extra to compute for every deployment in a bulk response, including ones with no usage — satisfying requirement 13 for free.

No change to storage or to enforcement: this is purely an additional read-side projection alongside the existing `used` projection, computed from the same timestamp that projection already uses. It also does not touch `Retry-After` (see requirement 9) — that remains seconds-until-reset on the 429 path; `resetsAt` is additive, on the bulk stats endpoints only.

## Changes for operational processes caused by this feature

- A new deployment-wide config setting (rate limit schedule: timezone, week start day, daily reset time) needs to be documented and, if left unset, defaults to UTC / Monday / midnight — equivalent to today's behavior for day and month, with week gaining a defined start day for the first time.
- On rollout, all previously-accumulated day/week/month usage counters are implicitly reset to zero (see above) — operators should be aware that this is a one-time, expected reset, not data loss to investigate.
- This is a breaking behavior change for any deployment currently relying on day/week/month acting as a rolling window rather than a calendar-aligned one; there is no config flag to preserve the old behavior, so this should be called out clearly in release notes.
- Minute/hour limits and their operational characteristics (smoothing peak load) are unaffected.
- Existing documentation describing these endpoints as strictly trailing windows with no reset to point at should be updated once this ships — that no longer holds for DAY/WEEK/MONTH, only for MINUTE/HOUR.
- The absence of `resetsAt` on MINUTE/HOUR entries must be documented as an expected, permanent omission (not a bug or a "not yet populated" placeholder) — a client should treat a missing field as "no single reset instant exists for this window," not poll waiting for it to appear.

astsiapanay:
Converts DAY/WEEK/MONTH token, cost, and request-count rate limit windows from a floating (rolling) model to fixed calendar windows anchored to a new deployment-wide `rateLimitSchedule` setting (timezone, week start day, daily reset time). MINUTE/HOUR keep their existing floating behavior unchanged.

### Applicable issues

- fixes #1951

### Description of changes

- Added `rateLimitSchedule` top-level config setting (`config/RateLimitSchedule.java`, `WeekDay.java`, `@ValidTimezone` validator) — defaults to UTC/Monday/00:00 if omitted, reproducing today's implicit UTC-midnight day/month behavior.
- Added `CalendarPeriod`/`CalendarWindowCalculator` for DST-safe (`ZonedDateTime`-based) computation of DAY/WEEK/MONTH period boundaries.
- Added `FixedRateBucket`/`CostFixedRateBucket` (`periodStart` + `count`), replacing the sliding-window `RateBucket`/`CostRateBucket` for DAY/WEEK/MONTH in `TokenRateLimit`, `CostRateLimit`, and the request-count limit's DAY tier; `RateWindow` is trimmed to MINUTE/HOUR only. Both new bucket types tolerate old floating-window JSON (`ignoreUnknown = true`), so previously-persisted counters are implicitly reset to zero on first read after rollout — no explicit migration step.
- Added `resetsAt` (ISO-8601 offset instant) to `ItemLimitStats`/`CostItemLimitStats`, populated only for the fixed windows (DAY/WEEK/MONTH) and omitted (not null) for MINUTE/HOUR, computable even for a deployment with zero usage.
- `RateLimiter` now takes a `ConfigStore` dependency so `increase()` can resolve the schedule internally without needing a `ProxyContext`.
- Updated `docs/dynamic-settings/roles.md`, regenerated `docs/open_api_core.yaml` via the project's own OpenAPI generator, and added a `rateLimitSchedule` example to `sample/aidial.config.json`.
- This is a hard cutover per the issue: no config toggle to restore the old floating day/week/month behavior.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [34593692685](https://github.com/epam/ai-dial-ci/actions/runs/34593692685)

| [deploy-review](https://chat-ai-dial-core-pr-1954.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34593692685/job/103244498304) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/34593692685/job/103246975965) |

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1961, opened 2026-09-11
feat: support overridePaths in deployment interfaces #1969

### LINKED ISSUE #1969: Extend `interfaces` config with `overridePaths`
### Name and Version

dial core 0.47.1

Today DIAL Core builds the upstream URL for a deployment configured with `baseUrl` / `interfaces.<type>.base_url` by concatenating that root with **the exact ingress path the request arrived on**:

```
POST {core}/openai/deployments/{id}/chat/completions  ->  POST {baseUrl}/openai/deployments/{id}/chat/completions
POST {core}/openai/v1/responses                       ->  POST {baseUrl}/openai/v1/responses
POST {core}/anthropic/v1/messages                     ->  POST {baseUrl}/anthropic/v1/messages
```

That locks Core to targets whose paths follow the DIAL API contract. A service that serves the same APIs at its own paths — for example `/v1/chat/completions`, `/v1/responses`, `/v1/messages` — cannot be fronted at all, because there is no way to configure the path Core calls.

Add an `overridePaths` map to an `interfaces` entry: keys name a Core API operation, values are the path applied to the base URL instead of the ingress path.

```json
{
  "applications": {
    "gpt-custom-paths": {
      "overrideName": "custom-gpt",
      "baseUrl": "https://custom-provider/",
      "interfaces": {
        "openaiChatCompletions": {
          "overridePaths": { "postAzureOpenaiChatCompletions": "/v1/chat/completions" }
        },
        "openaiResponses": {
          "overridePaths": {
            "postOpenaiResponses": "/v1/responses",
            "getOpenaiResponsesById": "/v1/responses/{id}",
            "deleteOpenaiResponsesById": "/v1/responses/{id}",
            "postOpenaiResponsesCancel": "/v1/responses/{id}/cancel"
          }
        },
        "anthropicMessages": {
          "overridePaths": {
            "postAnthropicMessages": "/v1/messages",
            "postAnthropicMessagesCountTokens": "/v1/messages/count_tokens"
          }
        }
      }
    }
  }
}
```

Key naming follows `{http_method}{api_provider}{api_type}`:

| Key | Core operation |
|---|---|
| `postAzureOpenaiChatCompletions` | `POST /openai/deployments/{id}/chat/completions` |
| `postAzureOpenaiEmbeddings` | `POST /openai/deployments/{id}/embeddings` |
| `postOpenaiResponses` | `POST /openai/v1/responses` |
| `getOpenaiResponsesById` | `GET /openai/v1/responses/{id}` |
| `deleteOpenaiResponsesById` | `DELETE /openai/v1/responses/{id}` |
| `postOpenaiResponsesCancel` | `POST /openai/v1/responses/{id}/cancel` |
| `postAnthropicMessages` | `POST /anthropic/v1/messages` |
| `postAnthropicMessagesCountTokens` | `POST /anthropic/v1/messages/count_tokens` |

Path values substitute two tokens:

- `{id}` — the id the operation addresses: the deployment name for the deployments-POST family, the response id for Responses item operations.
- `{overrideName}` — the deployment's `overrideName`, falling back to its name.

Expected behaviour:

1. An operation with an `overridePaths` entry is routed to `baseUrl` + the configured path.
2. An operation without one keeps today's routing (`baseUrl` + ingress path), so overriding a subset of an interface's operations is fine.
3. Applies to models, applications and interceptors.

Configuring the full URL per operation through the legacy `endpoint`/`responsesEndpoint` fields: those are forwarded verbatim, but there is only one field per interface family, so they cannot express the Responses item operations (`GET`/`DELETE`/`cancel` by id) or `count_tokens`, and they bypass the `interfaces` routing model rather than extending it.

khshanovskyi:
Adds `interfaces.<type>.overridePaths` so DIAL Core can front services that do not follow the DIAL API path contract (e.g. Switchyard serving `/v1/chat/completions`): a per-operation path replaces the default "base URL + ingress path" routing, with the exact tokens `{id}` and `{overrideName}` substituted. Applies to models, applications and interceptors.

### Applicable issues

- fixes #1969

### Description of changes

- `DeploymentInterface` gains `overridePaths` (`overridePaths`/`override_paths`), keyed by the new `OverridePathKey` enum: `postAzureOpenaiChatCompletions`, `postAzureOpenaiEmbeddings`, `postOpenaiResponses`, `getOpenaiResponsesById`, `deleteOpenaiResponsesById`, `postOpenaiResponsesCancel`, `postAnthropicMessages`, `postAnthropicMessagesCountTokens`. Unknown key names are tolerated for forward compatibility, mirroring unknown interface types.
- New `PathTemplateUtil` substitutes exactly the literal tokens `{id}` and `{overrideName}`; any other brace usage (typos like `{Id}`, unknown placeholders, literal braces) is rejected at config load by a dry-run render, keeping the brace syntax reserved for future tokens.
- `DeploymentEndpointUtil.resolveRequestUri` consults the override for the deployments-POST family (chat completions, embeddings, responses create, anthropic messages/count_tokens); a new `resolveResponseItemUri` covers responses GET/DELETE/cancel by id and replaced the three call sites that concatenated onto `resolveResponsesBaseUri` (item controller, interceptor hop, background poll). Overrides apply only in the interfaces/base-url flow — legacy `endpoint`/`responsesEndpoint` and translator-served interfaces are untouched, and behavior without `overridePaths` is byte-identical.
- `{id}` renders the deployment's name for the deployments-POST family (equal to the path id on direct calls, and still correct on the interceptor callback hop where the raw segment is the pseudo-id `interceptor`) and the hop's response id for responses item operations; `{overrideName}` renders the URL-encoded `overrideName`, falling back to the deployment name.
- Config validation: misplaced keys, malformed templates, `{id}` where the operation carries none, and overrides on `mode: translator` entries are rejected at load for models, applications and interceptors — including the `MergedConfigStore` partial-update write path, so a write cannot accept an entity the next full rebuild would drop.
- Tests: unit coverage in `PathTemplateUtilTest`, `DeploymentEndpointUtilTest`, `ConfigPostProcessorTest`; full-stack `OverridePathsApiTest` proving requests on standard Core paths reach the mock upstream on the overridden paths for chat completions (with `{overrideName}`), embeddings, anthropic messages/count_tokens, responses create + GET-by-id, an application, and unoverridden cancel/delete keeping default routing.
- Docs: new "Override paths per interface" section in `docs/dynamic-settings/models.md`, with pointers from `applications.md` and `interceptors.md`.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -180,6 +186,49 @@ static void validateSingleTranslator(Config config, String mapKey,
         onSkip.accept(ResourceTypes.TRANSLATOR, new InvalidEntityException(ResourceTypes.TRANSLATOR, mapKey, warnings));
     }
 
+    /**
+     * Targeted per-type helpers for {@link MergedConfigStore}'s partial-update path. Set the name from
+     * the map key and validate {@code overridePaths} — the one semantic check these types carry — so a
+     * write cannot accept an entity the next full rebuild would reject.
+     */
+    static void validateSingleApplication(Config config, String mapKey,
```
khshanovskyi, line 191: @KirylKurnosenka 
hi, can you check if these changes are fine with you (in this class)?

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -421,6 +470,79 @@ public static void validateDeploymentInterfaces(Model model, Map<String, Transla
...
+            validateOverridePaths(entry.getKey(), declared, field, warnings);
+        }
+    }
+
+    /** Reject malformed templates before an API write persists or encrypts the deployment. */
+    public static void requireValidOverridePaths(Deployment deployment) {
+        List<ValidationWarning> warnings = new ArrayList<>();
+        validateOverridePaths(deployment, warnings);
+        if (!warnings.isEmpty()) {
+            throw new HttpException(HttpStatus.UNPROCESSABLE_ENTITY,
```
KirylKurnosenka: I don't think `ConfigPostProcessor` should know anything about `http` by throwing `HttpException`

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
```diff
@@ -421,6 +470,79 @@ public static void validateDeploymentInterfaces(Model model, Map<String, Transla
...
+                    new JsonObject().put("validationWarnings", warnings).encode());
+        }
+    }
+
+    /**
+     * Validates every interface entry's {@code overridePaths} on a deployment whose interfaces get no
+     * other semantic validation — applications and interceptors. Models run the same per-entry check
+     * inside {@link #validateDeploymentInterfaces}.
+     */
+    public static void validateOverridePaths(Deployment deployment, List<ValidationWarning> warnings) {
```
KirylKurnosenka, line 481: can be `private`

config/src/main/java/com/epam/aidial/core/config/OverridePathKey.java
```diff
@@ -0,0 +1,85 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
+import jakarta.annotation.Nullable;
+import lombok.Getter;
+
+/**
+ * Core API operations whose upstream path an {@code interfaces.<type>.overridePaths} entry may
+ * replace. The string {@link #value} is the key used in {@link DeploymentInterface#getOverridePaths()}.
+ */
+@Getter
+public enum OverridePathKey {
```
astsiapanay: The enum's name is too generic.

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigValidationService.java
```diff
@@ -137,7 +141,7 @@ public ValidationResult validateOnly(AdminManifest entry, Config scratch) {
                     }
                 }
             }
-        } catch (IllegalArgumentException ex) {
+        } catch (IllegalArgumentException | HttpException ex) {
```
KirylKurnosenka: This service should know nothing about `http`

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
khshanovskyi: Agreed — fixed in 9d81e7b0. `requireValidOverridePaths` is gone; `ConfigPostProcessor` only accumulates `ValidationWarning`s now and no longer imports `HttpException`/`HttpStatus`/`JsonObject`. Each write surface maps them in its own idiom: `EntityResult`/`ValidationResult` in the admin services, and the existing `validationWarnings` 422 in `ConfigResourceController` (that body building is now shared by `checkCrossReferences`, `checkTranslator` and the new `checkOverridePaths`).

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigValidationService.java
khshanovskyi: Fixed in 9d81e7b0 — the `HttpException` import and the `catch (IllegalArgumentException | HttpException)` are gone. Override-path defects now return `ValidationResult(FAILED, ...)` like every other check here: folded into the existing warnings list for models, and via a small `describeOverridePathWarnings` helper for interceptors/applications.

server/src/main/java/com/epam/aidial/core/server/config/ConfigPostProcessor.java
khshanovskyi: You were right at the time — the only external caller was the public `requireValidOverridePaths` wrapper. Removing that wrapper (your other comment) inverted it: `validateOverridePaths(Deployment, List<ValidationWarning>)` is now the API the four write surfaces call, so it has to stay public. It is the same shape as its siblings here — `validatePricing`, `validateUpstreamInterfaces`, `validateTranslator`, `validateCrossReferences` are all public static void with a warnings accumulator. The per-entry overload it delegates to is private.

config/src/main/java/com/epam/aidial/core/config/OverridePathKey.java
astsiapanay: InterfacePathMapping?

khshanovskyi: Renamed in b77c481a — `OverridePathKey` -> `InterfacePathMapping`, along with the identifiers named after it (`pathKey` -> `pathMapping`, `findRequestPathKey` -> `findRequestPathMapping`, `Operation.key` -> `Operation.pathMapping`). The JSON key names inside `overridePaths` are unchanged, so no config or docs are affected.

astsiapanay (dismissed):

KirylKurnosenka (dismissed):

KirylKurnosenka (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1964, opened 2026-09-11
fix: NPE in streaming chat-completions function chain #1963

### LINKED ISSUE #1963: NPE in streaming chat-completions function chain: "Function call is failed with error. Try to recover SSE event"
## Summary

Streaming `/chat/completions` requests intermittently log:

```
Function call is failed with error. Try to recover SSE event
java.lang.NullPointerException
```

from `com.epam.aidial.core.server.vertx.stream.BufferingReadStream$BaseEventListener`. The exception has a `null` message and an empty stack trace, which is consistent with the JVM's `OmitStackTraceInFastThrow` optimization kicking in — meaning this is a *hot*, frequently-firing NPE, not a rare edge case, that has been failing silently because the generic `.recover()` swallows it and simply forwards the raw upstream SSE event unchanged.

## Root cause

`CollectResponseAttachmentsFn.apply()` returns a future with a **null value** for the common case of "no attachments in this chunk":

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/function/CollectResponseAttachmentsFn.java#L40-L42
```java
if (permittedAttachments.isEmpty()) {
}
```

This function is chained via `Future.compose(...)` in `BufferingReadStream.BaseEventListener.handle()`:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java#L268-L289

and wired for `/chat/completions` as:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/controller/DeploymentPostController.java#L417-L421
```java
new ChatCompletionSseListener(isChatCompletionsPath()
        ? List.of(new StripUsagePerModelFn(proxy, context), new CollectResponseChatCompletionAttachmentsFn(proxy, context),
                new CollectChatCompletionUsageFn(proxy, context))
        : List.of(new CollectResponseChatCompletionAttachmentsFn(proxy, context)));
```

The function right after it, `CollectChatCompletionUsageFn.apply()`, unconditionally dereferences the tree with no null check:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/function/CollectChatCompletionUsageFn.java#L27-L29
```java
public Future<JsonNode> apply(JsonNode tree) {
    usage = MergeChunks.merge(usage, tree.get("usage"));
```

So on any chunk with no attachments (i.e. almost every chunk), `tree` arrives as `null` here and `tree.get("usage")` throws immediately.

## Why this is a regression, not a long-standing bug

`BufferingReadStream.BaseEventListener.send()` explicitly accepts `@Nullable JsonNode tree` and treats `null` as "re-emit the original SSE event unchanged" — the *only* place in the pipeline designed to tolerate a null tree, since it's the terminal sink of the `compose` chain:

https://github.com/epam/ai-dial-core/blob/development/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java#L296-L305

Before #1858 (commit c17ca3218b, "feat: cache pricing decision tree for provider TTL/service-tier rate matrices"), the `/chat/completions` chain was only `StripUsagePerModelFn → CollectResponseChatCompletionAttachmentsFn`, and `CollectResponseChatCompletionAttachmentsFn` was the **last** function in the chain — so its occasional null return flowed straight into the null-tolerant `send()`, which was safe by construction. `StripUsagePerModelFn.apply()` always returns the tree unchanged, and `CollectResponseChatCompletionAttachmentsFn`'s own `collectAttachments()` override is fully null-guarded internally, so neither of those ever threw on their own account.

#1858 inserted `CollectChatCompletionUsageFn` *after* `CollectResponseChatCompletionAttachmentsFn`, turning what used to be "the chain's final output" into "an intermediate value fed into another function's `apply()`" — and that function assumes a non-null input.

## Impact

Beyond the log noise, the swallowed exception means `context.setPricingUsageNode(...)` in `CollectChatCompletionUsageFn` silently fails to update for most streaming chunks — including, plausibly, the terminal chunk that actually carries `usage` — so accumulated `usage`/`service_tier`/`custom_fields` used for pricing may be lost for affected streaming responses.

## Suggested fix

1. In `CollectResponseAttachmentsFn.apply()`, change the no-attachments early return to `Future.succeededFuture(tree)` (matching the sibling early return a few lines above it), so the tree is always passed through unchanged rather than nulled out.
2. Defensively guard the top of `CollectChatCompletionUsageFn.apply()` against a null `tree`, so a similar mistake elsewhere in the chain can't regress this again.
3. Consider guarding `BufferingReadStream.BaseEventListener.handle()` against a null/blank `event.getData()` before calling `ProxyUtil.MAPPER.readTree(data)` (independent hardening, not the cause of this particular report).

## Test gap

There is no unit/integration test that exercises the real 3-function chain used by `DeploymentPostController` (`StripUsagePerModelFn → CollectResponseChatCompletionAttachmentsFn → CollectChatCompletionUsageFn`) together — each function is currently only tested in isolation, which is how this regression slipped through.

astsiapanay:
Fixes a null-pointer exception in the streaming `/chat/completions` response-function chain that silently fires on nearly every streaming chunk with no attachments, logged as "Function call is failed with error. Try to recover SSE event" and swallowed by generic error recovery.

### Applicable issues

- fixes #1963

### Description of changes

- `CollectResponseAttachmentsFn.apply()` now returns `Future.succeededFuture(tree)` instead of a null-valued future when a chunk has no attachments, so the tree is passed through unchanged instead of dropped.
- `CollectChatCompletionUsageFn.apply()` now guards against a null `tree` input, preventing a similar mistake elsewhere in the chain from causing the same regression.
- `BufferingReadStream.BaseEventListener.handle()` now guards against a null/blank SSE `data` payload before parsing it as JSON.
- Added a regression test (`CollectResponseChatCompletionAttachmentsFnTest#testApplyPassesTreeThroughUnchangedWhenNoAttachments`) and a null-tolerance test (`CollectChatCompletionUsageFnTest#testTolerantOfNullTree`), plus a new `ChatCompletionFunctionChainTest` that exercises the real 3-function chain used by `DeploymentPostController` end-to-end, closing the test gap that let this regression through originally (each function was previously only tested in isolation).

### Checklist

### DIFF
--- a/server/src/main/java/com/epam/aidial/core/server/function/CollectChatCompletionUsageFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/CollectChatCompletionUsageFn.java
@@ -26,6 +26,10 @@ public class CollectChatCompletionUsageFn extends BaseResponseFunction {
 
     @Override
     public Future<JsonNode> apply(JsonNode tree) {
+        if (tree == null) {
+            return Future.succeededFuture(tree);
+        }
+
         usage = MergeChunks.merge(usage, tree.get("usage"));
         if (tree.get("service_tier") != null) {
             serviceTier = tree.get("service_tier");
--- a/server/src/main/java/com/epam/aidial/core/server/function/CollectResponseAttachmentsFn.java
+++ b/server/src/main/java/com/epam/aidial/core/server/function/CollectResponseAttachmentsFn.java
@@ -38,7 +38,7 @@ public abstract class CollectResponseAttachmentsFn extends BaseResponseFunction 
                 processAttachedFile(attachment, permittedAttachments);
             }
             if (permittedAttachments.isEmpty()) {
-                return Future.succeededFuture();
+                return Future.succeededFuture(tree);
             }
             return proxy.getTaskExecutor().submit(() -> {
                 proxy.getApiKeyStore().updatePerRequestApiKey(perRequestKey, json -> updateAutoSharedAttachments(json, permittedAttachments, perRequestKey));
--- a/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java
+++ b/server/src/main/java/com/epam/aidial/core/server/vertx/stream/BufferingReadStream.java
@@ -267,10 +267,10 @@ public class BufferingReadStream implements ReadStream<Buffer> {
         @SneakyThrows
         private Future<Void> handle(SseEvent event) {
             Future<JsonNode> result;
-            if (functions.isEmpty() || skipEvent(event)) {
+            String data = event.getData();
+            if (functions.isEmpty() || skipEvent(event) || data == null || data.isBlank()) {
                 result = Future.succeededFuture();
             } else {
-                String data = event.getData();
                 try {
                     JsonNode tree = ProxyUtil.MAPPER.readTree(data);
                     result = Future.succeededFuture(tree);
--- a/server/src/test/java/com/epam/aidial/core/server/function/ChatCompletionFunctionChainTest.java
+++ b/server/src/test/java/com/epam/aidial/core/server/function/ChatCompletionFunctionChainTest.java
@@ -1,0 +1,75 @@
+package com.epam.aidial.core.server.function;
+
+import com.epam.aidial.core.server.ProxyContext;
+import com.epam.aidial.core.server.data.ApiKeyData;
+import com.epam.aidial.core.server.util.ProxyUtil;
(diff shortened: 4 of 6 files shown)


ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

PR #1972, opened 2026-09-14
fix: NPE in ResourceService when blob contentLength is null #1971

### LINKED ISSUE #1971: NPE in ResourceService.toFileMetadata when blob contentLength is null (complex-resource sweep)
## Summary
`ComplexResourceSweepService` crashes with an NPE when reclaiming a "deleting" marker
resource, because `ResourceService.toFileMetadata` unboxes a nullable `Long` into a
primitive `long`.

## Observed error
```
Failed to process complex resource ref: complex_resource_refs/9a4c8327c5d6488aa7feb35903cf486b.json
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of
"com.epam.aidial.core.storage.service.ResourceService$Result.contentLength()" is null
	at com.epam.aidial.core.storage.service.ResourceService.toFileMetadata(ResourceService.java:504)
	at com.epam.aidial.core.storage.service.ResourceService.getResourceMetadata(ResourceService.java:490)
	at com.epam.aidial.core.storage.service.ResourceService.deleteResource(ResourceService.java:919)
	at com.epam.aidial.core.storage.service.ResourceService.deleteResource(ResourceService.java:911)
	at com.epam.aidial.core.server.service.resource.ComplexResourceService.reclaimDeletingResource(ComplexResourceService.java:886)
	at com.epam.aidial.core.server.service.resource.ComplexResourceSweepService.processRef(ComplexResourceSweepService.java:160)
```
Version: `aidial-core 0.48.0-dev.45`. Blob storage provider: GCP (Google Cloud Storage), jclouds `google-cloud-storage` provider (jclouds 2.7.3).

## Root cause
`ResourceService.Result.contentLength` is a boxed `Long` (`ResourceService.java` ~line 1305)
that is nullable by contract when derived from blob-store metadata
(`BlobMetadata.getContentMetadata().getContentLength()`, populated in `blobToResult`,
~line 1081) or from a Redis hash missing the `content_length` field (`toResult`, ~line 1164).

`toFileMetadata` (`ResourceService.java:502-509`) passes `result.contentLength()` directly
into `FileMetadata`'s constructor, whose parameter is a **primitive** `long`
(`storage/.../data/FileMetadata.java`), so a null value throws on unboxing:

```java
private static FileMetadata toFileMetadata(
        ResourceDescriptor resource, Result result) {
    return (FileMetadata) new FileMetadata(resource, result.contentLength(), result.contentType())
            ...
}
```

A second, same-shape bug exists at `ResourceService.java:607-609` (`getResourceStream`):
```java
Long length = metadata.getContentMetadata().getContentLength();
...
if (length <= maxSizeToCache) {   // NPE if length is null
```

### Why it surfaces via the sweep specifically
`FILE` and `SKILL` are the only `ResourceTypes` with `requireCompression=false`
(`ResourceTypes.java`), i.e. the only types that go through `toFileMetadata`. Both are
cached in Redis for 5 minutes. "Complex resources" only exist for `SKILL`
(`ComplexResourceController` registers `Map.of(ResourceTypes.SKILL, new SkillHandler())`
only). The marker resource (`.dial-resource`) is a normal small JSON document written via
the same `ResourceService.putResource` path any small `FILE` upload uses — not a special
placeholder — though on creation it briefly exists as a zero-byte "stub" blob before an
async `sync()` flushes the real bytes.

`ComplexResourceSweepService`'s `gracePeriod` (1 hour, `aidial.settings.json`) is deliberately
much longer than the FILE/SKILL Redis TTL (5 minutes), so `reclaimDeletingResource` →
`deleteResource` → `getResourceMetadata` is **guaranteed** to run after the Redis cache has
expired, forcing a cold blob-store metadata call (`ResourceService.blobGet(key, false)` →
`BlobStorage.meta()` → jclouds `blobStore.blobMetadata(bucket, key)`) on every reclaim.
Ordinary user-driven `FILE` deletes happen shortly after upload/access while Redis is still
warm, so they rarely take this cold path — but the same NPE is reachable there too under the
same timing conditions. This is a general latent bug in the shared `ResourceService`, not
something specific to skills; the sweep job's design (long, deliberate grace period) is just
what reliably exposes it.

The exact GCS-side condition that leaves `getContentLength()` null on this particular
metadata call wasn't pinned down with certainty from static analysis of the jclouds provider
jar — worth checking with whoever owns the GCS blob-store integration, but not required to
fix the immediate crash, since the contract already documents the field as nullable.

## Suggested fix
Guard both unboxing sites in `storage/src/main/java/com/epam/aidial/core/storage/service/ResourceService.java`:
- `toFileMetadata` (line ~504): default to `0L` when `contentLength()` is null.
- `getResourceStream` (line ~607-609): treat a null length as "not cacheable" rather than comparing `null <= maxSizeToCache`.

astsiapanay:
Fixes an NPE where `ResourceService.toFileMetadata`/`getResourceStream` unboxed a nullable blob-store `contentLength` into a primitive `long`.

### Applicable issues

- fixes #1971

### Description of changes

- `toFileMetadata`: defaults to `0L` when `Result.contentLength()` is null instead of unboxing directly.
- `getResourceStream`: treats a null blob-metadata content length as "read fully" (routes through the existing cache/full-read path, which derives length from the actual bytes) instead of comparing `null <= maxSizeToCache`.
- Added regression tests in `ResourceServiceTest` that simulate a blob backend returning a null content length for both code paths; verified they fail without the fix and pass with it.

### Checklist


Oleksii-Klimov (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

Oleksii-Klimov (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1977, opened 2026-09-14
fix: accept interfaces as the only routing target for applications

### LINKED ISSUE #1965: Application write API rejects `interfaces` as the only routing target
### Problem

An application written through the Configuration API is rejected with **HTTP 400** when `interfaces` is its only routing target:

```
At least application endpoint, MCP endpoint or function must be provided
```

The check in `ApplicationService` (~line 597) predates the typed `interfaces` map and was never taught about it — it only recognises `endpoint`, `function` and `mcp.endpoint`:

```java
} else if (application.getEndpoint() == null && application.getFunction() == null
        && (application.getMcp() == null || application.getMcp().getEndpoint() == null)) {
    throw new IllegalArgumentException("At least application endpoint, MCP endpoint or function must be provided");
}
```

### Reproduction

```
PUT /v1/applications/public/my-app
{
  "baseUrl": "http://localhost:4848",
  "interfaces": {"openaiChatCompletions": {}}
}
```

→ `400 At least application endpoint, MCP endpoint or function must be provided`

### Scope

* **Write path only.** The same application shape loaded from the config file works: it routes through `interfaces` normally, and is covered by `ApplicationInterfacesApiTest`.
* **Applications only.** Models and interceptors accept `interfaces`-only specs on the same blob write surface and route correctly.
* Adding a legacy `endpoint` next to `interfaces` is a workaround — `interfaces` takes precedence for routing, so the `endpoint` value is never used.

### Suggested fix

Treat a resolvable `interfaces` entry as a valid routing target in that validator, e.g. via `DeploymentEndpointUtil.isInterfaceDeclared` / `resolveServingEndpoint`, so the API write surface accepts the same application shapes the config file does.

### Notes

Pre-existing — the same line is on `development`, unrelated to any in-flight `interfaces`/`overridePaths` work. Found while verifying that `overridePaths` applies to blob-persisted deployments (it does, for all three deployment types).

khshanovskyi:
### Description of changes

- The application write API rejected applications whose only routing target is `interfaces` with `400 At least application endpoint, MCP endpoint or function must be provided` — a check in `ApplicationService` that predates the typed `interfaces` map, while the same shape loaded from the config file routes normally.
- `DeploymentEndpointUtil.hasRoutingInterface(deployment)`: the `interfaces` map contributes a servable type — an entry carrying a translator reference or a url of its own (the deployment-level `baseUrl` included), or one whose type a legacy field also serves. An entry resolving to nothing is not a routing target, mirroring the 503 the request path answers for that same shape.
- `ApplicationService.prepareApplication` now accepts such applications, and the validation message names the interface option.
- Tests: write-path regression in `ApplicationInterfacesApiTest` (an interfaces-only application is stored and serves through the interface; an entry without any url is still rejected) plus branch coverage for `hasRoutingInterface` in `DeploymentEndpointUtilTest`.

Fixes #1965

### Checklist

- [x] All tests pass (ApplicationInterfacesApiTest 5/5 including the new write-path regression, DeploymentEndpointUtilTest, application write/serving suites, checkstyle clean; full suite green apart from the known Redis-port flake, which passes on rerun)


ai-dial-actions:
<h1>Dependency Review</h1>
None

KirylKurnosenka (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1980, opened 2026-09-14
fix: reject key writes whose secret is already used by another key entity #1979

### LINKED ISSUE #1979: Admin API: key entities with the same secret silently overwrite each other
## Name and Version

AI DIAL Core

## What steps will reproduce the bug?

1. Create a key entity via the admin API:
   ```
   PUT /v1/keys/platform/key-a
   { "key": "same-secret", "project": "projectA", "roles": [...] }
   ```
   → 200 OK
2. Create a second key entity with a different name but the **same plaintext secret**:
   ```
   PUT /v1/keys/platform/key-b
   { "key": "same-secret", "project": "projectB", "roles": [...] }
   ```
   → 200 OK — no duplicate-secret check exists on any write path (PUT, `/v1/admin/apply`, `/v1/admin/validate`).
3. Send a request with `Api-key: same-secret`.

## What is the expected behavior?

A key write whose secret is already used by a different key entity is rejected, the same way a duplicate `$id` is rejected for schemas. `key-a` keeps its roles and project, and requests authenticated with the secret are attributed to it.

## What do you see instead?

Both writes succeed, but the auth store is keyed by the plaintext secret, so the second entity silently overwrites the first. Authentication with the secret resolves to whichever entity was written last, flipping its roles, rate-limit project, and author attribution. Anyone who knows a secret can register a second key entity with broader roles and take over the identity behind that secret. The same happens when a blob key reuses the secret of a file-defined key from `aidial.config.json`.

## Additional information

Fix in progress on branch `fix/key-secret-uniqueness`: reject a key write whose secret is already used by another key entity (409 on PUT, FAILED per entity on apply/validate), mirroring the schema duplicate-`$id` guard; the config-file migration endpoint remains the supported handoff for file-defined keys.

KirylKurnosenka:
ApiKeyStore.keys is keyed by the plaintext secret, so two Key entities created with the same secret silently collapsed: auth, roles, rate-limit project, and author attribution flipped to whichever entity was written last.

### Applicable issues

- fixes #1979

### Description of changes

- Strict duplicate-secret guard on all key write paths, mirroring the schema duplicate-`$id` guard. Shared checks live in `KeyValidator` (required fields, project/role, secret-collision), which every write path calls directly: `ConfigResourceController` (single PUT), `ConfigApplyService`/`ConfigValidationService` (`/v1/admin/apply` and `/v1/admin/validate`), and `ApiKeyStore`'s full-rebuild loader — replacing three near-duplicate implementations of the same checks. The collision predicate itself (`KeyValidator.isKeySecretChangedAndTakenByAnotherKey`) scans the folded `Config.keys` map, skipping the written entity's own entry.
- `PUT /v1/keys/platform/{name}` returns 409 before anything is written; `/v1/admin/apply` and `/v1/admin/validate` mark the entity FAILED (within-batch duplicates are caught too, via the scratch).
- The config-file migration endpoint pre-removes the file twin from the scratch (`scratch.getKeys().remove(secret)`), the same handoff `scratchSchemas.remove(id)` gives schemas — migration stays the only supported file→blob handoff for keys.
- Updates that keep the entity's own prior secret stay allowed, so a migrated key can still be updated with its unchanged secret.
- Replica events log a warning instead of rejecting (the blob is already written on the origin pod).
- Rebuild-time detection of legacy blob-vs-blob duplicate pairs is out of scope.
- New tests: 409/duplicate-secret cases on PUT, apply, and validate (including rotation, unchanged-secret update, and file-key-secret cases), plus a dedicated `KeyValidatorTest` (parameterized) covering the shared validator directly; migration and `ApiKeyStore`/`MergedConfigStore` regression suites verified locally.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (dismissed):

ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

PR #1988, opened 2026-09-16
fix: preserve encrypted secrets when migrating models and routes #1784

### LINKED ISSUE #1784: Slice D — Config-file migration endpoint POST /v1/admin/config/file/migrate

KirylKurnosenka:
`ConfigFileMigrateController` built its per-entity manifest spec with the response mapper, which honors `@JsonProperty(WRITE_ONLY)` on every `@EncryptedField` and drops it on serialization. Every migrated model/route landed in the `platform` bucket with its upstream `key`/`secretExtraData` silently missing — reported as `migrated` with no error, and never self-healing (a repeat run reports `skipped: already in blob`).

### Applicable issues

- Closes #1784

### Description of changes

- `ConfigFileMigrateController#collectManaged`: swap `ProxyUtil.MAPPER` to `BLOB_MAPPER` when building the manifest spec node — the config-write path already uses `BLOB_MAPPER` for this exact reason. `collectKeys` is unaffected; a file key's secret is the map key, not an object field.
- `ConfigApplyService#applyManagedEntity` (INTERCEPTOR/ROLE/ROUTE): add `secretFieldProcessor.encryptFields`/`decryptFields` around the blob put, matching `applyKey`/`applyModel`'s existing pattern. This closes the same gap one level up — a plain `/v1/admin/apply` on a route lost upstream secrets the same way, independent of migration. Interceptors/roles carry no `@EncryptedField` so this is a no-op for them; routes are what it fixes.
- New tests: `ConfigFileMigrateApiTest#testMigrateModelPreservesUpstreamSecrets` / `#testMigrateRoutePreservesAndEncryptsUpstreamSecret` assert the raw blob is `ENC[`-encrypted with no plaintext, the merged in-memory `Config` holds the decrypted plaintext, and GET leaks neither; `AdminApplyApiTest#testApplyRouteEncryptsUpstreamSecretAtRest` covers the same for `/v1/admin/apply` directly.

### Known related gap, out of scope here

Applications have the same underlying problem on every write path (not just these two), via a separate defect in `ApplicationService`'s serialization (plain mapper, no `SecretFieldProcessor`) plus `SecretFieldProcessor`'s non-transitive encrypted-field detection not descending into `Application.routes`. Fixing it also requires auditing every application copy/move path (publication copy, function-bucket copy) for AAD consistency, since ciphertext is bound to the blob's path. Tracked as #1989 rather than folded into this fix.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #1990, opened 2026-09-17
fix: name the model/deployment in config validation warnings

KirylKurnosenka:
Config validation warnings were opaque when read out of context (e.g. `FileConfigStore`'s reload log printed only the bare warning text, with no indication of which model/deployment/translator caused it). Each `ValidationWarning` message created inside `ConfigPostProcessor` now names the entity it's about, wherever that name is reliably known at validation time.

### Description of changes

- Added `ConfigPostProcessor.messageWithEntityPrefix(typeLabel, name, message)`, a null-safe helper that prefixes a warning message with `"<Type> '<name>': "` when the entity's name is known, and leaves the message untouched otherwise.
- Applied it to every warning created in `validatePricing`, `validateUpstreamInterfaces`, `validateCrossReferences`, `validateDeploymentInterfaces` (+ its private helpers `validateTranslatedInterface`/`validateTranslatorOutput`), and the `validateOverridePaths` override-path checks (shared by models/applications/interceptors, using `deployment.getClass().getSimpleName()` for the type word).
- `validateTranslator` gained an explicit `name` parameter (translators carry no name field of their own), threaded from its 5 callers using the identifier each already has locally (`name`/`mapKey`/`path`/`id`).
- Scoped to call sites where the entity's name is reliably set before validation runs (verified across `ConfigPostProcessor`'s own internal callers); the 4 external call sites where the entity isn't named yet (`ConfigResourceController`, `ApplicationService`, `ConfigValidationService`, `ConfigApplyService`) are left unchanged — they already tag results with the entity's id separately, and the helper's null-guard means they keep getting the same unprefixed message as before.

### Checklist


ai-dial-actions:
<h1>Dependency Review</h1>
None

khshanovskyi (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #2000, opened 2026-09-18
fix: enum @JsonValue schema + config parse error messages #1999

### LINKED ISSUE #1999: Admin config API rejects documented enum values; parse errors give no location info
### Name and Version

ai-dial-core 0.47.1

1. Look up the `InterfaceType`, `InterfaceMode`, or `Operator` schema in `docs/open_api_core.yaml` (generated spec). It lists enum values as the Java constant names, e.g. `InterfaceType`: `OPENAI_CHAT_COMPLETIONS`, `OPENAI_EMBEDDINGS`, `OPENAI_RESPONSES`, `ANTHROPIC_MESSAGES`.
2. Submit an admin config entity that uses one of those documented values, e.g. a `Translator` with `"in": "ANTHROPIC_MESSAGES"`, via the admin config API (or in `translators`/`interfaces` in the config file).
3. Separately, submit any admin config entity with a malformed value, e.g. an unknown enum value or a missing required field, via the admin config API (e.g. `PUT /v1/admin/config/...`).

1. A value that the OpenAPI spec documents as valid for the field should be accepted (or the spec should document the value the API actually accepts).
2. The parse error should say which field failed and why, e.g. naming the entity type and field.

### What do you see instead?

1. The request is rejected. Core's actual wire format for these enums is `camelCase` (`openaiChatCompletions`) or the Operator's own symbol (`==`, `!=`, ...) — set via a Jackson `@JsonValue` field — not the documented upper-case constant name. Root cause: the OpenAPI generator's Jackson module (`jsonschema-module-jackson`) only detects `@JsonValue` when it sits on a method; these enums (`InterfaceType`, `InterfaceMode`, `Operator`) have it on a Lombok-backed field instead, so schema generation silently falls back to `Enum.name()` for the documented values instead of the real wire values.
2. Any malformed admin config entity produces the unhelpful message `Failed to parse entity at line -1, column -1`. The `-1, -1` location is always synthetic/meaningless here because the failure happens while converting an already-parsed JSON tree (`ObjectMapper.treeToValue`), which has no token-stream position to report — so the "unknown location" fallback never triggers and the location is always wrong instead.

### Additional information

Affects `config/src/main/java/com/epam/aidial/core/config/InterfaceType.java`, `InterfaceMode.java`, `Operator.java` (enum doc/serialization mismatch), `openapi-generator/src/main/java/com/epam/aidial/core/openapi/DtoSchemaGenerator.java` (schema generation), and `server/src/main/java/com/epam/aidial/core/server/service/config/ConfigEntityCodec.java` (parse error message).

KirylKurnosenka:
Fixes the OpenAPI spec listing wrong enum values for `InterfaceType`/`InterfaceMode`/`Operator`, and the unhelpful "Failed to parse entity at line -1, column -1" error from the admin config API.

### Applicable issues

- fixes #1999

### Description of changes

- `DtoSchemaGenerator`: added a custom definition provider that reads `@JsonValue` off an enum's backing field (not just a method, which is all `jsonschema-module-jackson` checks). `InterfaceType`, `InterfaceMode`, and `Operator` all annotate the field via a Lombok `@Getter`, so the generated OpenAPI schema was silently falling back to `Enum.name()` (e.g. `OPENAI_CHAT_COMPLETIONS`) instead of the real wire value (`openaiChatCompletions`). `docs/open_api_core.yaml` will be regenerated by CI.
- `ConfigEntityCodec.treeToEntity`: the error message now names the failing field (via `JsonMappingException.getPath()`) and reports the underlying reason, instead of a synthetic, always-invalid `line -1, column -1` (meaningless here since the failure happens converting an already-parsed JSON tree, which has no token-stream position). Also chains the original exception as the cause, and gives `UnrecognizedPropertyException` a cleaner message (drops Jackson's `(class ..., not marked as ignorable)` boilerplate).
- Added test coverage: `DtoSchemaGeneratorTest` for the field-based `@JsonValue` schema, `ConfigEntityCodecTest` for the new error messages.

### Checklist


openapi-generator/src/main/java/com/epam/aidial/core/openapi/DtoSchemaGenerator.java
```diff
@@ -133,6 +142,45 @@ public void registerExternalSchema(String schemaName) {
...
 
+    /**
+     * {@link JacksonOption#FLATTENED_ENUMS_FROM_JSONVALUE} only detects {@code @JsonValue} on a method
+     * (jsonschema-module-jackson's {@code CustomEnumDefinitionProvider} reads member <em>methods</em>
+     * exclusively); an enum whose {@code @JsonValue} sits on the backing field instead — the usual shape
+     * for a Lombok {@code @Getter} enum such as {@link com.epam.aidial.core.config.InterfaceType} — falls
+     * through to the default {@code Enum.name()}-based schema. This mirrors that provider's own logic, one
+     * step earlier: reading the field directly instead of invoking a method.
+     */
+    private CustomDefinition createFieldJsonValueEnumDefinition(ResolvedType javaType, SchemaGenerationContext context) {
```
khshanovskyi: I have some doubts that it is reasonable to have have such method (and reflection API in it) - can we use for this purpose the capabilities of the `jackson` lib? 
I'm not pushing to this point, but if we can - it would be much better that creating our own impl

KirylKurnosenka: Good call — swapped it for Jackson's own introspection: `ObjectMapper.getSerializationConfig().introspect(...).findJsonValueAccessor()` resolves the `@JsonValue` accessor (field or method) the same way Jackson does internally, so there's no more hand-rolled field scanning. Pushed in b03100f6.

Katerina-Charakhovich (approved):

khshanovskyi (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #2001, opened 2026-09-20
fix: validate model rebuild invariants before the blob write

siarhei-fedziukovich:
### Description of changes

**The failure.** Core refused to start:

```
InvalidEntityException: Model 'kk_test_model_interface': Interface 'openai_chat_completions'
  declares no base_url and the model declares no baseUrl
    at ConfigPostProcessor.processModels(ConfigPostProcessor.java:333)
    at MergedConfigStore.rebuild(MergedConfigStore.java:1255)
    at MergedConfigStore.init(MergedConfigStore.java:252)
    at AiDial.start(AiDial.java:259)
```

The model had been written through the admin API and persisted to blob storage, after which every rebuild — including startup — aborted.

**Root cause.** `ConfigPostProcessor.validateDeploymentInterfaces` was called on **zero** pre-write paths, but runs unconditionally at rebuild (`processModels`) and on the post-write in-memory partial update (`validateSingleModel`). Both write paths are blob-first, memory-second:

| Path | Blob write | In-memory validation | Result under `onInvalidEntity=abort` |
|---|---|---|---|
| Single `PUT` | `ConfigResourceController:1645` | `:1669` | **500**, blob persisted, no compensating delete |
| Admin batch apply | `ConfigApplyService:300`, per entry | once after the loop, `:121` | **200 `APPLIED`**, blob persisted, failure only logged |

So an entity the rebuild is guaranteed to reject could be made durable, and the caller was told either "server error" or — on the batch path — "applied".

The underlying problem is drift: the three pre-write sites each hand-copied a *subset* of what `processModels` runs, and the subsets diverged. The repo already states this hazard as a design rule in three places (`checkTranslator`, `checkOverridePaths`, `ApplicationService.requireRenderableOverridePaths`: *"refused before the application is written, so the blob never holds config the next config rebuild would drop"*). `checkModel` was the sole outlier.

**The fix.** Extract the invariant set once, as `ConfigPostProcessor#validateModelInvariants` (pricing, upstream interfaces, deployment interfaces), and route both rebuild call sites and all three pre-write sites through it — so a validator added to the rebuild cannot silently reopen the gap:

- `ConfigResourceController.checkModel` — single `PUT`
- `ConfigApplyService.applyModel` — admin batch apply
- `ConfigValidationService` — `/v1/admin/validate` precheck

Note `validateDeploymentInterfaces` already performs the per-entry override-path check for models, so the standalone `validateOverridePaths` call is **dropped** at the model sites rather than kept alongside it — keeping both would report every override-path warning twice.

**Soft-validation rule.** `softValidation` now admits exactly what the rebuild admits: cross-references only, which a later interceptor write repairs and `resurrectInvalidModels` reinstates. Pricing, upstream-interface and deployment-interface warnings are fatal (`422` / `FAILED`) on all three surfaces — previously `softValidation=true` sanctioned writes the rebuild discards unconditionally, a second poison-pill class.

**Batch ordering needs no accommodation.** `DEPENDENCY_ORDER` sorts `Translator` (2) ahead of `Model` (7) and `mutateScratch` stages each success, so an in-batch translator is already in `scratch` when the model is validated; manifest author order is irrelevant. A named translator with no registry entry stays tolerated — it leaves the interface unserved (503 at request time), not the model invalid.

**Out of scope**, deliberately: abort/skip rebuild semantics, `MergedConfigStore` rebuild logic, and `FileConfigStore`, which stays fail-fast. Three post-write gaps remain and are much harder to trigger after this, but are untouched: `handleWriteError` has no `InvalidEntityException` branch (so post-write failures still surface as 500), `applyBatch` failures are only logged, and neither path deletes the blob it already wrote.

### Behaviour change

`softValidation` defaults to `false`, so deployments are unaffected unless they opted in — and a deployment that opted in was writing models the rebuild discards, so the `422` surfaces a previously silent defect. One existing test encoded the old contract: `AdminValidateApiTest.testV19SoftValidationAdmitsCacheRateWithoutTokenUnit` is inverted and renamed to `...StillRejectsCacheRateWithoutTokenUnit`. Its sibling `testV17` (dangling interceptor ref) is unchanged and guards against over-tightening.

### Testing

Full `:server:test` — **2253 tests, 0 failures, 0 errors**. `checkstyleMain` / `checkstyleTest` clean.

18 new tests:
- `ConfigPostProcessorTest` — 4 unit tests on the invariant set, including one that fails if a validator is added to the rebuild but not to `validateModelInvariants`, and one pinning that override-path warnings are not duplicated.
- `ModelWriteApiTest`, `ModelCrossRefValidationSoftModeApiTest`, `AdminApplyApiTest`, `AdminValidateApiTest` — each asserting both the `422`/`FAILED` **and** a follow-up `404` proving the blob was never written. Includes the production regression shape, the in-batch translator ordering proof, and a guard that an unregistered translator name is still accepted.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigApplyService.java
```diff
@@ -279,14 +279,13 @@ private EntityResult applyKey(Key key, String id, ParsedName parsed, List<Entity
 
     private EntityResult applyModel(Model model, String id, ParsedName parsed, Config scratch, List<EntityChange> pending) {
         List<ValidationWarning> warnings = new ArrayList<>();
-        ConfigPostProcessor.validateOverridePaths(model, warnings);
```
KirylKurnosenka, line 282: Is this removed for purpose?

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigValidationService.java
```diff
@@ -65,15 +65,13 @@ public ValidationResult validateOnly(AdminManifest entry, Config scratch) {
                 case AdminModelManifest modelManifest -> {
                     Model model = modelManifest.spec();
                     List<ValidationWarning> warnings = new ArrayList<>();
-                    ConfigPostProcessor.validateOverridePaths(model, warnings);
```
KirylKurnosenka, line 68: Is this removed for purpose?

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigApplyService.java
siarhei-fedziukovich: Yes, deliberate — keeping it would double-report every override-path warning.

For a `Model`, `validateDeploymentInterfaces` already performs this check. Its loop body ends with an unconditional per-entry call:

```java
// ConfigPostProcessor:499
validateOverridePaths(model, entry.getKey(), declared, field, warnings);
```

and that reaches the same private overload the public `validateOverridePaths(Deployment, List)` delegates to, with an identical `field` string (`"interfaces." + key`) and identical guards (`interfaces == null` → return, `declared == null` → continue). No branch above it can skip it — `validateTranslatedInterface`'s early returns exit that method, not the loop.

So coverage is unchanged, not reduced. This is also why `processModels` never called `validateOverridePaths` separately, and the public overload's own javadoc says so: *"Models run the same per-entry check inside `validateDeploymentInterfaces`."*

Pinned by `ConfigPostProcessorTest#testValidateModelInvariantsDoesNotDuplicateOverridePathWarnings`, which asserts exactly one warning for a single bad entry, and documented on `validateModelInvariants` so the next caller does not re-add it.

server/src/main/java/com/epam/aidial/core/server/service/config/ConfigValidationService.java
siarhei-fedziukovich: Same as the `ConfigApplyService` thread — deliberate, and coverage is unchanged.

`validateModelInvariants` → `validateDeploymentInterfaces` already runs the per-entry override-path check for models (`ConfigPostProcessor:499`), delegating to the same private overload as the public `validateOverridePaths(Deployment, List)`, with the same guards and the same `field` string. Keeping both here would report every override-path warning twice.

Worth noting for this file specifically: the check stays *fatal* in soft mode. It is now part of the `hardFailure` group rather than its own `invalidOverridePaths` flag, so `/v1/admin/validate` and `ConfigApplyService#applyModel` still agree — which is what the comment above the condition is guarding.

PR #2002, opened 2026-09-20
feat: add skipOnStartup mode to config.onInvalidEntity

siarhei-fedziukovich:
### Applicable issues

- fixes #

### Description of changes

Core refuses to start when a single API-managed entity fails validation:

```
Proxy failed to start: com.epam.aidial.core.server.config.InvalidEntityException:
Model 'kk_test_model_interface': Interface 'openai_chat_completions' declares no base_url and the model declares no baseUrl
	at ConfigPostProcessor.processModels(ConfigPostProcessor.java:333)
	at MergedConfigStore.rebuild(MergedConfigStore.java:1255)
	at MergedConfigStore.init(MergedConfigStore.java:252)
	at AiDial.start(AiDial.java:259)
```

Under the default `abort` mode a per-entity violation aborts the whole merged rebuild, and when that rebuild is the one `init()` performs, the pod never comes up.

**New third mode `skipOnStartup`** alongside the unchanged `abort` (still the default) and `skip`:

| Mode | Startup (initial merge) | Every rebuild / Configuration API write after it |
|---|---|---|
| `abort` | Core refuses to start | Rebuild aborts, merged config stays as-is |
| `skip` | Entity dropped and recorded; core starts | Entity dropped and recorded |
| `skipOnStartup` | Entity dropped and recorded; core starts | Same as `abort` |

A dropped entity is not lost — it goes to the existing invalid-entity store and surfaces through `GET /v1/admin/health/config`, the admin `GET` (`"status": "invalid"` + warnings) and the `dial_config_skipped_entities` / `dial_config_skip_events_total` metrics.

Implementation: the skip/abort decision moves from two hardcoded `MODE_SKIP.equals(onInvalidEntity)` checks (the rebuild semantic pass and `skipRouter`) into `isSkipEnabled()`, which is true for `skip` always and for `skipOnStartup` only while `initialized == false` — exactly the one rebuild `init()` runs. `init()` also logs the effective mode.

**Docs:** every `config.*` static setting now lives in a **Config Loading Configurations** block in the README. `config.onInvalidEntity` and `config.write.softValidation` were undocumented, and the block spells out that the environment variable is `aidial.config.onInvalidEntity` — `aidial.onInvalidEntity` sets an unrelated top-level key and is silently ignored, which is how an intended `skip` ends up running as `abort`.

**Known ceiling, documented in the README:** `skipOnStartup` trades a failed start for a stalled reload — while the offending entity is still there, every later rebuild aborts, so the pod keeps serving the config it booted with and stops picking up changes made elsewhere. Writes through the Configuration API on that pod still apply (in-place, not via rebuild). Entities defined in the `config.files` file itself are unaffected by all three modes: `FileConfigStore`'s load is unconditionally fail-fast, so the same defect in `aidial.config.json` still fails startup.

**Tests:** 3 new in `MergedConfigStoreTest`, using a blob model with the reported defect — startup skips and records instead of failing `init`, a post-init `rebuildNow()` aborts, and mode parsing is case-insensitive with unknown values falling back to `abort`. Full `./gradlew :server:test` green (2242 tests, 0 failures), checkstyle clean.


ai-dial-actions:
<h1>Dependency Review</h1>
None

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #2005, opened 2026-09-21
feat: support /openai/v1/chat/completions endpoint #2004

### LINKED ISSUE #2004: Support OpenAI-native `/openai/v1/chat/completions` endpoint
## Context

AI DIAL Core currently exposes chat completions only via the DIAL/Azure-style path `/openai/deployments/{deployment_name}/chat/completions`, which requires the caller to embed the deployment name in the URL. Plain OpenAI SDK clients instead call `POST /v1/chat/completions` with no deployment in the URL, and put the model name in the `model` field of the JSON body.

DIAL has already adopted this deployment-less convention for two other APIs — `/openai/v1/responses` and `/anthropic/v1/messages`. This issue asks to extend the same convention to chat completions, so that unmodified OpenAI SDK clients calling `.chat.completions.create(model=..., ...)` can target DIAL Core directly at `/openai/v1/chat/completions`.

## Goal

Add a new endpoint, `POST /openai/v1/chat/completions`, that behaves exactly like today's `/openai/deployments/{id}/chat/completions` in every respect (rate limiting, upstream routing, streaming, token-usage reporting, interceptor chaining, response headers) except that the target deployment is resolved from the request body's `model` field instead of from the URL path.

## Requirements / expected behavior

- `model` may name any deployment resolvable today via a deployment id (model, application, or toolset) — no new restriction on deployment type.
- A missing or blank `model` field should result in a `400 Bad Request`.
- A `model` value that doesn't resolve to any configured deployment should result in a `404 Not Found`.
- Response behavior (headers, token-usage stats, streaming vs. non-streaming, error handling) should match the existing deployment-path endpoint exactly — no stripped-down or "pure passthrough" response variant.
- The endpoint should be available unconditionally (no feature flag / gradual rollout mechanism needed).
- The new endpoint should be documented in the generated OpenAPI specification, consistent with how the existing chat-completions and responses endpoints are documented.

## Out of scope

- No changes to the existing `/openai/deployments/{id}/chat/completions` endpoint's external behavior.
- No new "default model" or alias concept in configuration — `model` must be supplied explicitly by the caller.

astsiapanay:
Adds a new OpenAI-native `POST /openai/v1/chat/completions` endpoint that resolves the target deployment from the request body's `model` field instead of the URL path, so unmodified OpenAI SDK clients can call DIAL Core directly.

### Applicable issues

- fixes #2004

### Description of changes

- Extracted the chat-completions request/response handling shared by every chat-completions-shaped controller (enhancement chain, upstream routing, streaming/non-streaming response handling, token-usage injection) out of `DeploymentPostController` into a new `BaseChatCompletionController`.
- Added `ChatCompletionsController`, serving `POST /openai/v1/chat/completions`: reads `model` from the request body (mirrors `ResponsesController`'s pattern for `/openai/v1/responses`), returns `400` for a missing/blank `model` and `404` for an unresolvable one, and otherwise behaves exactly like the existing `/openai/deployments/{id}/chat/completions` path (rate limiting, upstream routing, streaming, token-usage reporting, interceptor chaining, response headers).
- Registered the new route (`RouteTemplate`/`ControllerSelector`) and the controller for OpenAPI spec generation (`AnnotationEndpointCollector`).
- Added integration tests (`ChatCompletionApiTest`) covering the success path, missing `model`, and unknown `model`.

### Checklist


server/src/main/java/com/epam/aidial/core/server/controller/ChatCompletionsController.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private Future<Void> proceedToDeployment(ChatCompletionRequest request, String deploymentName) {
+        return verifyLimit()
+                .compose(ignore -> proxy.getTokenStatsTracker().startSpan(context)
+                        .map(ignored -> handleRequestBody(request)))
+                .otherwise(error -> handleRequestError(deploymentName, error));
+    }
+
+    private Future<Void> handleInterceptor(int interceptorIndex) {
+        return new ChatCompletionInterceptorController(proxy, context, interceptorIndex, requestedInterface())
```
Oleksii-Klimov, line 131: If interceptors are used, will the request end up in DeploymentPostController? If so, is it expected?

astsiapanay: No, the request should be landed to the new controller. I think the interceptor should route the request to DIAL Core back.
No, it's not.

astsiapanay: Yes, it's true. I don't think that's a bug. This is a convention how the core interacts to interceptor.

Oleksii-Klimov: It was for Azure endpoints. This controller appears to expect reentry from an interceptor. But if it never executes, then maybe it can be replaced with validation instead?

Oleksii-Klimov (approved):

astsiapanay:
/deploy-review

>GitHub actions run: [35595085946](https://github.com/epam/ai-dial-ci/actions/runs/35595085946)

| [deploy-review](https://chat-ai-dial-core-pr-2005.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/35595085946/job/106317939657) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/35595085946/job/106321693488) |

PR #2011, opened 2026-09-21
fix: respond 502 when interceptor headers cannot be parsed

### LINKED ISSUE #2003: Streaming chat completion: interceptor 400 (upstream context-length) is not forwarded to the client
### Name and Version

DIAL Core 0.46.0

Interceptor built on **aidial-interceptors-sdk 0.8** / **aidial-sdk 0.39** (pass-through, OpenAI chat-completions).

Client: OpenAI Python SDK, `stream=true`.

1. Put any SDK interceptor on a chat-completions model (the stock examples should do).
2. Send a **streaming** `POST /openai/deployments/<model>/chat/completions`.
3. Make the **model provider** reject it with **HTTP 400** before any SSE chunk — e.g. prompt + `max_tokens` over the model context window:

   `maximum context length is 262144 … requested 65536 output tokens and … at least 196609 input tokens`

4. The interceptor converts that to its own JSON 400. On the wire that response has **both** headers at once:

   ```
   HTTP/1.1 400 Bad Request
   Content-Length: <n>
   Transfer-Encoding: chunked
   ```

   (`Content-Length` is for the new JSON body; `Transfer-Encoding` was copied from the upstream 400.)

The client gets a **finished HTTP response** — ideally the same **400** and provider error body as without an interceptor (or as on the Core → adapter hop). If Core cannot parse the interceptor headers, it should still close the request with **some** status (e.g. 502), not leave the stream open.

### What do you see instead?

Inner hop (interceptor → Core → adapter → provider) is fine: provider **400** in <1s, interceptor converts `openai.BadRequestError` → `HTTPException` and answers **400**.

Outer hop (client → Core → interceptor) fails. Netty refuses the dual-header 400 (`ContentLengthNotAllowedException`). Core logs:

```
InterceptorController: Proxy failed to receive response header from origin.
Address: <interceptor>:80. Error:
```

(`Error` is empty.)

The client never sees 400 **or** 502. The stream stays open until read-timeout, then the SDK retries the same request.

### Additional information

Looks like Core cannot proxy a **non-SSE 4xx** from an interceptor when the inbound call was `stream=true`. FastAPI/Starlette `HTTPException` often sends both `Content-Length` and `Transfer-Encoding: chunked`; Netty then refuses the response.

PR #1045 already treats “provider 400 instead of SSE” as normal for Core → model. The same should hold for Core → interceptor → client.

sergey-zinchenko:
Fixes #2003.

When the interceptor hop cannot parse response headers (Netty `ContentLengthNotAllowedException` on `Content-Length` + `Transfer-Encoding`, connection reset, idle timeout, …), `handleProxyResponseError` only logged. The client stream stayed open until read-timeout.

This hop has no upstream retry pool. Match `handleProxyConnectionError`: reset the interceptor connection and return **502** with `ErrorData`.

We do **not** relax Netty to accept the illegal pair (RFC 9112 §6.2). The sender side is [ai-dial-interceptors-sdk#135](https://github.com/epam/ai-dial-interceptors-sdk/pull/135). Core’s contract is: if headers still cannot be read, the client gets a finished 502, not a hang.


ai-dial-actions:
<h1>Dependency Review</h1>
None

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
None

PR #2012, opened 2026-09-21
fix: delegate GCP Redis credentials resolution to google-auth-library

KirylKurnosenka:
## Problem

`GcpCredentialsResolver` implements Redisson's `CredentialsResolver.resolve(InetSocketAddress)` contract, which is documented as asynchronous and is invoked by Redisson from `BaseConnectionHandler.channelActive(ctx)` on a shared Netty event-loop thread. The previous implementation called a blocking IAM Credentials token fetch (`IamCredentialsClient.generateAccessToken(...)`) directly inside `resolve()`, stalling every other connection multiplexed on that same event loop — including Redisson's own keepalive PINGs — for the duration of the fetch. This is a plausible root cause of periodic `WriteRedisConnectionException` / `StacklessClosedChannelException` failures seen during config reloads.

The resolver also hand-rolled its own token cache/expiry tracking, duplicating logic already provided by `google-auth-library-oauth2-http`, which is already a dependency.

## Fix

- Replace the `IamCredentialsClient`/`IamCredentialsSettings`-based token fetch with `ImpersonatedCredentials` (built via its builder) and delegate to `OAuth2Credentials.getRequestMetadata(URI, Executor, RequestMetadataCallback)`, which:
  - self-caches the token with a refresh margin (no more hand-rolled `expireTime`/cache fields),
  - de-duplicates concurrent refreshes,
  - only runs the actual blocking HTTP fetch on a supplied `Executor` — here, a dedicated `newVirtualThreadPerTaskExecutor()` — never on the caller thread. The Netty event-loop thread inside `resolve()` now only ever does cheap cache bookkeeping.
- Added defensive checks around the returned request-metadata map (missing/malformed `Authorization` entry completes the future exceptionally instead of throwing inside the callback).
- Dropped the now-unused `google-cloud-iamcredentials` and `protobuf-java` dependencies from `storage/build.gradle`.
- Added an explicit `commons-codec` version force in the root `build.gradle`: removing `google-cloud-iamcredentials` incidentally dropped the transitive pin that had been keeping `commons-codec` at a version new enough for `UrlUtil`'s direct (but previously undeclared) use of `PercentCodec`.

**Scope:** `AwsCredentialsResolver` and `AzureCredentialsResolver` have the same class of issue but are intentionally deferred to a follow-up change.

## Testing

- Added `GcpCredentialsResolverTest` covering: successful resolve, upstream failure, missing `Authorization` header, and malformed `Authorization` header.
- `./gradlew :storage:test --tests "com.epam.aidial.core.storage.cache.GcpCredentialsResolverTest"` — passes
- `./gradlew :storage:checkstyleMain :storage:checkstyleTest` — clean
- `./gradlew :storage:test` (full module) — passes aside from pre-existing, unrelated Windows-environment failures present on `development` before this change (a file-lock in `ResourceServiceTest`'s cleanup and a backslash-escaping issue in a Windows-path JSON fixture used by `ResourceServiceSenderPodIdTest`), confirmed via baseline comparison.


ai-dial-actions:
<h1>Dependency Review</h1>
<ul><li>settings.gradle</li></ul>

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
```diff
@@ -1,56 +1,71 @@
...
     @SneakyThrows
     public GcpCredentialsResolver(String accountName) {
-        this.accountName = accountName;
-        IamCredentialsSettings settings = IamCredentialsSettings.newHttpJsonBuilder().build();
-        this.iamClient = IamCredentialsClient.create(settings);
+        this.credentials = ImpersonatedCredentials.newBuilder()
+                .setSourceCredentials(GoogleCredentials.getApplicationDefault())
+                .setTargetPrincipal(accountName)
+                .setScopes(SCOPES)
+                .setLifetime((int) TimeUnit.MINUTES.toSeconds(15))
```
astsiapanay: The default lifetime is 1h. Why do we set 15 mins?

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
```diff
@@ -1,56 +1,71 @@
...
+        credentials.getRequestMetadata(null, executor, new RequestMetadataCallback() {
+            @Override
+            public void onSuccess(Map<String, List<String>> metadata) {
+                List<String> headers = metadata.get(AUTHORIZATION_HEADER);
+                if (headers == null || headers.isEmpty() || !headers.get(0).startsWith(BEARER_PREFIX)) {
+                    future.completeExceptionally(new IllegalStateException(
+                            "GCP request metadata is missing a valid " + AUTHORIZATION_HEADER + " header"));
+                    return;
+                }
+                String token = headers.get(0).substring(BEARER_PREFIX.length());
```
astsiapanay: Why do we always take the first value? What if there are more than one?

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
KirylKurnosenka: Good catch — 15 minutes isn't something this PR introduced; it's the value the original implementation used before this fix (from #769: `LIFETIME = Duration.newBuilder().setSeconds(TimeUnit.MINUTES.toSeconds(15))`), and there's no rationale recorded for it in that commit. This PR is scoped to fixing the blocking-call-on-the-event-loop bug, so I deliberately left unrelated parameters like the lifetime unchanged.

If we want to move it to the library default (1h) — fewer background refreshes/IAM calls — I'm happy to do that either as a small follow-up or folded into this PR now, whichever you'd prefer.

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
KirylKurnosenka: Verified this against the exact pinned `google-auth-library-oauth2-http:1.23.0` source rather than assuming: `OAuth2Credentials.OAuthValue.create(token, additionalHeaders)` builds the metadata map as
`ImmutableMap.builder().put(AUTHORIZATION, ImmutableList.of(BEARER_PREFIX + token)).putAll(additionalHeaders).build()` — the `Authorization` entry is always inserted first as a *singleton* list, and `ImpersonatedCredentials` doesn't override `getAdditionalHeaders()` at all (only `GoogleCredentials` does, to optionally add `x-goog-user-project` — never `Authorization`). If `additionalHeaders` ever also contained an `Authorization` key, `ImmutableMap.Builder` would throw on the duplicate key at `build()` rather than silently producing a multi-element list — so for this exact type/version the list is guaranteed to have exactly one element.

That said, I agree relying on that guarantee implicitly (by just grabbing index 0) is fragile since the `Map<String, List<String>>` contract itself doesn't encode it. Want me to tighten the check to reject `headers.size() != 1` instead of just non-empty, and/or add a comment pointing at this guarantee?

astsiapanay: Right, please make adjustments. Don't rely on the first element only.

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
astsiapanay: Let's make it default

KirylKurnosenka: Done — removed the explicit `setLifetime(...)` call, so the builder now falls back to its own default (1h). Pushed in c9c79834.

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
KirylKurnosenka: Done — instead of indexing `headers.get(0)`, we now scan every entry in the list for one that starts with `Bearer ` and use that, so it no longer matters how many values are in the list or where the bearer one sits. Added `testResolveMultipleAuthorizationValuesPicksBearerEntry` covering a list where it isn't first. Pushed in c9c79834.

ai-dial-actions:
<h1>Dependency Review</h1>
<ul><li>settings.gradle</li></ul>

storage/src/main/java/com/epam/aidial/core/storage/cache/GcpCredentialsResolver.java
astsiapanay: Wrong. You should do
```
Want me to tighten the check to reject headers.size() != 1 instead of just non-empty, and/or add a comment pointing at this guarantee?
```

KirylKurnosenka: Follow-up: on reflection, the scan-and-pick approach was hiding behind a hypothetical that doesn't actually hold up — this exact code path (`OAuth2Credentials.OAuthValue.create()`) can never produce more than one `Authorization` value; a duplicate would throw at `build()` time instead. So I swapped it for a strict check that rejects (completes exceptionally) unless the list has exactly one entry, rather than silently tolerating a shape that can't occur. Pushed in 71b6af89.

KirylKurnosenka: Done exactly as you asked — `findBearerToken` now rejects (completes the future exceptionally) unless the list has exactly one `Authorization` entry, instead of scanning for a matching one among several. Pushed in 71b6af89, with `testResolveRejectsMultipleAuthorizationValues` covering the rejection.

astsiapanay (approved):

ai-dial-actions:
<h1>Dependency Review</h1>
<ul><li>settings.gradle</li></ul>

daryapyskwork:
/deploy-review

>GitHub actions run: [35753914463](https://github.com/epam/ai-dial-ci/actions/runs/35753914463)

| deploy-review | [Failed ❌](https://github.com/epam/ai-dial-ci/actions/runs/35753914463/job/106834760361) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/35753914463/job/106839085203) |

